PR #57, opened 2020-03-11
issue #36: added refresh button to ComparePane.

### LINKED ISSUE #36: Add "Refresh" button to Shadow Controller Compare GUI
Thank you for contributing a suggestion for a new feature in PC^2!  
In order to help us understand the proposed feature, please fill in the following fields:

Currently the Shadow Controller GUI (_ShadowControlPane_) has a "Compare" button that, when pushed, creates and displays an instance of a _ShadowCompareFrame_ containing a _ShadowComparePane_ displaying the current status of comparisons between the Shadow PC2 system and the Remote CCS.  The information in this compare pane is static -- it is a one-time snapshot.  This means it is necessary to constantly close the frame, go back to the ShadowControlPane, and invoke a new ShadowCompareFrame by pressing the "Compare" button again.  This is extremely frustrating.

**Feature Description**: please provide a clear and concise description of the proposed feature:  

Add a new button to the ShadowComparePane labeled "Refresh".  When this button is clicked it will update the contents of the compare pane without requiring closing the frame and restarting it.

One way is to continue to do what's currently done (close and reopen the ShadowCompareFrame each time).  But that's irritating.

A _better_ solution than the "Refresh" button would be to have the ShadowComparePane automatically listen for changes in shadow comparison status and dynamically update itself.  However, that solution requires a significantly increased amount of effort and should be listed under its own Feature Request.

Add a Refresh button to the ShadowComparePane.
Have the actionPerformed() method for the button compute new shadow comparison data and load it into the model for the JTable being used to display comparison status.

**Additional context**: please provide any additional information that will help us in considering your feature request.  

It's not totally trivial to have the button compute new comparison data, because the current implementation passes a set of _pre-computed_ comparison information into the ShadowCompareFrame (and hence to the ShadowComparePane) when the frame is constructed by the ShadowController.  The method to compute comparison information is in the ShadowController (where it has to be in order to know status).  

One potential solution is to make method **ShadowController.getJudgementComparisonInfo()** static in ShadowController.  However, this introduces over a dozen "cannot make static reference" errors within the method.  

A better solution is probably to pass a reference to the ShadowController to the ShadowCompareFrame (instead of passing the precomputed data), and have the ShadowComparePane call back to the ShadowController  getJudgementComparisonInfo() method to get the initial data.  This way the button will have access to the ShadowController and can make a similar callback.

clevengr:
Fixes #36

## Description
Adds a Refresh button to the ShadowComparePane.  To test:

Start a PC2 Server to be the Shadow CCS.
Start a PC2EFClient, select "Shadow Mode", configure a connection to a Primary CCS
(this requires either connecting to a foreign CCS, e.g. a Kattis contest like  https://open.kattis.com/clics-api/contests/prognova20/, or else starting another PC2 server running a contest as Primary and enabling the CLICS-API in that primary).
In the Primary CCS, submit some runs (or fetch some runs from the foreign primary).
On the Shadow, select "Start Shadowing" and then "Compare".
On the Shadow Compare pane, observe how many runs are listed.
Submit more runs on the Primary.
Hit "Refresh" on the Shadow Compare pane.

Check that the Shadow Compare pane updates with the new runs.

## Issue(s)
Fixes issue #36 

## Environment
Windows 8.1, Java jre 1.8.201

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ShadowCompareFrame.java
+++ b/src/edu/csus/ecs/pc2/ui/ShadowCompareFrame.java
@@ -2,11 +2,10 @@
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.Dimension;
-import java.util.Map;
 
 import javax.swing.JFrame;
 
-import edu.csus.ecs.pc2.shadow.ShadowJudgementInfo;
+import edu.csus.ecs.pc2.shadow.ShadowController;
 
 /**
  * A JFrame which displays a Shadow comparison results pane.
@@ -18,11 +17,11 @@ public class ShadowCompareFrame extends JFrame {
 
     private static final long serialVersionUID = 1L;
     
-    public ShadowCompareFrame(Map<String, ShadowJudgementInfo> map) {
+    public ShadowCompareFrame(ShadowController shadowController) {
         Dimension size = new Dimension(600,600);
         this.setPreferredSize(size);
 //        this.setMinimumSize(size);
-        this.getContentPane().add(new ShadowComparePane(map));
+        this.getContentPane().add(new ShadowComparePane(shadowController));
     }
 
 }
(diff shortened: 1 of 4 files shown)


clevengr:
This PR was verified working in 

Version 9.7build 20200516 (Saturday, May 16th 2020 01:37 UTC) build 6137~develop
Java version 1.8.0_201
OS: Windows 10 10.0 (amd64)

PR #83, opened 2020-05-12
start with an image that has hugo pre-installed

### LINKED ISSUE #65: Create a pc2ccs.github.io repo and populate it via pc2v9 .gitlab-ci.yaml
This is a follow on to the builds and nightly-builds ticket #50.
The intent is to a create a site like icpctools/cpcptools.github.io

A new target will be added to .gitlab-ci.yaml, following the icpctools/icpctools .gitlab-ci.yaml file.

The intent of this pc2ccs.github,.io is to be linked from icpc.global

troy2914:
also fix SSH_PRIVATE_KEY

fixes #65

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -203,6 +203,8 @@ push nightly to GitHub:
       done
 
 update website:
+  # an image with hugo preinstalled
+  image: icpctools/build:servlet4
   stage: website
   rules:
     - if: '$CI_COMMIT_REF_NAME == "master"'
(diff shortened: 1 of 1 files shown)


clevengr:
I do not think we want to rely on the existence of an image defined in another (relatively) unrelated project.  We can use this for testing, but we need to be sure all the PC2 stuff is "self-contained" so it doesn't suddenly break because people in that other project suddenly make a change.

clevengr:
I am approving and merging the PR so that we can verify the solution being tested works.  However, ultimately we need to fix this in a different way (one that is self-contained; see my previous comment).

Issue #93, opened 2020-05-16
Website should support downloading tar.gz files in addition to .zip


clevengr:
**Is your feature request related to a problem?** 
Yes -- the problem is that there is currently no way on the pc2ccs.github.io website to download the **tar.gz** artifacts from a build (even though every build creates the appropriate artifacts and posts them to the appropriate GitHub "build" repo).  

**Feature Description**:  
- On the "Home" page of the website, in the "downloads" section, do the following:
  -  Expand each of the three current download links to contain the file extension (for example, right now it says "Stable Release: PC² version 9.6.0-5102"; it should say "Stable Release: PC² version 9.6.0-5102.zip".
  - Add a _second_ link, on the same line but to the right of the existing link, which provides a link to download the **tar.gz** version of the file (for example, a second link like "PC² version 9.6.0-5102.tar.gz".
   - There should be a nice amount of space between the text label "Stable Release:" and the first link, and likewise space separating the two links.

- On the "Current Downloads" page, do the following:
   - add three new buttons (one blue, two red), each to the right of the corresponding existing buttons. 
   - change the label text in each of the three _existing_ buttons to indicate that the button downloads a **.zip** file.  (For example, the current existing blue button, for downloading the "latest stable release", should now read "Download stable PC&sup2; version (9.6.0-5102) in **.zip** format"; and analogous changes to the other two existing (red) buttons.)
   - Set the labels in the three _new_ buttons to have similar text, except saying "...in **.tar.gz** format".

- On the "Previous Stable Downloads" page, do the following:
   - add new links (that is, new table data entries containing links) to the right of each existing ".zip" link which links to (downloads) the corresponding **.tar.gz** file.   
   - Include the corresponding **.sha256** and **.sha512** links (in parentheses) next to the **.tar.gz** links.
   - Include links to download the **.tar.gz** files for the "API" downloads in a similar manner.
   - Since the file sizes are going to be different for the **.tar.gz** downloads, display each file size  _after_ the corresponding link, in parentheses, in front of the SHA links.  For example, a table row might now look like:

`PC2 System     pc2v9.6.0-5102.zip  (22.53 MB) (sha256, sha512)    pc2v9.6.0-51-2.tar.gz (23.02 MB) (sha256, sha512)  `
   with appropriate cell spacing between the row data items.

- On the "All Builds" page, make changes corresponding to those described above for the "Previous Stable Downloads" page (add the **.tar.gz** links, move the file sizes after the links, etc.)

[...]

PR #96, opened 2020-05-17
i95: Added draft "history" and "permission" pages for website

### LINKED ISSUE #95: Github.io website should include additional pages from old website
No, it's more of an incompleteness in the current website.

**Feature Description**
The [old ECS website](https://pc2.ecs.csus.edu/) had numerous pages accessible from the "Home Page".  Most of these have already been incorporated, in some form or the other, into the [new website](pc2ccs.github.io).  However, at least two old pages so far don't have anything equivalent on the new site:

- [The (somewhat abridged) history of PC&sup2;](https://pc2.ecs.csus.edu/pc2history.html)
- [Permissions](https://pc2.ecs.csus.edu/pc2permission.html)

It seems reasonable that we would want these pages to move forward to the new website.  

One thing that needs to be considered is "in what form?" In particular, while the "History" page can probably be transferred more-or-less "as-is", the "Permissions" page needs to be rewritten to get across the proper sentiment in the new "open-source" reality -- that technically the system is governed by the EPL 2.0 license, and practically speaking "we would consider it unacceptable if you were trying to use it to make a profit"...

An alternative is to NOT port either of those pages...

...

clevengr:
Adds new "history" and "permission" pages to the website **/content** folder.

### Issue which the PR fixes
Fixes #95 

Windows 10 using Eclipse 2019-12.

### Additional info:
All this PR does is define new pages (files) in the **/content** folder.  The added pages are _static_; they don't contain any references to other website data (such as required **layout** information).  I don't _think_ there are any other changes which need to be done to add these pages (but I could be wrong; if so, someone please comment in a review).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -218,7 +218,7 @@ update website:
     - cd $CI_PROJECT_DIR/website
     - hugo
     - cd ~/website
-    - sed -i -e 's/&amp;/\&/' index.html current/index.html all-builds/index.html previous/index.html
+    - sed -i -e 's/&amp;/\&/' index.html current/index.html all-builds/index.html previous/index.html history/index.html permission/index.html
     - git add .
     - git commit --allow-empty -m "Update website for pc2v9 commit $CI_COMMIT_SHA"
     - git push
--- a/website/content/history.md
+++ b/website/content/history.md
@@ -1,0 +1,204 @@
+
+---
+title: "A Bit of History"
+description: "The (somewhat abridged) history of PC2"
+date: 2020-05-17T01:09:41+00:00
+draft: false
+weight: 4
+icon: fas fa-highlighter
+---
+
+This page could be titled "PC&sup2; -- the Early Years".  It covers some (certainly not all) of what happened during the first two decades of PC&sup2; development...
+
+## 1988
+
+September: Mary Richmond, Dave Bosley, Paul Meyers, and some other guy named Doug Lane make the mistake of accepting a Senior Project called PC&sup2; from Professor John Clevenger in the Computer Science Department at CSUS.
+That Lane guy still has no clue. Bliss is ignorance.
+
+## 1989
+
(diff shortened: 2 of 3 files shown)


troy2914:
- [ ]  update sed on line 221 of .gitlab-ci.yml to include new pages.

otherwise the PC&sup2; is not rendered properly.   And was it intentional to list these pages above the All Builds page?

clevengr:
> * [ ]   update sed on line 221 of .gitlab-ci.yml to include new pages.
> 
> otherwise the PC² is not rendered properly. 

Fixed; added those files to the sed command.  Also, please review https://github.com/pc2ccs/pc2v9/wiki/Building-the-PC2-GitHub-Website/_edit#webpage-post-processing to verify that I've updated the Wiki description correctly.

> And was it intentional to list these pages above the All Builds page?

No; I missed the fact that "All Builds" had a much larger weight number.  I've set the weights for "History" and "Permission" to be even higher (so they'll display lower).   Please re-review (and approve if appropriate).

clevengr:
BTW, do I not understand "sed" correctly, or is it the case that if there are two (or more) occurrences of "&amp;" on a given line in the index.html files then only the FIRST one of those will be corrected?

troy2914:
> BTW, do I not understand "sed" correctly, or is it the case that if there are two (or more) occurrences of `&amp;` on a given line in the index.html files then only the FIRST one of those will be corrected?

I don't believe that is an issue before my #93,  where i have that fixed.

troy2914:
Seems we should have some history after 2009.  Regionals, SuperRagional usage.

clevengr:
> Seems we should have some history after 2009. Regionals, SuperRagional usage.

I don't disagree; that's why I put the line at the very top about this being documentation of "the first two decades".  We can remove that, and add more (recent) history.  But I will take a while for someone to dig through and find that information, then add it to the web page.  I would like to move on to more important things at the moment, so unless you are going to tackle doing that right away I suggest we approve this and move on...

PR #103, opened 2020-05-19
fix another Linx vs Unix.

troy2914:
The last PR broke building on the pipeline.  This fixes that.

### DIFF
--- a/projects/WTI-API/buildWTI.xml
+++ b/projects/WTI-API/buildWTI.xml
@@ -33,7 +33,7 @@
 	<condition property="isBuildingOnWindows">
 		<os family="windows"/>
 	</condition>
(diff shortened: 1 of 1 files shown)


troy2914:
Seems to be some timing between npm install and it being in the PATH that ant is using.
Adding npm and nodejs to build image.

troy2914:
And now gitlab seems to be having some issues.   Tried going go the url that it was complaining about in the rebuild,  and getting a message about being blocked.

troy2914:
> And now gitlab seems to be having some issues. Tried going go the url that it was complaining about in the rebuild, and getting a message about being blocked.

gitlab is unborked.   Build with new image (with nodejs and npm) succeeded.

Issue #107, opened 2020-05-20
Add V9.7 updates to Admin Guide and WhatsNew docs


clevengr:
The PC2 Contest Administrator's Guide needs to have a new sections for the following topics:
- Web Team Interface (WTI) project.  This should probably _replace_ the existing documentation on the (old) "EWTeam".
- Shadow Mode
- Public/Judges scoreboards (support for "freezing" the scoreboard)

The What's New document also needs to be updated to reflect the addition of the above features.

PR #113, opened 2020-06-01
I108 wti scoreboard2 password

### LINKED ISSUE #108: WTI pc2v9.ini should have a default "scoreboard2=" entry
**Feature Description**: 

  It is desirable to have the setup for the WTI be as hands-free as possible.  Currently, if the user runs the WTI on the same machine as the PC2 server and uses port 8080 (the default) for browser connections to the WTI webserver, the WTI server can be started "out of the box" with zero configuration changes.  This works because the WTI project contains its own copy of a "pc2v9.ini" file which includes the default PC2 server IP/port (localhost:50002) and default WTI port (8080).

  Issue #75 proposes to add a mechanism to display the current scoreboard on the WTI browser.  In order to support that, it will be necessary to provide a way for the WTI "client" to login to the PC2 server using a "scoreboard" account, similar to the way the old EWTeam used "scoreboard2".

In the EWTeam documentation there was a requirement that the user edit the pc2v9.ini file in the webserver to contain a "scoreboard2password=xxx" entry.   It is desirable to avoid having to require users of WTI to do the same thing.

In particular, this can be avoided if the pc2v9.ini file under the WTI project already contains a default entry for "scoreboard2password=xxx".  The default should be "scoreboard2", the same as the default password that gets assigned when "Generate Accounts" is used to create the "scoreboard 2" account.   

It's possible that this could be viewed as "part of Issue #75" and could have been entered there instead.  However, it's a stand-alone item (it can be implemented independently of the actual implementation of #75, so I filed it as a separate issue so it could be worked on separately.

- passwords could be hard-coded to some other value
- the system could dynamically recognize when the scoreboard 2 account password is changed and could create a new entry for the WTI on the fly.
- Move scoreboard calculation into the PC2 server; add a new set of packets allowing fetching scoreboard information directly from the server instead of having to be logged in as a "board".
- Implement the CLICS "scoreboard" API in the PC2 webserver, and have the WTI communicate with the CLICS interface to obtain the scoreboard.

The first is a poor idea; it will fail if the user changes the password. The second will require a LOT of additional development.  The third was considered in a previous development discussion and was shot down by the group as "too complex".  The  fourth is viable; at some point the CLICS API is planned to be fully implemented.  The drawback would be that it would require the user to start another module in order for the WTI to work.

- Add a line `scoreboard2password=scoreboard2` to the pc2v9.ini file which is copied into the WTI distribution when the WTI project is built.
- When Issue #75 is finally implemented, insure that it uses the value from the WTI pcv9.ini file as the password for logging in as scoreboard 2.

clevengr:
  Updates the **packageWTI.xml** file so that after it copies the (default) **pc2v9.ini** file from the pc2v9 project source into the WTI distribution, it appends to the file a set of entries defining default "wti scoreboard account" and "wti scoreboard password" values.  Also updates the WTI-UI .classpath to reference the WTI-API project (which is necessary for the WTI-UI project to build correctly).

The default scoreboard values inserted into the WTI **pc2v9.ini** are the same as those originally used by the EWTeam:  account "scoreboard2" with the PC2 default password (also "scoreboard2").

This allows an end user to make zero configuration changes in order to use the WTI; all they have to do is create the "scoreboard2" account.  In addition, having property values in the WTI's **pc2v9.ini** file allows the user to change to a different scoreboard account if desired; all they have to do is edit the property values in the **pc2v9.ini** prior to starting the WTI.

A separate PR  (#107) has been submitted which contains Admin Guide documentation for this change.

### Issue which the PR fixes
Fixes #108 

Windows 10 with Java 1.8.0_201 and Ant version 1.10.6.

### How to Test this PR :

 1. Download the code for this PR.
 1. Use Ant to run **pc2v9/package.xml** to build a complete PC2v9 distribution (this will automatically include the WTI project distribution using the PR-updated **packageWTI.xml** file).
 1. Copy the resulting **pc2-9.xbuild-xxx .zip** or **.tar.gz** PC2 distribution file (created in the **pc2v9/dist** folder by the previous step) to any convenient location, and  unzip it.
 1. Go into the **pc2buildxx/projects** folder in the distribution and copy the **WebTeamInterface-xx** **.zip** or **.tar.gz** file to any convenient location, then unzip it.
 1. Go into the resulting **WebTeamInterface-xx** folder
 1. Examine the **pc2v9.ini** file in the **WebTeamInterface-xx** folder; verify that it contains the following entries:
  * wtiscoreboardaccount=scoreboard2
  * wtiscoreboardpassword=scoreboard2

### DIFF
--- a/projects/WTI-API/packageWTI.xml
+++ b/projects/WTI-API/packageWTI.xml
@@ -36,6 +36,9 @@
 		
 		<!-- copy pc2v9.ini to staging -->
 		<copy file="pc2v9.ini" todir="${staging.dist.dir}"/> 
+		
+		<!-- add a default scoreboard2 password to the WTI's pc2v9.ini -->
+		<concat destfile="${staging.dist.dir}/pc2v9.ini" append="true">scoreboard2password=scoreboard2</concat>
 
 		<!-- copy scripts to staging /bin folder-->
 		<mkdir dir="${staging.dist.dir}/bin" />
--- a/projects/WTI-API/packageWTI.xml
+++ b/projects/WTI-API/packageWTI.xml
@@ -16,6 +16,10 @@
 	
 	<!-- the directory where the user documentation (user manual) can be found -->
 	<property name="doc.dir" location ="doc/User Manual" />
+	
+	<!-- the set of properties we want to add to the WTI pc2v9.ini file -->
+	<property name="wtiDefaultScoreboardInfo" 
+				value="wtiscoreboardaccount=scoreboard2${line.separator}wtiscoreboardpassword=scoreboard2${line.separator}" />
 
 	<!-- remove any files left over in the distribution directory -->
     <target name="cleandist" description="remove any previous dist files">
@@ -37,8 +41,15 @@
 		<!-- copy pc2v9.ini to staging -->
 		<copy file="pc2v9.ini" todir="${staging.dist.dir}"/> 
 		
-		<!-- add a default scoreboard2 password to the WTI's pc2v9.ini -->
-		<concat destfile="${staging.dist.dir}/pc2v9.ini" append="true">scoreboard2password=scoreboard2</concat>
+		<!-- add default scoreboard account/password properties to the WTI's pc2v9.ini, in front of the "# eof" line -->
+		<!--  NOTE: this will fail (i.e., do nothing) if somebody, someday, deletes the "# eof" line from the pc2v9/pc2v9.ini file.
+		      The ramification will be that an end user will have to manually add "wtiscoreboardaccount=xxx" and 
+		      "wtiscoreboardpassword=yyy" entries to their WTI pc2v9.ini file in order to connect to the PC2 server.
+		      -->
+		<replaceregexp file="${staging.dist.dir}/pc2v9.ini"
+		               match="# eof.*"
+		               replace="${wtiDefaultScoreboardInfo}${line.separator}# eof"
+		               byline="true"/>
 
 		<!-- copy scripts to staging /bin folder-->
 		<mkdir dir="${staging.dist.dir}/bin" />
--- a/projects/WTI-API/packageWTI.xml
+++ b/projects/WTI-API/packageWTI.xml
@@ -17,10 +17,6 @@
 	<!-- the directory where the user documentation (user manual) can be found -->
 	<property name="doc.dir" location ="doc/User Manual" />
 	
-	<!-- the set of properties we want to add to the WTI pc2v9.ini file -->
-	<property name="wtiDefaultScoreboardInfo" 
-				value="wtiscoreboardaccount=scoreboard2${line.separator}wtiscoreboardpassword=scoreboard2${line.separator}" />
-
 	<!-- remove any files left over in the distribution directory -->
     <target name="cleandist" description="remove any previous dist files">
         <delete dir="${staging.dist.dir}" />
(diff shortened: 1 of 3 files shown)


projects/WTI-API/packageWTI.xml
```diff
@@ -16,6 +16,10 @@
 	
 	<!-- the directory where the user documentation (user manual) can be found -->
 	<property name="doc.dir" location ="doc/User Manual" />
+	
+	<!-- the set of properties we want to add to the WTI pc2v9.ini file -->
+	<property name="wtiDefaultScoreboardInfo" 
+				value="wtiscoreboardaccount=scoreboard2${line.separator}wtiscoreboardpassword=scoreboard2${line.separator}" />
```
lane55, line 22: In general this is the wrong place to fix a source/configuration file, for a  number of good coding and testing reasons
This change should be done in the .ini file.

projects/WTI-UI/.classpath
```diff
@@ -1,9 +1,8 @@
...
-	<classpathentry kind="output" path="bin"/>
-</classpath>
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="con" path="org.eclipse.jst.j2ee.internal.web.container/WTI-API"/>
+	<classpathentry kind="lib" path="/pc2v9/vendor/lib/junit-4.11.jar"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="output" path="bin"/>
+</classpath>
```
lane55, line 8: In the future this change should be done in its own PR.  
It is not obvious from the PR that a change in the classpath needs to be done.

projects/WTI-API/packageWTI.xml
clevengr: I think you're missing the point.  This is not about editing the configuration.  This is about establishing the INITIAL (default) configuration.   This line puts an initial value (specifically, "wtiscoreboardaccount=scoreboard2") into the initial default .ini file.  The user can then either use the default, or can edit the .ini and change it.

Failure to include this line (i.e., this change) would mean the user would NEVER be able to start the WTI server without first editing the .ini file and putting a scoreboard account in it.  That would analogous to us having a default pc2v9.ini file which did not contain "server=localhost:50002" in it -- which would mean the user would always have to edit their pc2v9.ini file in order to start the system.  We don't set up pc2v9.ini that way; we put an initial default value in it so the user can execute the pc2 server out of the box with no configuration changes.  This PR change does precisely that same thing for the WTI project -- allows the user to start the WTI server without having to first edit the .ini file (i.e., it establishes a default value for the scoreboard account, while allowing the user to edit/change it if they want).

I request that you review the intent of this change and then that you retract your request for changes to this change.

clevengr:
> The .ini file should be modified/changed directly, using an ant build script to change
> a source file is a bad precedence.

First, please see my reply to your previous comment.  If that doesn't explain things, then let me add this:  I'm not sure what you mean by "the .ini file should be modified/changed directly...".   The WTI project has its own separate pc2v9.ini file.  That file is a copy of the pc2 server's pc2v9.ini file, augmented with changes required to configure the WTI server.   If you are suggesting there should be a separate .ini file for the WTI server and that this file should be modified directly rather than by the Ant script, I say that is a bad idea:  it would require keeping two copies of a file -- one for the pc2 server and a separate one for the WTI server.  That would allow them to easily get out of sync.  Having the WTI file being copied by the Ant script, edited as necessary, and then placed into the WTI project in edited fashion,  insures that the WTI copy is always up-to-date with respect to the original pc2 copy.

Again I request that you review your understanding of what this PR change is doing, and then request that you revoke your request to change it.

projects/WTI-UI/.classpath
clevengr: In principle I agree, and I'll consider doing such, "in the future".  That "future" would be at a time when it no longer is the case that changes submitted for review sit up to seven  weeks with zero comments (as has already happened).   For now, it's hard enough to move forward without further subdividing the work into additional PRs that require separate reviews.

clevengr: Doug and I had a lengthy phone conversation about this issue.

* His argument (he can correct me if I've misstated it) is that having the packageWTI.xml code modify the pc2v9.ini file could lead to unintended consequences down the road if the packageWTI.xml code was changed.
* While I don't disagree that this COULD happen, my argument is that
  * such a thing could happen ANYWAY, and
  * the approach suggested by him suffers from the known-bad-practice of maintaining two copies of a single source code file.

Neither of us was able to convince the other that their argument had substantially more weight, leaving an impasse.  However, in the process of the discussion it was pointed out that the EWTeam project is ALREADY doing things the way Doug suggests:  it maintains a separate project-specific copy of pc2v9.ini within the project.

Given that, and in order to resolve the conflict, I agree to accept Doug's review recommendation.  I will submit a PR update which removes the pc2v9.ini editing from packageWTI.xml, and also updates the WTI-specific copy of pc2v9.ini with the appropriate initial values (the ones which the editing was previously adding to the pc2v9.ini file).

I believe this should resolve this issue.

clevengr:
Umm... GitHub seems to have attached my last comment to the portion of Doug's review which was commenting on the update to the classpath.  I've already said that I agree with his recommendation and that _in the "future"_ I will follow that approach.  

However, my last comment wasn't actually about that issue; it was about the other portion of the review:  whether or not packageWTI.xml should be editing the pc2v9.ini file.  I've agreed to submit a PR update such that it no longer does that...

PR #134, opened 2020-11-15
change format(name) to print(name)

### LINKED ISSUE #133: Balloon Summary Report doesn't handle team names containing "%"
**Describe the issue**: 

 The `BalloonSummaryReport ` class has two places where it calls `printWriter.format()` with a single argument which is a team display name.  (One such call is in method `writeReport()`; the other is in method `truncPrint()`.)  The intent of these calls is to use the team name as a "format specifier" string made up of constant characters (those comprising the team name).  This works _unless_ the team name contains a "%" sign -- in which case the character _following_ the "%" in the team name is taken by the `format()` method as a formatting (substitution) parameter.  If the next character is in fact an _invalid_ formatting substitution character, an exception is thrown.  

**To Reproduce**: 

- Start a contest with at least one problem, one team account, and one judge account.
- Login to an Admin and edit a team account display name to contain the characters "%W" in the first 9 characters.
- Submit a program from that team, and judge the submission "Yes".
- On the Admin "Reports" tab, select "Balloon Summary" and click "View Report".

**Expected behavior**:

Generates a Balloon Summary report showing one team has solved one problem (i.e., has one balloon).

**Actual behavior**: 

Generates an exception like:

**s1t1 Exception in report: Conversion = 'W'
java.util.UnknownFormatConversionException: Conversion = 'W'**

The complete exception message is shown in
[Balloons Summary.Exception.txt](https://github.com/pc2ccs/pc2v9/files/5541309/Balloons.Summary.Exception.txt)

**Environment**: 

Version 9.7build 20200723 (Thursday, July 23rd 2020 03:20 UTC) build 6149~develop
Java version 1.8.0_201
OS: Windows 10 10.0 (amd64)

but I'll bet it happens in every version of the system...

**Additional context**: 

It's not clear how to solve the problem.  I tried adding a "\\" character in front of the "%W" in the team name, hoping that would cause the `format()` method to consider the "%" as "escaped", but that didn't work -- it still threw the exception.  One approach would be to simply remove all "%" characters from the team name before calling `format()`, but that's undesirable because it produces a Report where the team names don't match other Reports.  But it's probably better than having it throw an exception...

This issue was reported by Coach during the ECPC Qualifying Round 1 in November 2020.

troy2914:
avoids invalid % in name.  Which causes the report to print an exception.

This patch was developed by first writing a failing junit test (that was the hard part).  Then fixing the report.

fixes #133

### DIFF
--- a/src/edu/csus/ecs/pc2/core/report/BalloonSummaryReport.java
+++ b/src/edu/csus/ecs/pc2/core/report/BalloonSummaryReport.java
@@ -119,7 +119,7 @@ public class BalloonSummaryReport implements IReport {
                         } else {
                             String name = problem.getDisplayName();
                             int truncIndex = Math.min(name.toString().length(), 12);
-                            printWriter.format(" " + name.toString().substring(0, truncIndex));
+                            printWriter.print(" " + name.toString().substring(0, truncIndex));
                         }
                         
                     }
@@ -140,7 +140,7 @@ public class BalloonSummaryReport implements IReport {
 
     private void truncPrint(PrintWriter printWriter, String name, int maxlen) {
         int truncIndex = Math.min(name.toString().length(), maxlen);
-        printWriter.format(name.toString().substring(0, truncIndex));
+        printWriter.print(name.toString().substring(0, truncIndex));
     }
 
     private void truncPrint(PrintWriter printWriter, Account account, int maxlen) {
--- a/test/edu/csus/ecs/pc2/core/model/SampleContest.java
+++ b/test/edu/csus/ecs/pc2/core/model/SampleContest.java
@@ -138,7 +138,7 @@ public class SampleContest {
 
         String contestPassword = "Password 101";
         Profile profile = new Profile("Default.");
-        return createContest(siteNumber, numSites, numTeams, numJudges, initAsServer, profile, contestPassword);
+        return createContest(siteNumber, numSites, numTeams, numJudges, initAsServer, profile, contestPassword, true);
     }
 
     /**
@@ -153,7 +153,7 @@ public class SampleContest {
      * @param contestPassword
      * @return
      */
-    public IInternalContest createContest(int siteNumber, int numSites, int numTeams, int numJudges, boolean initAsServer, Profile profile, String contestPassword) {
+    public IInternalContest createContest(int siteNumber, int numSites, int numTeams, int numJudges, boolean initAsServer, Profile profile, String contestPassword, boolean assignColors) {
 
         String[] languages = { LanguageAutoFill.JAVATITLE, LanguageAutoFill.DEFAULTTITLE, LanguageAutoFill.GNUCPPTITLE, LanguageAutoFill.PERLTITLE, LanguageAutoFill.MSCTITLE, "APL" };
         String[] problemsNames = { "Sumit", "Quadrangles", "Routing", "Faulty Towers", "London Bridge", "Finnigans Bluff" };
@@ -231,7 +231,9 @@ public class SampleContest {
         contest.setProfile(profile);
         contest.setContestPassword(contestPassword);
 
-        assignColors(contest);
+        if (assignColors) {
+            assignColors(contest);
+        }
 
         return contest;
     }
--- a/test/edu/csus/ecs/pc2/core/report/BalloonSummaryReportTest.java
+++ b/test/edu/csus/ecs/pc2/core/report/BalloonSummaryReportTest.java
@@ -1,0 +1,240 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.core.report;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.Serializable;
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.Scanner;
+
+import edu.csus.ecs.pc2.core.IInternalController;
+import edu.csus.ecs.pc2.core.Utilities;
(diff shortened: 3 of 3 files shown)


troy2914:
btw, when I was testing this, I updated the build.xml test target to only test **/ReportTest.java
then ran "ant test"
saves time :)

clevengr (changes requested):
This PR only fixes one of the two places where the error exists in `BalloonSummaryReport`.  Please re-read the first two sentences in the original Issue #133 and fix the other occurrence of the problem.   The JUnit (which is very nice, by the way, and the effort to create it is very appreciated; something we all need to do on a regular basis moving forward) should also be updated to test that additional change.

Also, I believe that your JUnit will incorrectly fail if a team's name is "Exceptional Team", or if any Problem name or Balloon color starts with the word "Exception" (because both team names and problem names or colors are written into the report).  While I agree that having a _color_ starting with such a word is unlikely, it's totally within reason for a team to name themselves "Exceptional Team" or for a problem to be named "Exception to the Rule" or something like that.  I suggest making the failure test more comprehensive by looking for _either_ of the two strings "Exception in report:" or "Exception generating report", which are the two possible messages which can be written by `catch` clauses in the class.  

And while I won't quibble that changing to `format("%s",name)` is an "acceptable" fix for the issue, it does seem to me that changing it to instead call `printWriter.print(name)` would be a better solution... at least, it seems more descriptive of the intent.  (But I wouldn't reject the PR just for that...)

troy2914:
> This PR only fixes one of the two places where the error exists in `BalloonSummaryReport`. Please re-read the first two sentences in the original Issue #133 and fix the other occurrence of the problem. The JUnit (which is very nice, by the way, and the effort to create it is very appreciated; something we all need to do on a regular basis moving forward) should also be updated to test that additional change.
> 
> Also, I believe that your JUnit will incorrectly fail if a team's name is "Exceptional Team", or if any Problem name or Balloon color starts with the word "Exception" (because both team names and problem names or colors are written into the report). While I agree that having a _color_ starting with such a word is unlikely, it's totally within reason for a team to name themselves "Exceptional Team" or for a problem to be named "Exception to the Rule" or something like that. I suggest making the failure test more comprehensive by looking for _either_ of the two strings "Exception in report:" or "Exception generating report", which are the two possible messages which can be written by `catch` clauses in the class.
> 
> And while I won't quibble that changing to `format("%s",name)` is an "acceptable" fix for the issue, it does seem to me that changing it to instead call `printWriter.print(name)` would be a better solution... at least, it seems more descriptive of the intent. (But I wouldn't reject the PR just for that...)

changing to use print() instead of format.   Also we control what appears in the JUnit test.   Let me figure out a failing test for the other case refereced in #133.

lane55 (approved):

troy2914:
> I do not understand your dismissal of my previous review

It was not valid anymore because I had done a rebase when I fixed the 2nd condition and added a colon to the
string being searched for.

>  it appears that you have not adequately addressed the issue I raised. Specifically, your JUnit now contains the following statement while looking through the lines generated by the Balloon Summary report:
> 
> ```java
>                 if (data.indexOf("Exception:") != -1) {
>                     System.err.println("Found Exception in " + data);
>                     throw new AssertionError("BalloonSummaryReport threw an exception");
>                 }
> ```
> 
> This has the following flaws:
> 
>     * if a team name contains the string `Exception:`, the JUnit will incorrectly report that the BalloonSummaryReport threw an exception.
> 
>     * if a problem name contains the string `Exception:`, the JUnit will incorrectly report that the BalloonSummaryReport threw an exception.
> 
>     * if the BalloonSummaryReport DOES throw an exception but the output of that exception does _not_ contain the string "Exception:", the JUnit will FAIL to report that the BalloonSummaryReport threw an exception.
> 
> 
> As I've already said, I've seen many examples of teams using the word "Exception" in their name, and it's totally reasonable that a problem name would include that string. (The addition of the colon helps a little bit, but it doesn't work if a team calls themselves "Exception:None" or some similar nonsense; likewise for problem names).
> 
> While we cannot predict (nor prohibit) what teams will name themselves, I think it's very easy to modify the JUnit so that it is MUCH less likely to have these problems: stop looking for the string "Exception:", and start looking instead for the two specific strings which can be output by the `catch` clauses in the BalloonSummaryReport class: "Exception in report: " and "Exception generating report ". Note that BOTH these strings contain trailing blanks, so even if a team (or problem) was actually named exactly the indicated string, it would be VERY unlikely to also have a trailing blank.
> 
> In other words, I think the JUnit would be MUCH more robust if the code listed above were changed to:
> 
> ```java
>     if (data.indexOf("Exception in report: ") != -1 || data.indexOf("Exception generating report " != -1)) {
>         System.err.println("Found Exception in " + data);
>         throw new AssertionError("BalloonSummaryReport threw an exception");
>     }
> ```
> 
> This would solve all three of the above-listed shortcomings of the current code.

You are saying the test that the JUnit creates,  would contain Exception: in some part of the data?

Again this is a JUnit test, that we control the data.   The looking for Exception: is only in the JUnit.

troy2914:
rebased, and squashed my 2 commits.

clevengr (approved):
I (finally) understood what you were talking about regarding "the exception message coming from the JUnit":  you were talking about the JUnit method `testBalloonSummaryReportProblemName()`, which uses method `BalloonSummaryReport.writeReport()` to generate a report; `writeReport()` does not add any exception messages to the report -- rather, if `writeReport()` throws an exception then the JUnit method catches the exception and adds a (known) message to the report. 

On the other hand, **_I_** was talking about JUnit method `testBalloonSummaryReport()`, which uses method `Utilities.createReport()` (called by `BalloonSummaryReportTest.createAndViewReport()`)  to generate a report, and `Utilities.createReport()` DOES add exception messages to the report.  I was talking about the fact that your original version wouldn't detect the existence of those messages added to the report by `Utilities.createReport()`.  Your latest update makes it so that both types of exceptions (those added by the JUnit and those added by `createReport()` will be detected, so I'm approving the PR.

I'm not crazy about the fact that the JUnit writes stuff to System.err; I think we want to move in the direction of reducing such output.  But that effort is outside the scope of the current issue...

PR #137, opened 2020-11-25
i136: add "windows eol/eof" params to Viva pattern files.

### LINKED ISSUE #136: Viva test data files fail on GitLab
**Describe the issue**: 
The `VivaAdapaterTest` JUnit relies on a number of test files stored in `testdata/VivaAdapterTest`.  Several of these files represent "judge's test data files" which Viva is asked to validate, while others represent "Viva patterns" which are used to control the Viva configuration while performing testing.  These files are (of necessity) created on _some_ OS platform, and will thus have the EOLN/EOF characteristics of that platform.  

This means that the JUnits which use these files will work correctly on the original platform.  However, when the files are uploaded to GitHub and merged to the develop branch (triggering a GitLab build), the files fail to work properly because the JUnits are running on a (GitLab) Linux machine -- meaning that Viva is expecting the files to be in Linux format.  (In other words, Viva defaults to expecting files which are given to it to be in the format of the platform on which it is running.)

Viva has support for avoiding this issue.  Specifically, a Viva pattern file can contain a "parameter list" which says what EOLN/EOF conventions it should expect the input files to have.  Since the VivaAdapterTest data files were created on Windows, the pattern files which are used with them should contain a parameter list telling Viva to assume the files are in Windows format.  Specifically, that parameter list looks like: `# eolnstyle="windows" eofstyle="windows" #`
as the first line of the pattern file.

**To Reproduce**: 
Run the current VivaAdapterTest JUnit on a Linux machine.

**Expected behavior**: 
The JUnit should pass all tests, just like they do on a Windows machine.

**Actual behavior**: 
The JUnit fails on three of the tests -- specifically, all of those which are driven primarily by the end-of-line/end-of-file conditions (which are: `testFailOnMultipleDataPerLine()`, `testFailOnExtraDataPastSentinel()`, and `testFailOnNegativeData`.)

**Environment**: 
Testdata files created under Windows 10.  VivaAdapterTest runs successfully on this platform.
The same JUnits and testdata files fail on the GitLab Linux machine.

**Screenshots**:
The following shows the failure of the JUnit tests on the GitLab Linux machine (these same JUnit tests all pass on Windows 10):

image

**Additional Information**:
The recommended fix for this issue, suggested by Vanb (VIVA's author), is to add the following "parameter list" line to the Viva pattern files used by the VivaAdpaterTest JUnit:
`# eolnstyle="windows" eofstyle="windows" #`

clevengr:
Adds the Viva parameter list `# eolnstyle="windows" eofstyle="windows" #` to the four Viva pattern files used by the VivaAdapterTest JUnit, as recommended by Viva author Vanb.

### Issue which the PR fixes
Fixes #136 

Windows 10, Java 8_201.

- Download the code and run the VivaAdapterTest JUnit on a Windows machine; verify it passes all tests.
- Download and run the same JUnit on a Linux machine; verify it passes all tests.

### Additional Notes:
An alternative way to fix Issue #136  would be to add to all the Viva JUnits code which 

1. detects the current OS platform, 
2. reads the `SerializedFile`s containing each of the test data files in the `testdata/VivaAdapterTest` folder, 
3. tests each of those files to determine its _current_ OS eol/eof conventions, 
4. writes new data files conforming to the current OS platform, 
5. creates new SerializedFiles for each of those new data files, and 
6. uses the updated SerializedFiles in doing all the Viva tests.

In addition to being WAY more work than is necessary (as opposed to the simple fix given in this PR, which is to _use Viva's already-built-in support for this issue_), this would be subject to failure if new test data files were added in the future; the JUnit would have to be updated to recognized those new files which also need to be converted.  With the current approach (the one recommended by Vanb), new data files would "just work correctly".

### DIFF
--- a/testdata/VivaAdapterTest/Viva.InvalidPattern.viva
+++ b/testdata/VivaAdapterTest/Viva.InvalidPattern.viva
@@ -1,3 +1,4 @@
+# eolnstyle="windows" eofstyle="windows" #
 {
 [= n (n=0)]
 n:integer (minint<=n<=maxint)
--- a/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnDataPastZeroSentinel.viva
+++ b/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnDataPastZeroSentinel.viva
@@ -1,3 +1,4 @@
+# eolnstyle="windows" eofstyle="windows" #
 {
 [= n (n=0)]
 <n>;
--- a/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnMultipleDataPerLine.viva
+++ b/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnMultipleDataPerLine.viva
@@ -1,3 +1,4 @@
+# eolnstyle="windows" eofstyle="windows" #
 {
 n:integer ;
 }
--- a/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnNegative.viva
+++ b/testdata/VivaAdapterTest/Viva.ValidPattern.FailOnNegative.viva
@@ -1,3 +1,4 @@
+# eolnstyle="windows" eofstyle="windows" #
 {
 <n:integer (0<=n)>;
 }


ashoos72 (approved):
Looks good.

PR #141, opened 2020-11-26
update node and npm versions

### LINKED ISSUE #140: builds are failing due to nodejs/npm versions.
as of Nov 20, 2020 it seems nodejs v8.10.0 (with npm version 3.5.2) is no longer supported.

troy2914:
from v8.10.0/3.5.2 to v14.15.1 (with npm 6.14.8)

fixes #140

to reproduce the steps you would have to start with nodejs and npm on ubuntu18.04
which should get to v8.10.0/3.5.2
(or you can look at the fix_npm pipelines in gitlab.com to see that node and npm are being upgraded,  and the build is now working.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,6 +30,14 @@ job-build:
   stage: build
   script:
     - echo "Running ant on package.xml"
+    # we were running node/npm v8.10/3.5.2 which started failing in Nov of 2020
+    # install "n" the Node/npm version manager/updater
+    - npm install -g n
+    # Install the latest stable version of nodejs/npm
+    - n stable
+    - export PATH="/usr/local/bin:$PATH"
+    - npm --version
+    - node --version
     - ant -file package.xml
     - echo $CI_JOB_ID > dist/ci_job_id.txt
   artifacts:


clevengr (changes requested):
- I had to spend a half-hour trying to understand what "n" was and how it worked.  I'd like to see a comment added right before the command `npm install -g n` like
`# install the tool "n", which is a Node/npm version manager/updater`
and a comment right before the `n stable` command like
`# tell "n" to download the latest stable version of node`.

- I cannot find any documentation on the `stable` keyword for "n" -- docs  I found (https://www.npmjs.com/package/n) list only `latest` and `current` as similar keywords. Where did the `stable` keyword come from?  It would be nice if there was a comment pointing to an appropriate reference for "n".

- The two successful GitLab builds did not build/update the PC2 GitHub.io webpage -- so I can't download/test the resulting build to verify that all the Node/Angular stuff works correctly.  How come it didn't generate a new GitHub.io build?  How do I access the zip/.targz for the build to test it?

troy2914:
My reference was:
https://phoenixnap.com/kb/update-node-js-version

And
https://gitlab.com/pc2ccs/pc2v9/pipelines
the last 2 builds for fix_npm in the in job-build is where the fix was put. 
Why the call to pc2.ecs.csus.edu is not working is a different issue.

troy2914:
I added more documentation in the last commit/rebase.

troy2914:
only master and develop branches trigger github.io update.  the "update website" stage in .gitlab-ci.yml

troy2914:
Dirs already exist, which causes the grab to fail.
from https://gitlab.com/pc2ccs/pc2v9/-/jobs/873333979  click on Job artifacts "artifacts"

clevengr (approved):
LGTM, although it remains to be seen whether the "build" generated also works.

PR #111, opened 2020-05-31
add .gitattributes to enforce consistent eol settings

troy2914:
also ran
$ git add --renormalize .
to fix the existing files per the new .gitattributes

fixes #110

PR #152, opened 2020-12-03
add .gitattributes to enforce consistent eol settings

### LINKED ISSUE #110: add .gitattributes and fix eol settings in repo
.gitattributes allows controlling the eol settings per regex

177 java source has non-native eol currently.

troy2914:
also ran
$ git add --renormalize .
to fix the existing files per the new .gitattributes

fixes #110

this replaces #111


troy2914:
unclear why it considers those files are conflicted.   All i did was add the .gitattributes and run the normalize and commit that.

troy2914:
> unclear why it considers those files are conflicted. All i did was add the .gitattributes and run the normalize and commit that.

ah, because i had done a git fetch upstream,  but not a git merge upstream/develop before starting my branch.
I also noticed i forgot to git add the .gitattributes in the original.  So both of those have been resolved in the current push.

clevengr (commented):
I viewed the first 50 changed files (past the `.gitattributes` file) and the last 50 changed files and verified that they all had changes of the form "@@ -1,46 +1,46 @@" (that is, the number of inserted lines was exactly equal to the number of deleted lines).  That all seems correct.

I cloned the forked repo and build a new local distribution, then ran a smoke test:  loaded a contest with multiple problems, teams, and languages; logged in multiple teams (using simultaneous logins and with both the Fat Team Client and the WTI), submitted multiple runs and got back the expected AutoJudged results, and manually re-judged a run.  It looks to me like the build works correctly.

I have some questions about the `.gitattributes` file:

- I see the two lines
```
testdata/**  -text
samps/**  -text
```
I understand the negation for `testdata/**` (there might be testdata files that intentionally have a certain kind of Line/File ending).  But I'm wondering why `samps/**` is negated?  Wouldn't we always want the EOL/EOF markers in samps files to be updated on checkout?

- I see the following two lines:
```
*.doc     binary
*.doc     binary
```
Why is this line repeated? (Am I missing something?)

Approval pending your reply to the above...

troy2914:
primarily because i did not want to consider stuff under samps (all 1123 files).

clevengr (approved):
Ok.  Looking under samps/** I see that there ARE in fact some .exe files (as well as text files), so I guess that's an issue for another time.

Issue #196, opened 2021-03-03
Setting Allow multiple logins per team  - de-selected when other settings are updated


lane55:
**Describe the issue**:

When certain settings on the Settings tab are updated
the Setting Allow multiple logins per team is unchecked/ de-selected 

There may be a number of Settings fields that cause 
this bug.

**To Reproduce**:

1. Start server, start admin
1. check ```Allow multiple logins per team ```
1. Update (to save it)
1. Edit Scoring Properties, Update then Update 

**Expected behavior**:

Allow multiple logins per team should be checked

**Actual behavior**:

Allow multiple logins per team it NOT checked

**Environment**:

**Log Info**:

**Screenshots**:

**Additional context**:

lane55:
Assigning to clevengr to verify that this issue is a dup of issue #206

clevengr:
Duplicate of #206... although THIS issue (in the "Steps to Reproduce" claims that the Allow Multiple Logins state is lost merely by editing/updating the Scoring Properties; I was never able to demonstrate that failure.  Rather, as stated in Issue #206, the state is lost when the Admin is _restarted_ after enabling the checkbox.

PR #207, opened 2021-03-20
i206: preserve state of AllowMultipleLogins on Admin restart. Fixes #206

### LINKED ISSUE #206: Setting of "Allow Multiple Team Logins" is not preserved properly
**Describe the issue**: 
  At the 2020 PacNW Contest (in March 2021) we noticed that if changes were made on the Admin "Settings" tab after the "Allow Multiple Logins Per Team" checkbox had already been checked and saved ("Update" pressed), the value of the "Allow Multiple Logins" checkbox returned to "unchecked".

  Further testing seems to indicate the following:
- the checkbox state is properly saved on the server when it is initially checked and followed by an "Update" (that is, if you check the box, click "Update", then exit and restart the Admin, the box does come back "checked", as it should).
- When you FIRST make a change to some other settings, whether it is changing ONE setting or MULTIPLE settings, pressing UPDATE incorrectly causes the "Allow Multiple Logins" checkbox to get cleared.
- If you make SUBSEQUENT changes to other settings (whether a single change or multiple changes), pressing UPDATE does NOT alter the state of the "Allow Multiple Logins" checkbox.

**To Reproduce**: 
- Start an Admin
- On the Settings tab, check "Allow Multiple Logins per Team"
- Click Update.
- Exit the Admin
- Restart the Admin, select the Settings tab, verify that "Allow multiple logins per team" is checked (it should be).
- Change any other setting on the Settings tab
- Click "Update".

**Expected behavior**:   
The new setting you changed should be reflected on the Settings screen, but the "Allow multiple logins per team" checkbox should NOT have changed (it should still be CHECKED).

**Actual behavior**: 
Pressing UPDATE after changing some other setting causes the "Allow multiple logins per team" checkbox to become UNCHECKED.  Shutting down the Admin and restarting it also causes the checkbox to come up UNCHECKED -- meaning, the "checked" state has been permanently lost.

**Environment**: 
 - Windows 8.1 with Java 1.8.0_201 (but it probably happens under any environment)
 - PC2 Version 9.7.0-6181

clevengr:

Fixes the issue of losing the state of "Allow Multiple Logins per Team" when the Admin is shut down and restarted.

The root cause of the issue was that the state of the "allow multiple logins" condition was being tracked by a boolean field (variable) which was not being saved/restored when the Admin was restarted.

Specifically, the field `allowMultipleTeamLogins` was initialized "false" when the AdminView's ContestInformationPane was constructed.  Subsequently, checking the "Allow multiple logins per team" checkbox then clicking "Update" was saving the state of the checkbox in the Contest Settings on the server (as it should), along with updating the value of the boolean variable.  However, the _value of the boolean variable was NOT being saved on the server_.   

The result was that when the Admin is restarted, `populateGUI()` repopulates the checkbox _from the contest settings fetched from the server_ (as it does for all GUI components).  This caused the checkbox itself to become "checked" (because the state of the checkbox was saved on the server), implying that the previous activation of that setting was preserved (which in essence it was).  

However, when any OTHER setting on the GUI was changed, the `updateContestInformation()` method (the actionListener for the Update button) was invoking `getFromFields()`, which was updating the contest information _based on the value of the boolean variable, not the current state of the checkbox_.

The above analysis revealed that _the only place the boolean field was being used was in `getFromFields()`_, so rather than complicate things by trying to also save the boolean field in the contest settings on the server, instead the field was simply _eliminated_, and the current value of the `allowMultipleTeamLogins` checkbox was used to update the contest settings.

### Issue which the PR fixes

Fixes #206 

Windows 8.1, Java 1.8.0_201, Eclipse Version: 2019-12 (4.14.0)

- start a server with --load sumithello 
- start an Admin
- select the "Settings" pane
- Check "Allow multiple logins"
- Click Update
- Exit Admin

- Restart Admin
- Verify that "Allow Multiple Logins is still checked (it should be; this worked prior to this PR)
- Change any setting other than "Allow Multiple Logins"
- Click Update

Previously, the above sequence would cause the "Allow Multiple Logins" checkbox to become UNCHECKED.  Now, the checkbox remains checked while the other settings are updated.

To verify this, stop and then restart the Admin again; observe that the Allow Multiple Logins checkbox is still checked.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
@@ -213,7 +213,6 @@ public class ContestInformationPane extends JPanePlugin {
     private JCheckBox shadowModeCheckbox;
 
     private JCheckBox allowMultipleTeamLoginsCheckbox;
-    private boolean allowMultipleLoginsPerTeam = false;
 
     private Component rigidArea1;
     
@@ -857,7 +856,7 @@ public class ContestInformationPane extends JPanePlugin {
         String maxFileSizeString = "0" + getMaxOutputSizeInKTextField().getText();
         long maximumFileSize = Long.parseLong(maxFileSizeString);
         newContestInformation.setMaxFileSize(maximumFileSize * 1000);
-        newContestInformation.setAllowMultipleLoginsPerTeam(allowMultipleLoginsPerTeam);
+        newContestInformation.setAllowMultipleLoginsPerTeam(getAllowMultipleTeamLoginsCheckbox().isSelected());
 
         //fill in values already saved, if any
         if (savedContestInformation != null) {
@@ -870,7 +869,9 @@ public class ContestInformationPane extends JPanePlugin {
             newContestInformation.setContestShortName(savedContestInformation.getContestShortName());
             newContestInformation.setExternalYamlPath(savedContestInformation.getExternalYamlPath());
             
+            //TODO: why is the following being done here when it is overridden below?
             newContestInformation.setFreezeTime(savedContestInformation.getFreezeTime());
+            
             newContestInformation.setLastRunNumberSubmitted(savedContestInformation.getLastRunNumberSubmitted());
             newContestInformation.setAutoStartContest(savedContestInformation.isAutoStartContest());
         }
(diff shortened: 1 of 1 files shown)


lane55:
Note that this PR may also fix Issue #196 - assigned issue #196 to clevengr to verify that #196 is a dup of issue #206

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
```diff
@@ -213,7 +213,6 @@
     private JCheckBox shadowModeCheckbox;
 
     private JCheckBox allowMultipleTeamLoginsCheckbox;
-    private boolean allowMultipleLoginsPerTeam = false;
```
lane55, line 216: Good change, allowMultipleLoginsPerTeam should not be used for MultipleTeamLogin state, the JCheckBox should be and is now used.

lane55 (approved):
Code fix approved.   I also created a new dist based on i206 branch and tested.
Confirmed that PR fixes bug.

PR #272, opened 2021-06-18
I239 shadow scoreboard comparison

### LINKED ISSUE #239: Provide a way to compare Primary CCS Scoreboard with PC2 Scoreboard

Yes, when comparing shadow and pc2 scoreboards a very manual
process must be used to compare the ranks and scores.

**Feature Description**: 

Add a feature that shows a comparison between pc2 standings and the primary CCS standings.

Add a feature to view/save to disk the pc2 CLICS scoreboard JSON.

Yes, manual.  Time consuming and potential mistakes.

Add two new buttons/features on the Shadow Tab that will:
1 - view the pc2 CLICS scoreboard JSON
2 - compare the pc2 scoreboard with the Primary CCS scoreboard

Create the following methods. 

```
// in class public class ScoreboardJSON
 public String createJSON(IInternalContest contest)  // logs via StaticLog
 public String createJSON(IInternalContest contest, Log log)
 public String createJSON(String xml)
```
All methods will throw parsing and other appropriate Exceptions

**Additional notes**: 

Consider adding a View button to show the Primary CCS scoreboard JSON.

Once code complete replace pc2 CLICS scoreboard method, use the createJSON method.

**Additional context**: 

none.

clevengr:

Provides a facility for automatic comparison of the scoreboard generated by a PC2 Shadow system with the scoreboard generated by the Remote CCS being shadowed.   
- Adds a new button "Compare Scoreboards" to the main Shadow Controller GUI.
- Pressing "Compare Scoreboards" displays a new window containing the current PC2 and Remote CCS scoreboards, shown side-by-side.
- Rows which agree (in Rank, Team Name, Number Solved, and Total Time) between the two scoreboards are shown in green; rows which differ are shown in red; rows which exist in one scoreboard but have no corresponding row in the other (for example, a row with "Rank=5" in one scoreboard but no such row in the other) are shown as a set of question marks.
- A single scroll bar allows simultaneous scrolling of the two scoreboards.
- A _Refresh_ button allows the user to refresh (update) the scoreboard display.

### Issue which the PR fixes
Fixes #239.

Eclipse 2019-12, Windows 10, Java 1.8_201

- Configure a PC2 Server with a contest.  This will be the Shadow system.  Include in the Shadow configuration the information necessary to connect to a Primary (Remote) CCS (that is, to connect to the CCS being shadowed).
- Start a Remote CCS with the same contest configuration as that used to configure the PC2 Shadow.
- Submit a set of runs to the Remote CCS.
- Use the PC2 Admin to start the contest on the PC2 Shadow.
- Start a PC2 Event Feed client, select the "Shadow Mode" tab, and click "Start Shadowing".  This will cause the PC2 Shadow to connect to the remote CCS and fetch runs from its event feed.
- Click the "Compare Scoreboards" button on the PC2 Event Feed (Shadow) client.  This will display the current PC2 and Remote CCS scoreboards.   (Note: the scoreboards will likely be out of sync initially, since the PC2 Shadow will not necessarily be judging runs at the same speed as the Remote CCS.  However, once all runs have been judged then the Scoreboards should be identical.  If not, press the "Compare Runs" button to see if PC2 judged one or more runs differently; if so, adjust the PC2 judgement to match that assigned by the Remote CCS).

Note:  rather than configuring a remote CCS, another way to accomplish the above is to connect to an _existing, completed_ contest, and to configure the PC2 Shadow with that contest's configuration.  For example, connect to the Kattis "NADC21" contest, at https://nadc21.kattis.com/clics-api/contests/nadc21.  This will allow the PC2 Shadow to fetch the Remote CCS Event Feed.  This is how the initial development of the PR was accomplished/tested.  The complete CDP for configuring a PC2 Shadow to match the Kattis NADC21 contest can be found here:

[cdp.zip](https://github.com/pc2ccs/pc2v9/files/6674740/cdp.zip)


src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
```diff
@@ -0,0 +1,391 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import edu.csus.ecs.pc2.core.log.Log;
+import edu.csus.ecs.pc2.core.standings.json.ScoreboardJsonModel;
+import edu.csus.ecs.pc2.core.standings.json.Team;
+import edu.csus.ecs.pc2.core.standings.json.TeamScoreRow;
+import edu.csus.ecs.pc2.ui.ShadowCompareScoreboardPane.ScoreboardType;
+
+public class ShadowScoreboardComparator {
```
lane55, line 21: Rename this class (ShadowScoreboardComparer ?  ShadowScoreboardUtilityy?)    There are over 10 implements Comparator classes that implement implements java.util.Comparator and are used with sort() to sort.  
This class name implies that this class is a java.util.Comparator rather than a class that returns a list of comparisons.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
```diff
@@ -0,0 +1,391 @@
...
+     *  If an error occurs while converting either input JSON string to a collection of corresponding TeamScoreRow
+     *  objects, an array of size zero is returned.
+     *  
+     * @param pc2ScoreboardJson a CLICS Scoreboard JSON string for the current PC2 scoreboard.
+     * @param remoteScoreboardJson a CLICS Scoreboard JSON string for the remote CCS being shadowed.
+     * 
+     * @return an array of comparisons between rows of the PC2 and Remote CCS scoreboards; the returned array
+     *          will be empty (size=0) if an error occurs while processing the input JSON strings.
+     */
+    public ShadowScoreboardRowComparison [] compare (String pc2ScoreboardJson, String remoteScoreboardJson) {
```
lane55, line 62: fyi - this method is nearly 100 lines long and it is not a simple switch or if/else set of blocks.  
In the future consider moving the for rank block of code into its own method.   Then both
that new method and the compare method could have their own unit tests.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
```diff
@@ -0,0 +1,391 @@
...
+        
+        //convert the provided teamsJson, assumed to be a CLICS teams description, into a POJO
+        // containing a list of teams
+        ObjectMapper mapper = new ObjectMapper();
+        
+        List<Team> teamList = null;
+        try {
+            teamList = mapper.readValue(teamsJson, new TypeReference<List<Team>>() {});
+        } catch (IOException e) {
+            log.severe("Exception converting teams JSON description to TeamList class: " + e.getMessage());
```
lane55, line 237: Add the exception to the log message.   s/e.getMessage()/e.getMessage(), e);
If it is severe we want the stack trace in the log.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
```diff
@@ -0,0 +1,391 @@
...
+     * {@link TeamScoreRow}s taking into account only the teamId.  That is, this method ranks TeamScoreRows by the numeric TeamId.
+     * 
+     * This is useful, for example, when other tests (e.g. a call to {@link TeamScoreRow#compareTo(TeamScoreRow)}
+     * has determined that two TeamScoreRows are "equal" (that is, have the same rank, number solved, and total
+     * points).  This method provides a way to order different teams in the absence of having the Team Display Name
+     * available for alphabetic ordering.
+     * 
+     * @author John Clevenger, PC2 Development Team (pc2@ecs.csus.edu)
+     */
+    private class TeamScoreRowByTeamIdComparator implements Comparator<TeamScoreRow> {
```
lane55, line 365: fyi - this should be in its own class (and yes this is a Comparator)

src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardPane.java
```diff
@@ -0,0 +1,545 @@
...
+     *                  specified {@link ShadowScoreboardRowComparison} array and used to populate the returned TableModel.
+     * @param comparedResults an array of ShadowScoreboardRowComparisons containing comparisons of corresponding rows of the
+     *          pc2 and remote scoreboards.
+     * 
+     * @return a {@link TableModel} populated with table information for the specified scoreboard.
+     */
+    private TableModel getUpdatedResultsTableModel(ScoreboardType scoreboardType, ShadowScoreboardRowComparison[] comparedResults) {
+
+        //define the columns for the table
+        //TODO: use Enum name/values instead of hard-coded strings for columnNames
```
lane55, line 401: fyi - I disagree that this string array should be an enum.   One, it is busy work (no real value add) Two as is this array is consistent with all the other column definitions.   Three not all lists should be enums.  Four enums should be used in switch statements and in locations where they are likely to be changed to add new enum values.

src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardPane.java
```diff
@@ -0,0 +1,545 @@
...
+            
+            + "\n\nThe effect of the above is that early in a contest when the PC2 Shadow system is still catching up to the Remote CCS, there will typically"
+            + "\nbe many rows where one scoreboard has data that the other does not (because of ties within ranks due to the submissions which that system"
+            + "\nhas/has not judged), and there will be many rows where there is a single rank in both scoreboards but the data values will not match"
+            + "\n(the rows will be red) because the PC2 Shadow system has not judged precisely the same number of submissions as the remote CCS."
+            
+            + "\n\nAt the end of the contest the Scoreboard Comparison screen will show the true comparison of scoreboards for the PC2 Shadow and the Remote CCS."
+            + "\n(If the scoreboards do not match at the end of the contest, the most likely reason is because the PC2 Shadow system has given a different"
+            + "\njudgement than the Remote CCS to one or more submissions; see the PC2 Shadow \"Compare Runs\" grid to find and resolve such differences.)"
+            + "";
```
lane55, line 530: Why is there an empty string at the end of this string, can it be removed ?

lane55 (changes requested):
There are missing log messages for state changes.
Stack traces are not being logged in catch blocks.
And there are other improvements noted.

src/edu/csus/ecs/pc2/shadow/ShadowController.java
```diff
@@ -345,9 +357,168 @@ private IRemoteContestAPIAdapter createRemoteContestAPIAdapter(URL url, String l
...
+     * remote CCS Contest API "/scoreboard" endpoint. 
+     * 
+     * @return a String containing the remote CCS scoreboard JSON.
+     */
+    private String getRemoteScoreboardJSON() {
+        
+        //get scoreboard from remoteAPIAdaptor
+        String remoteJson = remoteContestAPIAdapter.getRemoteJSON("/scoreboard");
+        
+        //return scoreboard
```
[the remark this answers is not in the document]
clevengr: What "debug code comment" are you talking about?   There's nothing in the vicinity of this note which is "debug code"...

src/edu/csus/ecs/pc2/shadow/ShadowController.java
```diff
@@ -345,9 +357,168 @@ private IRemoteContestAPIAdapter createRemoteContestAPIAdapter(URL url, String l
...
+            retJson += "\"name\":\"" + teamName + "\"";
+            retJson += "}";
+            
+            first = false ;
+        }
+        
+        //close the array of teams
+        retJson += "]";
+        
+        //debug
```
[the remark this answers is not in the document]
clevengr: Done.

src/edu/csus/ecs/pc2/shadow/ShadowController.java
```diff
@@ -345,9 +357,168 @@ private IRemoteContestAPIAdapter createRemoteContestAPIAdapter(URL url, String l
...
+     */
+    public String getRemoteTeamsJSON() {
+        
+        log.info("Fetching Remote CCS team JSON");
+        
+        //get team list from remoteAPIAdaptor
+        String remoteJson = remoteContestAPIAdapter.getRemoteJSON("/teams");
+        
+        //debug
+        System.out.println ("Remote CCS team json: ");
```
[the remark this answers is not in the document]
clevengr: Done.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
clevengr: Done.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
clevengr: I am assuming that "fyi" and "In the future" means that you accept that I'm not going to change this now -- for several reasons.  First, there would be strong coupling between such a new method and the existing method, requiring passing multiple parameters and/or changing local variables to being fields (which increases vulnerability).  Second, I'm pretty certain that we all agreed years ago to set the "Warning limit" on method size in PC2 code to 200 lines; this method is less than 50% of that.  Third, I don't think the method is particularly confusing or hard to read as-is:  it starts with a variety of sanity checks and data updates, followed by a main loop which is less than 40 lines (of which over 1/3 are comments/blanks).

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
clevengr: The suggested change will not work; method Log.severe() does not support passing a second parameter of type Exception.  However, I have made an equivalent change so that the exception gets logged.  I also removed the "debug output" which follows the log statement.

src/edu/csus/ecs/pc2/shadow/ShadowScoreboardComparator.java
clevengr: If it was in a separate file it would be accessible to other classes, and that's not the intent.
And yes, I know what a Comparator is.  I just disagree that because you call something a Comparator that it necessarily means it has to implement java.util.Comparator.   If I write a class which compares two things, it's a comparator.  If it doesn't "implement java.util.Comparator", it's not a java.util.Comparator -- but it can still be a class that compares things -- a comparator.

src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardPane.java
clevengr: Well, we can spend some time, sometime, discussing/debating why our opinions on this differ...  which they do.

src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardPane.java
clevengr: It was there to make it easier to update the previous lines of text without constantly worrying about making sure the last line of the extended string ends with a semi-colon.  Technically the empty string itself (the double quotes) is not needed -- but the semicolon is.   I removed the double quotes -- although I think it's easier to read when they are present.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -53,9 +53,7 @@
 
     private JPanel buttonPanel = null;
 
-    private JButton startButton = null;
-
-    private JButton stopButton = null;
+    private JButton startStopButton = null;
```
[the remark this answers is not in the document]
clevengr: I don't understand the statement `there are a number of Exceptions that are not being logged`; there is only one `catch` statement in this entire class.  I added logging to that one catch.

I also note that this particular issue existed previously; it wasn't introduced by nor is it directly related to I239...

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -209,6 +224,8 @@ private void startShadowing() {
             if (success) {
                 currentlyShadowing = true;
                 shadowingStatusValueLabel.setText("ON");
+                getStartStopButton().setText("Stop Shadowing");
+                getStartStopButton().setToolTipText("Stop shadowing operations");
```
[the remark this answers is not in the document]
clevengr: Logging added here.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -476,14 +468,17 @@ private void updateGUI() {
...
 
-        getStartButton().setEnabled(!currentlyShadowing);
-        getStopButton().setEnabled(currentlyShadowing);
+        getStartStopButton().setEnabled(true);
         getUpdateButton().setEnabled(false);
 
         if (currentlyShadowing) {
             shadowingStatusValueLabel.setText("ON");
+            getStartStopButton().setText("Stop shadowing");
+            getStartStopButton().setToolTipText("Stop the currently active shadowing of the remote CCS");
```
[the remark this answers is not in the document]
clevengr: The code to which this comment refers is not (necessarily) a change of state.  This is in method `updateGUI()`, which is invoked every time something in the GUI changes (not just when shadow on/off state changes occur.  The purpose of this code is to insure the GUI text correctly represents the CURRENT state, not to effect a CHANGE of state.  Logging this every time the GUI changes would, I think, cause confusion in the logs.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -476,14 +468,17 @@ private void updateGUI() {
...
         } else {
             shadowingStatusValueLabel.setText("OFF");
+            getStartStopButton().setText("Start shadowing");
+            getStartStopButton().setToolTipText("Start shadowing the currently specified remote CCS");
```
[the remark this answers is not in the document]
clevengr: Same response as above; this code does not represent a change of state.

src/edu/csus/ecs/pc2/shadow/ShadowController.java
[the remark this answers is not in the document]
lane55: I thought that // return scoreboard was about returning a scoreboard class.
Seems that the variable name remoteJson, threw me off.   if the variable
name were remoteScoreboardJson not remoteJson that comment might
have been more clear to me.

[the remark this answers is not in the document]
lane55: no change needed.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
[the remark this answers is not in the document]
lane55: In general - not in this specific class only.   There is no good way to note a general change/improvement throughout the code review.  That is why I stated in general so the idea of improving the code to make
bugs easier to find.
 Be it one or a hundred un-logged exception/stack traces,
they should be logged.   This also avoids the time spent going back through all the code to
make sure that we have logging to help us debug and see state changes.
We should not lose the opportunity to log exceptions, state changes, etc.   We may not
get back to this code for years.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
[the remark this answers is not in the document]
lane55: This change of state should be logged... wherever the state changes.

src/edu/csus/ecs/pc2/shadow/ShadowController.java
```diff
@@ -345,9 +357,160 @@ private IRemoteContestAPIAdapter createRemoteContestAPIAdapter(URL url, String l
...
+     * @return a String containing the PC2 scoreboard JSON.
+     */
+    private String getPC2ScoreboardJSON() {
+        
+        ScoreboardJson sbJsonObject = new ScoreboardJson();
+        String pc2Json ;
+        try {
+            pc2Json = sbJsonObject.createJSON(localContest, log);
+        } catch (JsonProcessingException | IllegalContestState | JAXBException e) {
+            log.warning("Exception creating PC2 scoreboard JSON: " + e.getMessage());
```
lane55, line 424: May or may not output stack trace to log thus,
log.warning("Exception creating PC2 scoreboard JSON: " + e.getMessage(), e);

lane55 (approved):
Approved.

Issue #274, opened 2021-07-05
"Permissions" class does not handle "Edit Permissions" and "Edit Run" properly


clevengr:
**Describe the issue**: 
The `Permission` class has a method `loadDescription()` which initializes a hashtable with the textual descriptions for each permission (that is, for each element of the `Permission.Type` enum).   Currently the method contains the following:
```java
        ...
        hash.put(Type.EDIT_PERMISSIONS, "Edit Permissions");
        hash.put(Type.EDIT_PERMISSIONS, "Edit Runs");
        ...
```
The second line overwrites the first line's entry in the hashtable, with the effect that classes such as `EditAccountPane` which call `Permission.getDescription()` get back the display text "Edit Runs" when they ask for the textual representation of `EDIT_PERMISSIONS`.  

A (very confusing) side effect of this is that when a user wants to enable/disable the `EDIT_RUN` permission on an account, they are highly likely to be tricked into selecting the permission whose display name is shown as `EDIT_RUNS` (not noticing that there is ALSO an entry, further down the scrollable list, named `EDIT_RUN` (singular)).   The result is that the user thinks they have enabled/disabled the EDIT_RUN permission when in fact they have enabled/disabled the EDIT_PERMISSIONS permission.

**To Reproduce**: 
Method 1:
- Start a remote CCS which supplies a CLICS Event Feed for a contest.
- Start a PC2 Server configured with a CDP matching the remote CCS.
- Start an Event Feed client.
- Select "Shadow Mode" on the EF Client, click "Start Shadowing", and the click "Compare Runs".
- Select a run on the Compare Runs grid.
- Click "Resolve Selected", which is an instruction to update the run in PC2 so that its judgement is the same as the judgement specified by the remote CCS.

This will produce the message "You are not authorized to Edit Runs".

- Start a PC2 Admin.
- Select Configure>Accounts.
- Select the Event Feeder client account being used for the Shadow, then click "Edit".
- On the "Permissions" pane, check the "Edit Runs" checkbox, then click "Update".  This should add the "Edit Runs" permission to the Feeder account.
- On the Event Feed Shadow, again select a run and click "Resolve Selected".

**Expected behavior**: 
 Adding the "Edit Runs" permission to the Feeder account should allow the Feeder Shadow to update the run.

**Actual behavior**: 
[...]

PR #277, opened 2021-07-29
I273 shadow resolve runs

### LINKED ISSUE #273: Add support for "resolving" Shadow runs which differ from Primary CCS
**Feature Description**: 
It is quite common that at least some runs which are judged by a PC2 Shadow get assigned a different judgement than that assigned by the remote (Primary) CCS.  Reasons for this can include different execution times on the Shadow Judges vs. the Primary Judges (resulting in a run getting a TLE on one system but not the other); differences in the way systems handle certain languages (for example, Kattis considers a Java program with a package statement which fails to execute because the wrong full-package path wasn't provided to be a "compilation error", while PC2 examines the package structure and supports execution of the program (resulting in a WA, RTE, or some other runtime judgement)); and other reasons.

This means that when the Shadow and Remote Scoreboards are compared (using the facilities added in Issue #239), the scoreboards may not be identical because certain runs have different judgements which affect the scoring outcome.  In such cases it is common to have a discussion with the Primary CCS staff, identify the reason for the discrepancy on a submission-by-submission basis, and (typically) accept the Primary CCS as "authoritative".  This means that it is necessary to go into PC2 (e.g. using the Admin "Edit Run" capability) and change the judgements of non-matching runs so that PC2 is told to use the same judgement as that assigned by the Primary CCS.

The process of updating all these mismatched runs is time-consuming and tedious.  Worse, it almost always comes at or near the very end of a contest, when there is time pressure to "finalize the results", e.g. for publication or for a Closing Ceremony.

The objective of this Issue is to provide automated support for "resolving" selected submissions -- meaning, updating them so that PC2 assigns the same judgement as that assigned by the remote Primary CCS.  (This facility would of course only be used following consultation between the Primary and Shadow system teams.)

- Add the ability on the Shadow "Runs Grid"  (class `ShadowCompareRunsPane`, added in Issue #239) to "select" one or more runs (submissions). Selection could be done by adding a checkbox column, or by utilizing the JTable "Row Selection" capability.
- Add a new button `Resolve Selected Judgements`.  When the `Resolve Selected Judgements` button is pressed, it first prompts with a warning that continuing is going to change the judgements of all selected submissions in PC2 to match those of the Remote Primary CCS.  If the user clicks "OK", the button fetches each selected row in the table and invokes the PC2 Server to update the submission judgement to match that of the Remote Primary CCS.   

Updating the submission in the PC2 Server can be done in a way analogous to the code in class `EditRunPane` (see method `updateRun()`).

**Additional Notes:**
A nice-to-have (but not critical) feature would be that there would appear on the Runs Grid a marker/indication of some kind noting those runs (submissions) whose judgements have been changed by the above process.    Even more powerful would be a way to see the _previous_ value of the judgement (before it was changed).  It's arguable however that these should be filed as separate enhancements/issues, since they are likely to be non-trivial.

clevengr:
  Adds the ability to select one or more runs on the "Shadow Compare Runs" grid and click "Resolve Runs", which (after appropriate prompting/confirmation of intent) updates the selected runs in PC2 so that the assigned judgement in PC2 matches the judgement assigned by the Remote CCS being shadowed.  Also updates the "Overridden?" column in the grid to indicate that the PC2 judgement has been overridden (replaced by the Remote CCS judgement).

### Issue which the PR fixes
  Fixes #273.

 Windows 10, Eclipse 2020-12, Java 1.8.0_201.

- Start a PC2 Server, loading a CDP for a contest which is already configured into a remote CCS.  For example, load the NADC 2021 configuration CDP from the following file:  [cdp.zip](https://github.com/pc2ccs/pc2v9/files/6898161/cdp.zip).  This matches the NADC 21 configuration loaded into Kattis.
- Start a PC2 Admin.  In the Admin:
  - Edit the "Feeder 1" account , adding the "Edit a Run" permission.  Note: do NOT select the "Edit Runs" permission; this is a mistaken name for the "Edit Permissions" permission (see #274).  Check the "Edit a Run" box.
  - Start the contest.
- Start several AutoJudges (5 is recommended), each configured to judge all problems in the contest loaded into the server.
- Start an Event Feed client; login as "ef1".
- Select "Shadow Mode" on the EF client, then click "Start Shadowing".
- After a short delay, click the "Compare Runs" button on the EF client.  This will display a Runs grid, which may be blank initially (this is because it takes a while for the Shadow to pull enough data from the remote CCS to actually get to where there are runs in the Event Feed.)
- Click the "Refresh" button repeatedly, until runs start showing up.
- Click the header in the "Match?" column, to sort the runs by Match status.
- Continue clicking the "Refresh" button and scrolling through the runs, until a NON-Matching run appears (it will be red, and the "Non-matches" status count in the summary at the bottom will become greater than zero).  For the NADC 21 contest, it takes about 200 "matching" runs before the first "non-match" shows up.
- Select one or more "non-matching" runs in the grid, using standard grid multi-select (Ctrl/Shift) as desired (non-matching runs are highlighted in red).
- Click the "Resolve Selected" button, then click "Yes" in answer to the confirmation prompt.

Expected result:  the selected runs will be updated so that their judgement in PC2 matches the judgement specified by the remote system.  This will result in the runs moving to the "Match=yes" status, turning green and being inserted into their proper place in the grid according to whatever the current sort key is for the table.  It will also result in the "Overridden?" column for those runs being changed to "Y", indicating that those runs were changed by the Shadow client.   (To see the original status of the runs, go to the PC2 Admin, select the run, and click "View Judgements".)

### DIFF
--- a/src/edu/csus/ecs/pc2/clics/CLICSJudgementType.java
+++ b/src/edu/csus/ecs/pc2/clics/CLICSJudgementType.java
@@ -28,7 +28,7 @@ import java.util.Map;
  * defined by an instance of the class is one of the CLICS "Big 5" judgement types and for mapping the instance judgement type
  * into a recommended "Big 5" judgement.
  * 
- * The class also provides method {@link #getCLICSAcronym(String)} which accepts an arbitrary judgement string (for example,
+ * The class also provides method {@link #getCLICSAcronymFromDisplayText(String)} which accepts an arbitrary judgement string (for example,
  * "Incorrect Output Format") and returns the corresponding CLICS acronym if such an acronym has been defined (see the CLICS
  * Contest API for all currently-defined strings and their corresponding acronyms).
  * 
@@ -172,6 +172,50 @@ public class CLICSJudgementType {
             return name;
         }
 
+    }
+    
+    /**
+     * An enumeration of all the acronyms which represent correct ("yes", "accepted") submission judgements.
+     * 
+     * @author John Clevenger, PC2 Development Team (pc2@ecs.csus.edu)
+     *
+     */
+    public static enum CLICS_YES_ACRONYM {
+        
+        AC  ("Accepted"),
+        ACY ("Yes"),
+        ACC ("Correct"),
+        APE  ("Accepted - Presentation Error");
+        
+        private final String name;
+
+        private CLICS_YES_ACRONYM(String name) {
+            this.name   = name;
+        }
+        
+        public String getValue() {
+            return name;
+        }
+    }
+    
+    /**
+     * Returns true if the specified {@link CLICS_JUDGEMENT_ACRONYM} represents a "Yes" (Accepted) judgement.
+     * 
+     * @param acronym the acronym to be tested.
+     * 
+     * @return true if the specified acronym represents a "Yes" judgement.
+     */
+    public static boolean isYesAcronym(CLICS_JUDGEMENT_ACRONYM acronym) {
+        
+        String targetName = acronym.name();
+        
+        for (CLICS_YES_ACRONYM nextName : CLICS_YES_ACRONYM.values()) {
+            if (targetName.toString().contentEquals(nextName.toString())) {
+                return true;
+            }
+        }
+        
+        return false;
     }
 
     /**
@@ -422,14 +466,15 @@ public class CLICSJudgementType {
     }
     
     /**
-     * Returns the {@link CLICS_JUDGEMENT_ACRONYM} element which corresponds to the specified text string, or null
-     * if there is no CLICS_JUDGEMENT_ACRONYM whose text string matches the specified text.
+     * Returns the {@link CLICS_JUDGEMENT_ACRONYM} element having a "display name" (that is, a toString() value) 
+     * which corresponds to the specified text string, or null
+     * if there is no CLICS_JUDGEMENT_ACRONYM whose display name string matches the specified text.
      * 
-     * @param text a String giving a judgement message; for example "Wrong Answer"
+     * @param text a String giving a judgement display name message; for example "Wrong Answer"
      * 
      * @return the CLICS_JUDGEMENT_ACRONYM element corresponding to the received text (e.g. CLICS_JUDGEMENT_ACRONYM.WA), or null. 
      */
-    public static CLICS_JUDGEMENT_ACRONYM getCLICSAcronym (String text) {
+    public static CLICS_JUDGEMENT_ACRONYM getCLICSAcronymFromDisplayText (String text) {
           
         for (JudgementMapping mapping : judgementStringMappings) {
             
@@ -441,6 +486,43 @@ public class CLICSJudgementType {
         }
         
         //Text string not found
+        return null;
+    }
+    
+    /**
+     * Returns true if the specified String name corresponds exactly to the declared enum element "name" of a CLICS acronym.
+     */
+    public static boolean isCLICSAcronym(String name) {
+        
+        for (CLICS_JUDGEMENT_ACRONYM clicsAcronym : CLICS_JUDGEMENT_ACRONYM.values()) {
+            String clicsName = clicsAcronym.name();
+            if (name.toLowerCase().contentEquals(clicsName.toLowerCase())) {
+                return true;
+            }
+        }
+        return false;
+    }
+    
+    /**
+     * Returns the element of the CLICS_JUDGEMENT_ACRONYM enum whose declared acronym name (that is, the declared value of the enum element,
+     * not the "display name") matches the specified text, or null if there is no
+     * element of the CLICS_JUDGEMENT_ACRONYM enum with a matching declared name.
+     * 
+     *  For example, if the specified text is "AC", this method will return {@link  edu.csus.ecs.pc2.clics.CLICS_JUDGEMENT_ACRONYM#AC}.
+     *  If the specified text is "Accepted", this method will return null (because there is no enum element whose declared name is "Accepted"
+     *  (even though that happens to be the "display name" associated with the element "AC")).
+     *  
+     *  @return the CLICS_JUDGEMENT_ACRONYM element matching the specified string name, or null if there is no such match.
+     */
+    public static CLICS_JUDGEMENT_ACRONYM getCLICSAcronymFromElementName(String name) {
+        
+        for (CLICS_JUDGEMENT_ACRONYM clicsAcronym : CLICS_JUDGEMENT_ACRONYM.values()) {
+            String clicsName = clicsAcronym.name();
+            if (clicsName.contentEquals(name)) {
+                return clicsAcronym;
+            }
+        }
+        
         return null;
     }
     
--- a/src/edu/csus/ecs/pc2/clics/CLICSJudgementType.java
+++ b/src/edu/csus/ecs/pc2/clics/CLICSJudgementType.java
@@ -175,7 +175,8 @@ public class CLICSJudgementType {
     }
     
     /**
-     * An enumeration of all the acronyms which represent correct ("yes", "accepted") submission judgements.
+     * An enumeration of all the acronyms which represent correct submission judgements (that is, judgements which represent
+     * a 'yes' or 'accepted' condition assigned by a judge).
      * 
      * @author John Clevenger, PC2 Development Team (pc2@ecs.csus.edu)
      *
--- a/src/edu/csus/ecs/pc2/core/report/ExtractPlaybackLoadFilesReport.java
+++ b/src/edu/csus/ecs/pc2/core/report/ExtractPlaybackLoadFilesReport.java
@@ -125,7 +125,7 @@ public class ExtractPlaybackLoadFilesReport implements IReport {
         
         //if we found a judgement string, try to convert it to a corresponding CLICS acronym
         if (result!=null && result.length()>0) {
-            CLICS_JUDGEMENT_ACRONYM acronym = CLICSJudgementType.getCLICSAcronym(result);
+            CLICS_JUDGEMENT_ACRONYM acronym = CLICSJudgementType.getCLICSAcronymFromDisplayText(result);
             if (acronym!=null) {
                 result = acronym.name();
             }
--- a/src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
@@ -10,7 +10,6 @@ import java.util.ArrayList;
 import javax.swing.JButton;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
-import javax.swing.SwingUtilities;
 
 import edu.csus.ecs.pc2.core.IInternalController;
 import edu.csus.ecs.pc2.core.model.Account;
@@ -94,13 +93,6 @@ public class CheckPermissionsPane extends JPanePlugin {
     @Override
     public void setContestAndController(IInternalContest inContest, IInternalController inController) {
         super.setContestAndController(inContest, inController);
-    }
-
-    public void showMessage(final String message) {
-        SwingUtilities.invokeLater(new Runnable() {
-            public void run() {
-            }
-        });
     }
 
 }
--- a/src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
@@ -1,0 +1,106 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.ui;
+
+import java.awt.BorderLayout;
+import java.awt.FlowLayout;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.util.ArrayList;
+
+import javax.swing.JButton;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
+
+import edu.csus.ecs.pc2.core.IInternalController;
+import edu.csus.ecs.pc2.core.model.Account;
+import edu.csus.ecs.pc2.core.model.ClientId;
+import edu.csus.ecs.pc2.core.model.IInternalContest;
+import edu.csus.ecs.pc2.core.security.Permission;
+
+/**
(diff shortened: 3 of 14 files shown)


src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -303,10 +361,372 @@ public void actionPerformed(ActionEvent arg0) {
...
+     * 
+     * @return true if one or more runs were "resolved" (updated in PC2); false if no runs were changed.
+     */
+    private boolean resolveSelectedRuns() {
+        
+        //make sure there are some runs selected
+        if (!runsAreSelected()) {
+            
+            //display error dialog
+            JOptionPane.showMessageDialog(this, "There are no runs selected for resolving.", "No runs selected; nothing to resolve", JOptionPane.ERROR_MESSAGE);
```
lane55, line 439: This is more of an info message rather than an error message.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -303,10 +361,372 @@ public void actionPerformed(ActionEvent arg0) {
...
+                    if (success) {
+                        numRunsUpdated++;
+                    }
+                    
+                } else {
+                    
+                    //there is no known judgement matching the judgement string
+                    JOptionPane.showMessageDialog(this, "Cannot update submission " + submissionId + ": invalid remote judgement: " + remoteJudgement, 
+                                                        "Invalid judgement acronym", JOptionPane.WARNING_MESSAGE);
+                    log.log (Log.INFO, "Attempted to update pending submission: " + submissionId);
```
lane55, line 620: This log message does not contain enough context Update to: 
"Attempted to update pending submission: " + submissionId) + 
" Invalid/Non-CLICS remote judgment acronym " + remoteJudgement

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -303,10 +361,372 @@ public void actionPerformed(ActionEvent arg0) {
...
+                found = true;
+                break;
+            }
+        }
+        
+        //if we didn't find the run, return failure
+        if (!found) {
+            
+            JOptionPane.showMessageDialog(this, "Failed to find submission " + submissionId, "Submission not found", JOptionPane.WARNING_MESSAGE);
+            log.log (Log.WARNING, "Failed to find run to be updated: " + submissionId);
```
lane55, line 663: more log context update to:
"Failed to find run to be updated: " + submissionId +",  submission not found in run list "

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -303,10 +361,372 @@ public void actionPerformed(ActionEvent arg0) {
...
+                //we found a matching PC2 judgement; check if the new judgement is a "yes"
+                boolean solved = CLICSJudgementType.isYesAcronym(newJudgement);
+                
+                //build a new JudgementRecord for PC2 containing the desired judgement values
+                JudgementRecord judgementRecord = new JudgementRecord(judgement.getElementId(), getContest().getClientId(), solved, false);
+
+                //duplicate the existing RunResultFiles, with null executionData (since we haven't actually re-executed the run)
+                RunResultFiles runResultFiles = new RunResultFiles(targetRun, targetRun.getProblemId(), judgementRecord, null);
+                
+                log.log(Log.INFO, "Sending new JudgementRecord to PC2 server: "
```
lane55, line 683: This message does not show the run info/id update last statement
+ " isSolved=" + judgementRecord.isSolved() 
+ " run " + run);
If we were to actually spin off a thread for each update then we'd need to know which
run the JudgementRecord is for.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -303,10 +361,372 @@ public void actionPerformed(ActionEvent arg0) {
...
+        for (int submissionId : submissionIdMap.keySet()) {
+                        
+            //make sure the run is not "Pending" in either system
+            Integer modelRow = submissionIdMap.get(submissionId);
+            String pc2Judgement = (String) resultsTable.getModel().getValueAt(modelRow, 4);      // 4 = PC2 Judgment
+            String remoteJudgement = (String) resultsTable.getModel().getValueAt(modelRow, 5);   // 5=remote CCS judgement
+            
+            if (pc2Judgement.toLowerCase().contains("pending") || remoteJudgement.toLowerCase().contains("pending")) {
+                
+                JOptionPane.showMessageDialog(this, "Cannot update submission " + submissionId + " because it still has judgements Pending",
```
lane55, line 594: This message dialog is in a loop, consider if there are 100 run judgements
that cannot be updated the user will have to click ok 100 times.
If this is not likely, then no change is needed.   

Wanted to note that *in general* showing a message dialog in a loop is a bad
idea - but if the likelihood is low that there will be more than a couple
times this dialog appears - then it is ok.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -333,8 +753,7 @@ private void saveCSVFile() {
                     try {
                         saveFile.createNewFile();
                     } catch (IOException e) {
-                        // TODO Auto-generated catch block
-                        e.printStackTrace();
+                        log.log(Log.SEVERE, "Exception saving file: " + e.getMessage(), e);
```
lane55, line 756: more context in log information that answers: which files was not saved?
Update message to:
"Exception saving judgement comparison CSV file : " + e.getMessage(), e);

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -415,4 +836,51 @@ private String getCSVString(ShadowJudgementInfo info) {
...
+
+        public void runChanged(RunEvent event) {
+
+            if (event != null) {
+                
+                //get the run for which we have received a server "changed" notification
+                Run updatedRun = event.getRun();
+                
+                //check if the run update was for the run we requested the server to update
+                if (updatedRun.getElementId().equals(runWeRequestedServerToUpdate.getElementId())) {
```
lane55, line 862: This seems odd, not scalable.   Should the requested run be a Map rather than
a single variable (run) ?

Will there ever be the case where there is more than one resolved run/judgement at a time?

Note: if this is not scalable then a logoff and login would be needed to update the data as a workaround

lane55 (changes requested):

src/edu/csus/ecs/pc2/core/report/ExtractPlaybackLoadFilesReport.java
```diff
@@ -125,7 +125,7 @@ private void writeJudgementRecord(PrintWriter printWriter, Run run, JudgementRec
         
         //if we found a judgement string, try to convert it to a corresponding CLICS acronym
         if (result!=null && result.length()>0) {
-            CLICS_JUDGEMENT_ACRONYM acronym = CLICSJudgementType.getCLICSAcronym(result);
+            CLICS_JUDGEMENT_ACRONYM acronym = CLICSJudgementType.getCLICSAcronymFromDisplayText(result);
```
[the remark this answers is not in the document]
clevengr: No; this is a bug fix that was necessary to implement Resolve Run Judgements.  The Resolve Run Judgements function was failing because of this issue.

src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public void setContestAndController(IInternalContest inContest, IInternalController inController) {
+        super.setContestAndController(inContest, inController);
+    }
+
+    public void showMessage(final String message) {
+        SwingUtilities.invokeLater(new Runnable() {
+            public void run() {
+            }
```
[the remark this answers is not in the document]
clevengr: Well yes, yes it was.  But that wasn't really the problem; the problem was that the `showMessage()` method itself is never used (hence it didn't matter whether it had an empty `run()` body).  The method is now gone.

src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
```diff
@@ -0,0 +1,106 @@
...
+import edu.csus.ecs.pc2.core.model.IInternalContest;
+import edu.csus.ecs.pc2.core.security.Permission;
+
+/**
+ * A UI that provides a button to display the PC2 {@link Permission}s associated with the currently logged-in account.
+ * 
+ * 
+ * @author John Clevenger, PC^2 Development Team (pc2@ecs.csus.edu).
+ */
+public class CheckPermissionsPane extends JPanePlugin {
```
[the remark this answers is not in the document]
clevengr: Well, sure, if you like that better.  I made that change.  The problem now is that there exists a class named "DisplayPermissionsPane()" that, when you open the pane, doesn't actually display any permissions at that time (you have to hit a button to get it to display the permissions).    But I don't care that much, so I renamed it as you suggested.

src/edu/csus/ecs/pc2/ui/CheckPermissionsPane.java
```diff
@@ -0,0 +1,106 @@
...
+    public String getPluginTitle() {
+        return "CheckPermissions Pane";
+    }
+
+    @Override
+    public void setContestAndController(IInternalContest inContest, IInternalController inController) {
+        super.setContestAndController(inContest, inController);
+    }
+
+    public void showMessage(final String message) {
```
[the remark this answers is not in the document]
clevengr: Done.

src/edu/csus/ecs/pc2/ui/DevelopmentPane.java
```diff
@@ -56,7 +58,9 @@ public DevelopmentPane() {
 
         tabbedPane.addTab("Options", null, optionsPane, null);
         
-    }
+        tabbedPane.addTab("Check Permissions", null, checkPermissionsPane, null);
```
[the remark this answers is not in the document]
clevengr: Again, I think it's slightly confusing if you select a pane named "Display permissions" and it isn't displaying any permissions.  But I'm fine if you want to change the title to match the new class name.

src/edu/csus/ecs/pc2/clics/CLICSJudgementType.java
```diff
@@ -173,6 +173,50 @@ public String getValue() {
...
+     * @param acronym the acronym to be tested.
+     * 
+     * @return true if the specified acronym represents a "Yes" judgement.
+     */
+    public static boolean isYesAcronym(CLICS_JUDGEMENT_ACRONYM acronym) {
+        
+        String targetName = acronym.name();
+        
+        for (CLICS_YES_ACRONYM nextName : CLICS_YES_ACRONYM.values()) {
+            if (targetName.toString().contentEquals(nextName.toString())) {
```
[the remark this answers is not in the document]
clevengr: There is no equivalent to `String.equalsIgnoreCase()` for `String.contentEquals()` (that is, there's no method like `contentEqualsIgnoreCase()`.  This means either the code would have to use the less robust `String.equalsIgnoreCase()`, or would have to add some heavy additional logic (like using Pattern matching) to accomplish this same thing.  I opted for the more robust `contentEquals()` based on the fact that what is being checked are elements of the iterator over `CLICS_YES_ACRONYM.values()`, which returns the literal declared value of each element of the enum, which are always upper-case.  

I don't know what javadoc you are talking about that shows lower case, unless you are talking about the statement
```java
An enumeration of all the acronyms which represent correct ("yes", "accepted") submission judgements.
 ```
That statement is not a list of acronym values (as evidenced by the fact that the strings in parentheses are in ALL lower-case, whereas the acronym values are all strictly upper-case and the acronym names are CamelCase.

Assuming it was that javadoc statement you are referring to, I have updated the javadoc to make it clearer what the intent of the comment was.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
clevengr: Done.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
clevengr: Done.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
clevengr: At this point in the code it is processing exactly (only) the runs which the user has selected.  The only way the "100 times" scenario could happen in this case is if the user selected 100 _pending_ runs and told the system to resolve them.  If they do that, I think they DESERVE to have to click OK 100 times -- it will teach them not to try to resolve Pending (unjudged) runs.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
clevengr: In the current implementation, it is processing each update request separately, sending a message (packet) to the server requesting that the current run be updated, then waiting for the server to send back a response.  Since we typically only have a few (or perhaps a few dozen) runs to be resolved, and they are almost always done either one-at-a-time (as the Shadow and Primary teams decide what resolution is required based on the specific case in question) there's not much chance of a scalability issue; the only things that ever have to be resolved are runs where the Shadow and the Primary disagree (which tend to be small in number and have to be resolved individually or in small groups).

We could certainly try making it more efficient by supporting batch handling of multiple updates -- but again, such batches don't happen very often and tend to be small.  I do not have the time to work on what I consider an unnecessary refinement right now, given the pressure of getting this working prior to the upcoming contests.  Feel free to file a separate Technical Deficiency Issue for this if you think it's that important.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
clevengr: Done.

lane55 (approved):
Approved.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
lane55: No need to optimize at this time.  Since the batches will be small, the need to be scalable is moot.

PR #280, opened 2021-08-05
i279: fix unit test failure

### LINKED ISSUE #279: Unit test failure stops build, unit test method testcreateContestStandings
**Describe the issue**: 

Unit test is failing.  ScoreboardJsonTest.testcreateContestStandings fails.

java.io.FileNotFoundException under Unix.
Change \ in unit test filename to / to fix.

**To Reproduce**: 

Run the JUnit under Unix/Linux.

**Expected behavior**: 

Unit test during build ScoreboardJsonTest.testcreateContestStandings passes.

**Actual behavior**: 

Unit test during build ScoreboardJsonTest.testcreateContestStandings fails.

From gitlab build 

Testcase:
testcreateContestStandings(edu.csus.ecs.pc2.services.core.ScoreboardJsonTest):
Caused an ERROR
[junit] null
[junit] javax.xml.bind.UnmarshalException
[junit] - with linked exception:
[junit] [java.io.FileNotFoundException:
/builds/pc2ccs/pc2v9/testdata\ScoreboardJSONTest\testNADC21Prac2EF\nadcPractice-event-feed.xml
(No such file or directory)]

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

lane55:

Fixes java.io.FileNotFoundException under Unix so build will complete unit tests without error.
Specific fix : change \ in unit test filename to /

### Issue which the PR fixes

#279 

Windows with target of Unix

merge this change into develop branch to trigger new build
and unit tests.

### DIFF
--- a/test/edu/csus/ecs/pc2/services/core/ScoreboardJsonTest.java
+++ b/test/edu/csus/ecs/pc2/services/core/ScoreboardJsonTest.java
@@ -134,7 +134,7 @@ public class ScoreboardJsonTest extends AbstractTestCase {
         /**
          * pc2 Standings XML for NADC Practice 2.
          */
-        String xmlFilename = "testdata\\ScoreboardJSONTest\\testNADC21Prac2EF\\nadcPractice-event-feed.xml";
+        String xmlFilename = "testdata/ScoreboardJSONTest/testNADC21Prac2EF/nadcPractice-event-feed.xml";
 
         ContestStandings contestStandings = ScoreboardUtilites.createContestStandings(new File(xmlFilename));
 


clevengr:
I actually think it would be better to use File.separator instead of hard-coding "/".  The behavior of accepting "/" for "\" seems to me to be a Windows-specific thing (whether it's in the Windows JVM or in the NT API, I'm not sure).  It seems reasonable to expect that this behavior might not _always_ work in all JVMs on all platforms, whereas it seems reasonable to assume that use of File.separator WILL always work because that's precisely what it exists for.   But I'm going to go ahead and approve the PR just so we can fix the build problems.

clevengr (approved):

PR #316, opened 2022-02-23
Add some divisionRank support

### LINKED ISSUE #315: Add support for proper Overall Division Standings to Scoreboard
**Describe the issue**: 
Currently the XML code generated by the Scoreboard does not fully support display of accurate "Division Standings".
Divisions are implemented as collections of "Groups".  For example, if there are two "Divisions" (say, D1 and D2) in a contest with three "Sites" (say, California, Oregon, Washington), this is supported by defining six "Groups":  CalD1, CalD2, OreD1, OreD2, WashD1, and WashD2.  The combination of XML and XSL used by the Scoreboard HTML allows for generating pages showing "Overall Division Standings" (e.g. "D1 Overall Standings" and "D2 Overall Standings") and also pages showing "Per-Site" (which is really "Per-Group") standings (e.g. "CalD1 Standings" or "WashD2 Standings").   

HOWEVER, the _rankings_ shown on the "Overall Division Standings" pages are not correct:  they show the current rank for a team _in the entire contest_, not the rank _within their Division_.  For example, if in the overall contest a team in D2 has currently solved a number of problems which give them rank "2" _overall_ (contest-wide), then looking at the D1 Overall Standings will show the top D1 team in Rank 1, followed by the next D1 team as being in rank **_3_**.

The reason for this is apparently that the Default Scoring Algorithm was updated long ago to output into XML the values for "rankOverall" and "rankInGroup" for each team, but it doesn't have any code to include "rankInDivision" (which would be calculated as "rank in the set of all Groups in the same Division).

**To Reproduce**: 
- Define a contest with 12 teams split into two Divisions and two sites.  That is, define FOUR "Groups":  Site1D1, Site2D1, Site1D2, and Site2D2, and assign three teams to each of these four Groups.
- Add at least one problem identified as belonging to each "Division" (i.e. one problem assigned solely to D1 (so, to Groups Site1D1 and Site2D1) and another problem assigned solely to D2 (so, assigned to Groups Site1D2 and Site2D2).
- Submit a correct ("Yes") solution for one team in D1.
- Submit a later correct solution for one team in D2.
- Submit a later correct solution for another team in D1.
- Start a Scoreboard and open the HTML page generated by the Scoreboard.

**Expected behavior**:
The two teams with "yes" solutions in D1 should be ranked 1 and 2 in the D1 Overall Standings; the one team with a "yes" in D2 should be ranked 1 in the "D2 Overall Standings".

**Actual behavior**: 
The two teams in D1 will have ranks "1" and "3" (because the team in D2 is currently occupying the "2" overall rank).
The one team that should be ranked "1" in D2 will have rank "2" (because the Overall standings are showing contest-wide standings, not per-division standings, and the team currently in rank "1" is in a different Division).

**Possible Solution**
- Add code to the DSA, independent of any other code, to compute, for each team, a value "rankInDivision" which is computed by finding the team's current rank among all teams in the same Division -- that is among all teams in any Group which is part of the same  Division.
- Output the "rankInDivision" value in the ContestStandings XML entry for each team.
- Update the Scoreboard XSL for the Overall Division Standings page(s) to use the "rankInDivision" value instead of the "rankInContest" value.

troy2914:
This depends on the group displayName ending with D\d+

fixes #315

btw, also added an overalRank (== rank)

### DIFF
--- a/src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
+++ b/src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
@@ -14,6 +14,8 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.Vector;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import edu.csus.ecs.pc2.VersionInfo;
 import edu.csus.ecs.pc2.core.PermissionGroup;
@@ -579,7 +581,7 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
         Hashtable<Group, Integer> groupIndexHash = new Hashtable<Group, Integer>();
         int groupCount = 0;
         for (Group group : groups) {
-            // no refence to groups that should not be displayed on scoreboard
+            // no reference to groups that should not be displayed on scoreboard
             if (!group.isDisplayOnScoreboard()) {
                 continue;
             }
@@ -587,7 +589,26 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
             groupIndexHash.put(group, Integer.valueOf(groupCount));
             groupCount++;
         }
-        
+        Hashtable<Group, Integer> divisionIndexHash = new Hashtable<Group, Integer>();
+        int divisionCount = 0;
+        // TODO this bit should in the future can probably go away when divisions are supported better.
+        int lastFound = 0;
+        for (Group group : groups) {
+            // no reference to groups that should not be displayed on scoreboard
+            if (!group.isDisplayOnScoreboard()) {
+                continue;
+            }
+            Pattern pattern = Pattern.compile("D\\d+$");
+            Matcher matcher = pattern.matcher(group.getDisplayName());
+            if (matcher.find()) {
+                Integer found = Integer.parseInt(matcher.group(0).substring(1));
+                if (found.intValue() > lastFound) {
+                    lastFound = found.intValue();
+                }
+                divisionIndexHash.put(group, found);
+            }
+        }
+        divisionCount = lastFound;
         String teamVarDisplayString = contestInformation.getTeamScoreboardDisplayFormat();
         
         StandingsRecord[] srArray = new StandingsRecord[treeMap.size()];
@@ -626,6 +647,20 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
             groupRank[i] = 0;
             groupIndexRank[i] = 0;
         }
+        long[] divisionNumSolved = new long[divisionCount];
+        for (int i = 0; i < divisionCount; i++) {
+            divisionNumSolved[i] = -1;
+        }
+        int[] divisionRank = new int[divisionCount];
+        int[] divisionIndexRank = new int[divisionCount];
+        long[] divisionScore = new long[divisionCount];
+        long[] divisionLastSolved = new long[divisionCount];
+        for (int i = 0; i < divisionCount; i++) {
+            divisionRank[i] = 0;
+            divisionIndexRank[i] = 0;
+            divisionScore[i] = 0;
+            divisionLastSolved[i] = 0;
+        }
         while (iterator.hasNext()) {
             Object o = iterator.next();
             StandingsRecord standingsRecord = (StandingsRecord) o;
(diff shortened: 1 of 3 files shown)


src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -579,15 +581,34 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
...
             groupHash.put(group.getElementId(), group);
             groupIndexHash.put(group, Integer.valueOf(groupCount));
             groupCount++;
         }
-        
+        Hashtable<Group, Integer> divisionIndexHash = new Hashtable<Group, Integer>();
+        int divisionCount = 0;
+        // TODO this bit should in the future can probably go away when divisions are supported better.
+        int lastFound = 0;
+        for (Group group : groups) {
```
clevengr, line 596: Add comment:  
//for every group to be displayed, check if the Group name ends in "D" or "d" followed by a digit.
//If so, add the Group to a hashtable with the Group as the key and the Group number (digit) as the value

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -579,15 +581,34 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
...
+            // no reference to groups that should not be displayed on scoreboard
+            if (!group.isDisplayOnScoreboard()) {
+                continue;
+            }
+            Pattern pattern = Pattern.compile("D\\d+$");
+            Matcher matcher = pattern.matcher(group.getDisplayName());
+            if (matcher.find()) {
```
clevengr, line 603: //get the last character (the digit identifying the Group's Division)

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -579,15 +581,34 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
...
+                Integer found = Integer.parseInt(matcher.group(0).substring(1));
+                if (found.intValue() > lastFound) {
+                    lastFound = found.intValue();
+                }
```
clevengr, line 607: //insert the current Group into the hashtable with its Division number as the value

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -579,15 +581,34 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
...
+                divisionIndexHash.put(group, found);
+            }
+        }
```
clevengr, line 610: //"lastFound" (which is really "highestFound", not necessarily the "last" found) is the highest Division number; save it.
// Note that this is merely the maximum POSSIBLE number of Divisions; if the Division digits skipped some value
// (e.g. if D1 and D3 were defined but not D2), divisionCount will actually be greater than the actual number of divisions).

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -626,6 +647,20 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
             groupRank[i] = 0;
             groupIndexRank[i] = 0;
         }
+        long[] divisionNumSolved = new long[divisionCount];
+        for (int i = 0; i < divisionCount; i++) {
```
clevengr, line 651: The use of "long" here seems to imply that "int" wouldn't be enough -- but that would imply that there could be more than 2^32 solutions in a Division.  Seems like overkill...

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -649,6 +684,7 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
             standingsRecordMemento.putLong("points", standingsRecord.getPenaltyPoints());
             standingsRecordMemento.putInteger("solved", standingsRecord.getNumberSolved());
             standingsRecordMemento.putInteger("rank", standingsRecord.getRankNumber());
+            standingsRecordMemento.putInteger("overalRank", standingsRecord.getRankNumber());
```
clevengr, line 687: "overalRank" should be spelled "overallRank"...

clevengr (approved):

lane55 (approved):
Approved.

Build new dist from fork, tested with PacNW and updated xsl, that fixed the issue.
tirvial new tag overalRank mispelled should be overallRank (two L's)

Issue #295, opened 2022-01-18
Add support for per-problem memory limits and MLE judgement


clevengr:
**Feature Description**: 
Each contest problem should have a "memory limit" associated with it; when a submission for the problem is executed the system should enforce this memory limit in a manner similar to how it currently enforces a time limit. This would allow support for returning a "MLE" (Memory Limit Exceeded) submission result.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** 
- Add a "memory limit" field to class **core.model.Problem**.  Memory limit should probably be specified in MB (so a 2GB limit would be specified as "2048").
- Update the **Configure Contest > Settings** tab to allow specification of a contest-wide (global) memory limit (in MB).  This is the limit which will apply to any problem which does not explicitly set its own problem-specific limit.
- Update the **Configure Contest > Problems** tab to include display of the current memory limit for each problem.
- Update the **General** tab in the **EditProblemPane** class to support a new field, "Memory Limit (MB)".  The value in this field wil become the memory limit for the specified problem.
- Update the YAML loading routine(s) to support loading both a contest-wide and a problem-specific memory limit (in MB).
  - The contest-wide memory setting would appear in **contest.yaml** and should apply unless it is explicitly overridden by a memory limit in a specific **problem.yaml** file.
- Provide support for ability to specify an (OS-specific) _sandbox_ which is to be used to execute each test case of a submission.
  - The sandbox should enforce the problem-specified memory limit (and probably also the time limit) in an OS-specific way.
  - The sandbox should be user-selectable on a per-problem basis.

lane55:
requirements change now both sandbox command and memory limit are per problem, in the problem.yaml file.
There is no global memory limit

johnbrvc:
After doing some research on how to deal with MLE's in the PC2 sandbox, it seems there are 2 cases:
1) java and kotlin (kotlin is really java)
2) C++/C/Python

The 1st case is handled by the JVM and the "ulimit -d" method can not be used since java uses the same Linux address space as the code it is interpreting.   This can be (and already is) handled by the command lines options to the java execute command, namely, the  -Xmx### option, where ### is the maximum heap size for the JVM to allow for the user program.  Something like: -Xmx2048m would set the limit to 2GB.  

[...]

PR #347, opened 2022-03-10
I 295 sandbox util work

lane55:

Adds classes for reading and writing sandbox property files.

### Issue which the PR fixes

none, it is partial work for #295 

All.

Run/review  JUnits: SandboxUtilitiesTest and SandboxResultsTest

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/SandboxInput.java
+++ b/src/edu/csus/ecs/pc2/core/execute/SandboxInput.java
@@ -15,11 +15,22 @@ import java.util.Properties;
  */
 public class SandboxInput {
 
-    private static final String SANDBOX_PROPERTIES_FILENAME = "sandbox.properties";
+    /**
+     * Judges data file name, aka sumit.in
+     */
+    public static final String DATAFILENAME_KEY = "datafilename";
+    
+    /**
+     * Source file name, aka Sumit.java
+     */
+    public static final String SOURCEFILENAME = "sourcefilename";
+
+    /**
+     * Name of sandbox settings input file.
+     */
+    public static final String SANDBOX_PROPERTIES_FILENAME = "sandbox.properties";
 
     public static final String TIMELIMIT_KEY = "timelimit";
-
-    public static final String INPUTFILENAME_KEY = "inputfilename";
 
     public static final String MEMLIMIT_KEY = "memlimit";
 
@@ -28,9 +39,7 @@ public class SandboxInput {
     public static final String PROBLEM_LETTER_KEY = "problemLetter";
 
     public static final String RUNID_KEY = "runid";
-
-    private Properties inputProperties = new Properties();
-
+    
     private String outputPropertiesFilename = SANDBOX_PROPERTIES_FILENAME;
 
     /**
@@ -41,15 +50,15 @@ public class SandboxInput {
      * @return
      * @throws IOException
      */
-    public String store(File outputDirectory) throws IOException {
+    public String store(File outputDirectory, Properties sandboxProperties) throws IOException {
 
         String filename = SANDBOX_PROPERTIES_FILENAME;
         if (outputDirectory != null) {
-            filename = outputDirectory.getAbsolutePath() + File.pathSeparator + SANDBOX_PROPERTIES_FILENAME;
+            filename = outputDirectory.getAbsolutePath() + File.separator + SANDBOX_PROPERTIES_FILENAME;
         }
 
         FileOutputStream fileOutputStream = new FileOutputStream(filename, false);
-        inputProperties.store(fileOutputStream, "PC^2 Sandbox Input properties on " + new Date());
+        sandboxProperties.store(fileOutputStream, "PC^2 Sandbox Input properties on " + new Date());
 
         outputPropertiesFilename = filename;
         return outputPropertiesFilename;
--- a/src/edu/csus/ecs/pc2/core/execute/SandboxInput.java
+++ b/src/edu/csus/ecs/pc2/core/execute/SandboxInput.java
@@ -1,0 +1,66 @@
+// Copyright (C) 1989-2022 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.core.execute;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Date;
+import java.util.Properties;
+
+/**
+ * Sandbox Input
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class SandboxInput {
+
+    private static final String SANDBOX_PROPERTIES_FILENAME = "sandbox.properties";
+
+    public static final String TIMELIMIT_KEY = "timelimit";
+
+    public static final String INPUTFILENAME_KEY = "inputfilename";
+
+    public static final String MEMLIMIT_KEY = "memlimit";
+
+    public static final String LANGUAGE_KEY = "language";
+
+    public static final String PROBLEM_LETTER_KEY = "problemLetter";
+
+    public static final String RUNID_KEY = "runid";
+
+    private Properties inputProperties = new Properties();
+
+    private String outputPropertiesFilename = SANDBOX_PROPERTIES_FILENAME;
+
+    /**
+     * Store input sandbox properties.
+     * 
+     * @param outputDirectory
+     *            the output directory, use execute directory
+     * @return
+     * @throws IOException
+     */
+    public String store(File outputDirectory) throws IOException {
+
+        String filename = SANDBOX_PROPERTIES_FILENAME;
+        if (outputDirectory != null) {
+            filename = outputDirectory.getAbsolutePath() + File.pathSeparator + SANDBOX_PROPERTIES_FILENAME;
+        }
+
+        FileOutputStream fileOutputStream = new FileOutputStream(filename, false);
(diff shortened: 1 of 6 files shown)


src/edu/csus/ecs/pc2/core/execute/SandboxInput.java
```diff
@@ -15,11 +15,22 @@
...
 
-    private static final String SANDBOX_PROPERTIES_FILENAME = "sandbox.properties";
+    /**
+     * Judges data file name, aka sumit.in
+     */
+    public static final String DATAFILENAME_KEY = "datafilename";
+    
+    /**
+     * Source file name, aka Sumit.java
+     */
```
clevengr, line 25: This is a comment regarding a very very very minor issue, BUT,  in the spirit of accuracy:  "aka" means "Also Known As".  So what the above comment says is that "This is the source file name, also known as Sumit.java".  I think what is meant is "e.g." (short for "exempli gratia") which means "For example".

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
```diff
@@ -7,14 +7,20 @@
 import java.util.Properties;
 
 /**
- * Sandbox Results loader
+ * Sandbox Results loader.
  *
```
clevengr, line 11: The comment indicates that this class is a "loader"... yet the NAME of the class implies that it is an encapsulation of "results".  At a minimum, the header comments should provide enough additional information to make it clear why this is a "loader" when it appears to be a "Results".  Better, the class name should be changed to match the class function if necessary.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
```diff
@@ -33,8 +39,14 @@ public SandboxResults(File propertiesFilename) throws IOException {
         resultsProperties.load(fileInputStream);
     }
 
-    public String getProperty(String s) {
```
clevengr: For some reason GitHub will not let me comment on certain lines in this file (probably because they have not been changed IN THIS VERSION).  This comment is about the `public SandboxResults()` constructor, above.
The existence of this no-arg constructor allows a developer to construct a SandboxResults object that has no results in it -- which is an object that exists in what is called an Invalid State.  This is in general, bad design; it should not ever be possible to have an object which is in an Invalid State.  (Yes, I know it's possible all over the existing code; that doesn't mean we should allow propagation of bad design).  This constructor should be eliminated; or, if it is required for some reason (like an IDE complaining if there's no no-arg constructor), it should be made private.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
```diff
@@ -33,8 +39,14 @@ public SandboxResults(File propertiesFilename) throws IOException {
...
-        return resultsProperties.getProperty(s);
+    /**
+     * Return value for property from sandbox file.
+     * 
+     * @param key
+     * @return null or key value.
+     */
```
clevengr, line 47: This method doesn't return the "key value", it returns the "value stored under the specified key".  There's a difference.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
```diff
@@ -62,5 +74,9 @@ public int getExitCode() {
             }
         }
     }
+
```
clevengr, line 77: Again, problems with GitHub's commenting rules.  This comment is about line 65 (above):  `Object obj = resultsProperties.get("exitcode").`
With all the attention given elsewhere to using constants to define hash keys, it seems at best inconsistent that a hard-coded string ("exitcode") is used here.

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
```diff
@@ -10,7 +10,7 @@
 import edu.csus.ecs.pc2.core.model.Run;
 
 /**
- * 
+ * Input to Sandbox. 
  *
```
clevengr, line 14: It seems inaccurate that the comment here is "Input to Sandbox" when the class is "SandboxUtilities".  A "Utilities" class is in general quite different from something which provides input.  What if the Utilities are later expanded to produce output, for example?

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
```diff
@@ -23,28 +23,35 @@
      *
```
clevengr, line 23: The first line of JavaDoc comments should end in a period.  More importantly, it's not at all clear what the comment means -- what does it mean to "write input"?  It would be quite helpful if there was some kind of explanation here...

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
```diff
@@ -23,28 +23,35 @@
...
+    public static String writeSandboxInputProperties(String executeDirectory, Run run, String sourceFileName, Problem problem, Language language, File judgesDataFile, int memlimitK, int timelimitMS) throws IOException {
 
         Properties properties = new Properties();
 
         properties.put(SandboxInput.RUNID_KEY, Integer.toString(run.getNumber()));
 
-        properties.put(SandboxInput.MEMLIMIT_KEY, inputFile);
-        properties.put(SandboxInput.TIMELIMIT_KEY, timelimitMS);
+        if (memlimitK > 0) {
+            properties.put(SandboxInput.MEMLIMIT_KEY, Integer.toString(memlimitK));
```
clevengr, line 42: Memory limits are almost always expressed these days in Mb, not Kb.  This variable name is either very misleading (if the actual data is really a value in Mb), or else the intent is bad -- that is, if the intent really is to force people to specify memory limits in units of Kb.  This would mean for example that if I want to specify a memory limit of 4GB I would have to write it as 4096000.  There is a reason the CLICS spec specifies memory limit values as being Mb, and I think it's a good one that we should stick to.

test/edu/csus/ecs/pc2/core/execute/SandboxResultsTest.java
```diff
@@ -0,0 +1,73 @@
...
+        
+        int count = sandboxResults.getProperties().size();
+        assertEquals("Expected number of properties in "+propFilename, 4, count);
+        
+        
+//        String keyName = SandboxInput.MEMLIMIT_KEY;
+//        String str = props.getProperty(keyName);
+//        int value = Integer.parseInt(str);
+//        assertEquals("For run "+run.getNumber()+" expecting value for key "+keyName, timeLimit, value);
+
```
clevengr, line 49: Considering that this entire code base is supposed to be for supporting memory limits, it's very surprising to find the MEM_LIMIT_KEY test commented out... why?

clevengr (commented):
I have reviewed all the files in this PR and made suggestions where I thought it appropriate.  However, the contents of this PR are not something I'm ready to approve for merging, both because they are incomplete in terms of the Issue which the PR is targeting (which I understand was already understood), but more importantly because I'm still not convinced that adding the complexity of dealing with these Sandbox Properties files is the correct path we should be taking regarding providing a solution for implementing MLE support.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
lane55: Fixed.  Please state an action to take, like "Remove public SandboxResults() or make it private" then
go into an explanation. thanks.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
lane55: typo, fixed parameter doc (which is less work than class or method doc), which may be done later.

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
lane55: Agreed. Added and now using
public static final String EXITCODE_KEY = "exitcode";

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
lane55: Will remove comment and add doc later, added TODO

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
lane55: no change, will do doc improvements later.   This is prototype code not final code.

src/edu/csus/ecs/pc2/core/execute/SandboxUtilities.java
lane55: changed to memlimitM

src/edu/csus/ecs/pc2/core/execute/SandboxResults.java
lane55: not doing doc at this time.   Once a design has been decided on will do doc work.
This PR is not intended to be merged to develop as described in a previous email.

clevengr (commented):
I understand the notion "this is prototype code, so not doing documentation at this time; will fix later".  However, the problem is that if someone (me) is expected to review the code, it is MUCH harder to review the code when there is nothing explaining, at least in general, what each class/method is intended to do.

test/edu/csus/ecs/pc2/core/execute/SandboxResultsTest.java
lane55: Prototyping PR/code to not be used.
closing PR

lane55:
This is a prototype PR as indicated in an earlier email.

PR #352, opened 2022-03-15
I 335 add option for raw report contents export

### LINKED ISSUE #335: For Reports add a "bare" option that will not output header and footer

Yes.   For a number of reports the header and footer
are extra info and must be removed to use the data/file.

**Feature Description**: 

Provide a way to create reports wihtout headers/footers, a "raw" report.

Yes, hand editing reports

On Reports tab

1. Create a checkbox "Bare" that if checked will not output header and footer

2. Add a button "Export" which will create a bare report and put is into a viewer
just like View does

For Reports CLI

1.  Add a --bare option that if checked will not output header and footer

**Additional context**:

lane55:

In the PacNW contest the JSON had to be stripped of a standard
pc2 report header and footer.   This feature providea a way to
avoid that extra step.

Adds a way to output report content without header and footers ("raw" content output).
The pc2reports has a new option --raw to support the new feature
On Reports tab added a new Export Reports Contents button 

### Issue which the PR fixes

#335 

All.

On Reports tab (server or admin) select: Event Feed JSON

Create a report file using Reports tab "Export Reports Contents" button 

Expected

No  header or footer output

### DIFF
--- a/src/edu/csus/ecs/pc2/core/report/Reports.java
+++ b/src/edu/csus/ecs/pc2/core/report/Reports.java
@@ -5,6 +5,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.PrintWriter;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Vector;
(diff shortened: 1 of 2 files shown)


clevengr (approved):
Looks good.  Approved.

PR #358, opened 2022-03-20
I 295 new executable and yaml load

lane55:

Creates yaml with mem limit and sandbox variables:  memory-limit-in-Meg, sandbox-type,
     sandbox-command-line, sandbox-program-name

\# in problem.yaml  (added example in samps/contests/sumitMTC/config/sumit/problem.yaml)
limits:
    timeout: 10
    memory-limit-in-Meg: 2048

sandbox-type: NONE
sandbox-command-line: '{:sandbox} {:memlimit} {:timelimit}'
sandbox-program-name: 'pc2sandbox.sh

Loads yaml with mem limit and sandbox variables
 
In Executable substitute mem limit and sandbox variables

Each of these are substituted based on the run's problem

{:memlimit} for int getMemoryLimitMB() 
{:sandboxcommandline} for String getSandboxCmdLine() 
{:sandboxprogramname} for String getSandboxProgramName() 

In Problems report new variables are now present.

### Issue which the PR fixes

partial #295

java version "1.8.0_121"

Use sumitMTC (samps/contests/sumitMTC/config/sumit/problem.yaml)
to test loading new variables.

Add a new problem and use Export YAML to create new problem.yaml with new variables.

Test the load/set variables by either
1 - Export - conetst.yaml and view the problem.yaml
2 - on Problem click the Report and search for Sandbox and Mem limit
  
**Review values for these lines
  Memory limit          :

  Sandbox Type          : 
  Sandbox program name  : 
  Sandbox command line  :** 

Test executable variable substitutions by loading values from yaml 
then executing the run and reviewing the before and after in the log.

Search for the phrase
**command pattern before substitution:**
and then
**command pattern after substitution:** 
check for the variable substitution

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2022 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.execute;
 
 import java.io.BufferedInputStream;
@@ -2396,7 +2396,8 @@ public class Executable extends Plugin implements IExecutable {
         String afterString = new Integer(afterInt).toString();
         return replaceString(origString, beforeString, afterString);
     }
-
+    
+    
     /**
      * return string with all field variables filled with values.
      * 
@@ -2512,7 +2513,7 @@ public class Executable extends Plugin implements IExecutable {
             } else {
                 log.config("substituteAllStrings() problem is undefined (null)");
             }
-
+            
             if (executionData != null) {
                 if (executionData.getExecuteProgramOutput() != null) {
                     if (executionData.getExecuteProgramOutput().getName() != null) {
@@ -2524,10 +2525,33 @@ public class Executable extends Plugin implements IExecutable {
                 newString = replaceString(newString, "{:exitvalue}", Integer.toString(executionData.getExecuteExitValue()));
                 newString = replaceString(newString, "{:executetime}", Long.toString(executionData.getExecuteTimeMS()));
             }
+            
             String pc2home = new VersionInfo().locateHome();
             if (pc2home != null && pc2home.length() > 0) {
                 newString = replaceString(newString, "{:pc2home}", pc2home);
             }
+            
+            // TODO REFACTOR replace vars with constants for: memlimit, sandboxcommandline,sandboxprogramname
+            newString = replaceString(newString, "{:memlimit}", Integer.toString(problem.getMemoryLimitMB()));
+
+            if (newString.indexOf("{:sandboxcommandline}") > -1) {
+                if (!StringUtilities.isEmpty(problem.getSandboxCmdLine())) {
+
+                    newString = replaceString(newString, "{:sandboxcommandline}", problem.getSandboxCmdLine());
+                    newString = substituteAllStrings(inRun, newString);
+                } else {
+                    // remove {:sandbox} if no sandbox defined in pc2
+                    newString = replaceString(newString, "{:sandboxcommandline}", "");
+                }
+            }
+
+            if (!StringUtilities.isEmpty(problem.getSandboxProgramName())) {
+                newString = replaceString(newString, "{:sandboxprogramname}", problem.getSandboxProgramName());
+            } else {
+                newString = replaceString(newString, "{:sandboxprogramname}", "");
+            }
+             
+
         } catch (Exception e) {
             log.log(Log.CONFIG, "Exception substituting strings ", e);
             // carrying on not required to save exception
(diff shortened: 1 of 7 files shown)


src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2524,10 +2525,33 @@ public String substituteAllStrings(Run inRun, String origString) {
...
+            newString = replaceString(newString, "{:memlimit}", Integer.toString(problem.getMemoryLimitMB()));
+
+            if (newString.indexOf("{:sandboxcommandline}") > -1) {
+                if (!StringUtilities.isEmpty(problem.getSandboxCmdLine())) {
+
+                    newString = replaceString(newString, "{:sandboxcommandline}", problem.getSandboxCmdLine());
+                    newString = substituteAllStrings(inRun, newString);
+                } else {
+                    // remove {:sandbox} if no sandbox defined in pc2
+                    newString = replaceString(newString, "{:sandboxcommandline}", "");
```
clevengr, line 2544: I don't understand this statement.  The comment above it says it is removing "{:sandbox}", but the statement itself is replacing "{:sandboxcommandline]".  Either the comment is wrong or the statement is wrong (or I REALLY don't understand what is going on).  Either the comment needs to be corrected or the statement needs to be corrected.  (Or I would accept a brief explanation of how the statement actually does do what the comment says it does...)

test/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoaderTest.java
```diff
@@ -3710,5 +3712,24 @@ private IInternalContest fullLoadSampleContest(String sampleName) throws Excepti
...
+        Problem[] problems = inContest.getProblems();
+        for (Problem problem : problems) {
+            if (letter.equalsIgnoreCase(problem.getLetter())) {
+                return problem;
+            }
+            
+        }
+        return null;
+    }
+
```
clevengr, line 3733: The above block of code doesn't look like it has anything to do with i295.  I'm ok with it; it's just sort of "out of scope" (unless I'm missing something....)

clevengr (commented):
There is one comment I've made which I think needs to be addressed; as soon as that's addressed I see no reason not to approve this PR.

src/edu/csus/ecs/pc2/core/execute/Executable.java
lane55: typo in comment.

The condition is: if the variable sandboxcommandline is empty then substitute with ""
if not empty use the value from problem.

clevengr: So, are you intending to correct it?   (As someone famously once said, "If the documentation doesn't match the code, both are wrong"... ;)

test/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoaderTest.java
lane55: This method is directly used (see above) to test testLoadSandboxAndMemoryLimit.
It is used to find a problem in the model by problem letter.   It is not extraneous,
is a helper method for the testLoadSandboxAndMemoryLimit.

src/edu/csus/ecs/pc2/core/execute/Executable.java
lane55: pushing fix in comment.

test/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoaderTest.java
clevengr: Aha; sorry, I missed that above.  Emily Litella...

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: Well, I guess totally eliminating documentation is one way to make sure the documentation isn't inaccurate :(

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2524,10 +2525,35 @@ public String substituteAllStrings(Run inRun, String origString) {
...
+                }
+            }
+
+            if (!StringUtilities.isEmpty(problem.getSandboxProgramName())) {
+                newString = replaceString(newString, "{:sandboxprogramname}", problem.getSandboxProgramName());
+            } else {
+                newString = replaceString(newString, "{:sandboxprogramname}", "");
+            }
+
```
clevengr, line 2555: All this time spent looking at this block of code has made me realize there's a substantial problem (more than just a wayward/inaccurate comment).  This block of code is calling method on "problem", without first checking that problem is not null.  It seems to me that this entire block of code belongs further up, inside the block around line 2500 that performs a not-null check for problem.  (Or else this block should re-perform the null check and provide the corresponding logging; I think it makes more sense to move it to the above place which is already doing "problem-related" processing>

clevengr (commented):
See the new comments in Executable; I think there's a missing null-check (which could be easily solved by moving a block of code upwards).

clevengr (approved):
Approved; I'll merge.

PR #359, opened 2022-03-21
I 332 edit run should not always select yes judgement

### LINKED ISSUE #332: Edit run mistakenly always selects (defaults to) Yes judgement
**Describe the issue**: 

Edit run, the judgement is always Yes.

Judgement should match the Run's judgement 

**To Reproduce**: 

Edit any judged run that is not a Yes.

**Expected behavior**: 

The judgement should match the ("no") judgement shown
in the run table.

**Actual behavior**: 

Always selects the Yes judgement.
As a result edit run assumes the run has changed and will prompt
to save changes.

**Environment**: 

All

**Log Info**:  

**Screenshots**:  

**Additional context**:

lane55:

Fixed bug where edit run always showed judgement as Yes.
Now edit run (if run judged) will show judgement.

Sometimes a validated run judgement, status on the Runs Tab, will
be different than the judgement displayed/selected in edit run.
In this case hover over the Judgement label to reveal the tool tip
which will show the judgement in edit run and the judgement in the Run Tab status column.

### Issue which the PR fixes

#332 

The bug is edit run for every run regardless of judgement selects the judgement Yes.

Add/submit runs to the system both with Yes and No judgements.

Edit the "No" runs and edit run will select a No judgement, before always selected Yes.
  (Note if judgement different than on Run tab, hover over the Judgement combo box tab 
to view both the judgement assigned (usually RTE) and the Validator result string.

For every edit run will log the initial judgement (with validator result string too).

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/EditRunPane.java
+++ b/src/edu/csus/ecs/pc2/ui/EditRunPane.java
@@ -434,10 +434,26 @@ public class EditRunPane extends JPanePlugin {
 
         ElementId judgementId = null;
 
-        if (run.isJudged()) {
-            judgementId = run.getJudgementRecord().getJudgementId();
-        }
+        judgementLabel.setText("Judgement");
 
+        if (run.isJudged()) {
+            JudgementRecord jr = run.getJudgementRecord();
+            judgementId = jr.getJudgementId();
+
+            // Add judgement description to judgement tool tip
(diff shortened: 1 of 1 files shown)


johnbrvc (approved):
Finally!  Seems to work as advertised.  I noticed the CI with the tooltip on the judgement details slipped in.  That's nice too.

clevengr (approved):

PR #360, opened 2022-03-21
I 295 develop work

clevengr:
Adds code to Executable to check for and invoke a PC2 Internal Sandbox.  

### Issue which the PR fixes
Partial for #295 

Win 10.1, Java 1.8_201, Eclipse Version: 2019-12 (4.14.0)
Partial development; no tests available yet.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -40,6 +40,7 @@ import edu.csus.ecs.pc2.core.model.Judgement;
 import edu.csus.ecs.pc2.core.model.JudgementRecord;
 import edu.csus.ecs.pc2.core.model.Language;
 import edu.csus.ecs.pc2.core.model.Problem;
+import edu.csus.ecs.pc2.core.model.Problem.SandboxType;
 import edu.csus.ecs.pc2.core.model.ProblemDataFiles;
 import edu.csus.ecs.pc2.core.model.Run;
 import edu.csus.ecs.pc2.core.model.RunFiles;
@@ -211,6 +212,8 @@ public class Executable extends Plugin implements IExecutable {
     private String packageName = "";
 
     private String packagePath = "";
+    
+    private boolean usingSandbox = false;
 
     public Executable(IInternalContest inContest, IInternalController inController, Run run, RunFiles runFiles) {
         super();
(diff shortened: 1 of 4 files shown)


src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -211,6 +212,8 @@
     private String packageName = "";
 
     private String packagePath = "";
+    
+    private boolean usingSandbox = false;
```
lane55, line 216: remove usingSandbox field.   More below on moving the code that
sets usingSandbox into a new method.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1811,9 +1813,78 @@ protected boolean executeProgram(int dataSetNumber) {
 
             }
 
-            log.log(Log.DEBUG, "before substitution: " + cmdline);
+            //check whether the Problem is configured to use a sandbox
```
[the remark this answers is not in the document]
lane55: Suggested sig is:
private boolean isUsingSandBox(BufferedOutputStream stderrlog, BufferedOutputStream stdoutlog) throws IOException {
change name to be more descriptive.

src/edu/csus/ecs/pc2/core/execute/Executable.java
lane55: rather than using a field, use a local variable in executeProgram for the two places where this field variable is used.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2416,6 +2538,8 @@ public String replaceString(String origString, String beforeString, int afterInt
      *              {:outfile}
      *              {:ansfile}
      *              {:pc2home}
+     *              {:sandbox} -the sandbox program (only under non-Windows OS's)
```
lane55, line 2541: doc, Unclear why there is a windows only variable?
I suggest removing this reference to {:sandbox}

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1994,9 +2075,10 @@ public void run() {
             //get rid of the TLE timer (whether the TLE-kill task has been fired or not)
             log.info("cancelling TLE-Timer (note: this does not stop any already-running TLE-Timer tasks...)");
             timeLimitKillTimer.cancel();
-            
-//            System.out.println("  Process run time was " + getExecutionTimeInMSecs() + "ms");
 
+
+            //////// need code here to deal with sandbox results..
```
lane55, line 2080: "TODO" Comments like this should use the convention of prefixing the comment with TODO.

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: done.

src/edu/csus/ecs/pc2/core/execute/Executable.java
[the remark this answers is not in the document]
clevengr: Done, except that I simplified the signature (there's no need to pass the log streams into the method, as it throws an exception when any condition needing those logs occurs and the logging is handled in the catch.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2119,6 +2201,46 @@ public void run() {
         return proceedToValidation;
     }
 
+    /**
+     * Generates, in the execution directory, a file whose name corresponds to the PC2 Internal Sandbox program name
+     * and whose contents are the lines contained in the PC2 Internal Sandbox code.
+     * 
+     * @return true if creation of the sandbox file in the execution directory was successful; false if not.
+     */
+    private boolean generatePC2Sandbox() {
```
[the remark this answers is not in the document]
clevengr: Done

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: Well, it's not that it's "Windows-only", it's that it's "Not Windows".  Anyway, I wrote this comment to make it clear that the "{:sandbox}" substitution in only used on non-Windows systems.  But I grant it's a bit confusing; I've eliminated the reference.  I also corrected the substitution variable names.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1813,68 +1812,16 @@ protected boolean executeProgram(int dataSetNumber) {
...
-                usingSandbox = true;
+            //determine whether the Problem is configured to use a sandbox 
+            try {
+                usingSandbox = isUsingSandbox();
+            } catch (Exception e) {
+                //an exception means there is something wrong about sandbox usage (e.g. not allowed on this platform)
+                log.severe("Exception during sandbox usage check: " + e.getMessage());
+                stderrlog.close();
+                stdoutlog.close();
+                executionData.setExecuteSucess(false);
```
lane55, line 1823: Please add the following line before the return
executionData.setExecutionException(e);

Until we refeactor executable to throw exceptions, it is helpful to save the exception so we
can access it later  using the executionData.   Adding the above line gives us that access.

lane55 (changes requested):
All else is good, one change add line
executionData.setExecutionException(e);
where specified.

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: Good point; done.

lane55 (approved):
Perfect.

Issue #361, opened 2022-03-26
DefaultScoringAlgorithm causes Null-pointer Exception when formatting team name on older profiles


johnbrvc:
**Describe the issue**:  DefaultScoringAlgorithm gets Null-pointer exception causing a warning pop-up when starting the pc2admin client.

Standings tab is empty.

**To Reproduce**: 

Load an older profile before #254.  Run pc2admin.  

**Expected behavior**: 

No warning should be issued, 
and the standings in Standings tab should be generated properly.  

**Actual behavior**: See "To Reproduce"

A warning will pop-up about not being able to create the standings HTML. 

The Standings tab will have no standings rows.

This is due to a Null-pointer being returned from ContestInformation.getTeamScoreboardDisplayFormat().

**Environment**: PC2 V9.7.0 being loaded into newer version of Pc2

**Log Info**:  
220325 163703.740|WARNING|AWT-EventQueue-0|parseAndDisplay|Trouble creating or parsing SA XML 
|java.lang.NullPointerException: Cannot invoke "String.indexOf(String)" because "origString" is null
|    at edu.csus.ecs.pc2.util.ScoreboardVariableReplacer.substituteDisplayNameVariables (ScoreboardVariableReplacer.java:79)
|    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.createStandingXML (DefaultScoringAlgorithm.java:701)
|    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.getStandings (DefaultScoringAlgorithm.java:462)
|    at edu.csus.ecs.pc2.ui.StandingsPane.parseAndDisplay (StandingsPane.java:151)
|    at edu.csus.ecs.pc2.ui.StandingsPane$1.run (StandingsPane.java:195)
|    at java.awt.event.InvocationEvent.dispatch (InvocationEvent.java:318)

**Additional context**: A member variable, teamScoreboardDisplayFormat, was added by i-254 that causes deserialization of older class objects not to restore that field properly, and it is set to null.

johnbrvc:
A quick and easy fix that should work going forward is to set the teamScoreboardDislayFormat member to ScoreboardVariableReplacer.TEAM_NAME if it is null in the accessor, getTeamScoreboardDisplayFormat().
Changing the serial number of the class could work too, but it would not make it possible to load older profiles or pc2zip data,
and maintaining this capability is desired:

public String getTeamScoreboardDisplayFormat() {
[...]

PR #362, opened 2022-03-28
i361 - Prevent NPE caused by uninitialized name format string

johnbrvc:
When deserializing older objects of type ContestInformation that do not contain the teamScoreboardDisplayFormat member, make sure that member gets set to the default when the accessor is called.  Before the fix, bringing up the pc2admin would cause a warning about not being able to create the Standings HTML.  This was caused by the name format string being null because it was not set during the deserialization of the old ContestInformation object.

This addresses #361 - DefaultScoringAlgorithm causes Null-pointer Exception when formatting team name on older profiles

Bringing up pc2admin no longer causes the warning or the null pointer exception.

Note: while a more "correct" approach to fixing this issue may be to change the serial number of the ContestInformation class, this would make make restoring older contests virtually impossible.  It was therefore decided to just set the missing member variable to the default value, ScoreboardVariableReplacer.TEAM_NAME, which happens to be {:teamname}

### DIFF
--- a/src/edu/csus/ecs/pc2/core/model/ContestInformation.java
+++ b/src/edu/csus/ecs/pc2/core/model/ContestInformation.java
@@ -760,6 +760,11 @@ public class ContestInformation implements Serializable{
      * @return
      */
     public String getTeamScoreboardDisplayFormat() {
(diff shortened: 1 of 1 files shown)


lane55 (approved):
Approved.

clevengr (approved):

PR #363, opened 2022-03-29
i318: merge rather than replace group list on load of teams.tsv

### LINKED ISSUE #318: teams.tsv import unconditionally clears group - it shouldn't.
**Describe the issue**: 

When teams tsv is imported (ICPC - Import CMS tsv files) the group field is set to empty/null.

It would be helpful if any future NPE or other Exceptions show a dialog to the user,
if variable useConfirmGUI is true (meaning the load is done via GUI rather than command line/no GUI)

Start a server with a sample with a teams.tsv, --load samps\contests\valtest
Start Admin
Select Accounts tab, note all teams are assigned groups
ICPC -> Import CMS tsv files
Select \contests\valtest\config\teams.tsv, answer Yes
Select Accounts tab

**Expected behavior**: 
Team account information should not change

**Actual behavior**: 
No teams are assigned a group

NPE logged into log

**Environment**: 
All, tested using
Version 9.8build 20220223 (Wednesday, February 23rd 2022 23:46 UTC) build 6215~develop
Java version 1.8.0_121
OS: Windows 10 10.0 (amd64)

**Log Info**:  

220228 082208.548|INFO|AWT-EventQueue-0|loadTSVFiles|Exception in loadTSVFiles()
java.lang.NullPointerException: null
    at java.io.File.<init> (File.java:277)
    at edu.csus.ecs.pc2.core.imports.LoadICPCTSVData.checkFiles (LoadICPCTSVData.java:253)
    at edu.csus.ecs.pc2.core.imports.LoadICPCTSVData.loadFiles (LoadICPCTSVData.java:85)
    at edu.csus.ecs.pc2.core.imports.LoadICPCTSVData.loadFiles (LoadICPCTSVData.java:72)
    at edu.csus.ecs.pc2.ui.ICPCLoadPane.loadTSVFiles (ICPCLoadPane.java:434)
    at edu.csus.ecs.pc2.ui.ICPCLoadPane$5.actionPerformed (ICPCLoadPane.java:372)

**Screenshots**:  

**Additional context**:

lane55:

Groups were being blanked out on Accounts list on load  of teams.tsv and groups.tsv.
Now merges model's groups and loaded groups from groups.tsv.

### Issue which the PR fixes

#318 

java version "1.8.0_121"

tldr;   Start server with sample contest that has teams.tsv, groups.tsv, 
   load the same teams.tsv using ICPC tab -> Import CMS tsv files
   view Accounts, and confirm that groups are blanked out.

Load sample contest mini
 pc2server --login s --contestpassword contest  --load mini
Start Admin 
note that groups are populated for accounts
ICPC tab -> Import CMS tsv files 
load either groups.tsv or teams.tsv from samps\contests\mini\config
 
Expected

In Accounts tab, groups are present.

Actual:

All team groups are blank.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/imports/LoadICPCTSVData.java
+++ b/src/edu/csus/ecs/pc2/core/imports/LoadICPCTSVData.java
@@ -23,10 +23,8 @@ import edu.csus.ecs.pc2.ui.UIPlugin;
  * Read input .tsv files, validate then if valid load into contest.
  * 
  * @author pc2@ecs.csus.edu
- * @version $Id$
  */
 
-// $HeadURL$
 public class LoadICPCTSVData implements UIPlugin {
 
     public static final String TEAMS2_TSV = "teams2.tsv";
@@ -87,13 +85,13 @@ public class LoadICPCTSVData implements UIPlugin {
             String instFilename = groupsFilename.replaceFirst(GROUPS_FILENAME,"institutions.tsv");
             // this must be loaded before accounts, and no harm before groups
             ICPCTSVLoader.loadInstitutions(instFilename);
-            Group[] groups = ICPCTSVLoader.loadGroups(groupsFilename);
+            Group[] groups = ICPCTSVLoader.loadGroups(groupsFilename, contest.getGroups());
             
-//            for (Group group : groups) {
-//                group.setSite(getContest().getSites()[0].getElementId());
-//            }
-            
+//          for (Group group : groups) {
+//              group.setSite(getContest().getSites()[0].getElementId());
+//          }
             Account[] accounts = ICPCTSVLoader.loadAccounts(teamsFilename);
+            
             if (!accountsFilename.equals("")) {
                 HashMap<Integer, String> passwordMap = ICPCTSVLoader.loadPasswordsFromAccountsTSV(accountsFilename);
                 if (!passwordMap.isEmpty()) {
@@ -144,7 +142,6 @@ public class LoadICPCTSVData implements UIPlugin {
                      * UpdateAccounts
                      */
                     Account[] updatedAccounts = (Account[]) accountList.toArray(new Account[accountList.size()]);
-
                     getController().updateAccounts(updatedAccounts);
                     info("Sent accounts from " + teamsFilename + " to server");
                 } else {
@@ -152,6 +149,7 @@ public class LoadICPCTSVData implements UIPlugin {
                     // since not sending to server, add to existing contest.
                     
                     Group[] updatedGroups = (Group[]) groupList.toArray(new Group[groupList.size()]);
+                    
                     for (Group group : updatedGroups) {
                         contest.updateGroup(group);
                     }
@@ -227,6 +225,8 @@ public class LoadICPCTSVData implements UIPlugin {
             
             accountList.set(i, existingAccount);
             i ++;
+            
+            
         }
         
     }
--- a/src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
+++ b/src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
@@ -259,7 +259,7 @@ public final class ICPCTSVLoader {
     }
     
     /**
-     * Load groups into this class.
+     * Merge/Add group data from filename and authoritativeGroups.
      * 
      * @param filename name of groups.tsv file.
      * @return list of groups 
@@ -270,10 +270,11 @@ public final class ICPCTSVLoader {
     }
     
     /**
-     * Merge/Load groups into this class
+     * Merge/Add group data from filename and authoritativeGroups.
+     * 
      * @param filename name of groups.tsv file.
      * @param authoritativeGroups an array of groups (from contest/model)
-     * @return list of groups
+     * @return unique groups from authoritativeGroups and filename's group list.
      * @throws Exception
      */
     public static Group[] loadGroups(String filename, Group[] authoritativeGroups) throws Exception {
@@ -295,11 +296,6 @@ public final class ICPCTSVLoader {
 
         // 1 Label groups fixed string (always same value)
         // 2 Version number 1 integer
-
-        // TODO CCS do check for 'groups' when tsv file contains that field.
-//        if (!fields[0].trim().equals("groups")) {
-//            throw new InvalidValueException("Expecting 'groups' got '" + fields[0] + "' in " + filename);
-//        }
 
         i++;
 
--- a/src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
+++ b/src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
@@ -2,7 +2,9 @@
 package edu.csus.ecs.pc2.imports.ccs;
 
 import java.io.FileNotFoundException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 
 import edu.csus.ecs.pc2.core.PermissionGroup;
 import edu.csus.ecs.pc2.core.Utilities;
@@ -232,7 +234,7 @@ public final class ICPCTSVLoader {
     }
 
     private static ElementId getGroupForNumber(int groupId) {
-
+        
         for (Group group : groups) {
             if (group.getGroupId() == groupId) {
                 return group.getElementId();
@@ -255,15 +257,27 @@ public final class ICPCTSVLoader {
     public static void setGroups(Group[] groups) {
         ICPCTSVLoader.groups = groups;
     }
-
+    
     /**
      * Load groups into this class.
      * 
-     * @param filename
-     * @return list of groups or new Group[0]
+     * @param filename name of groups.tsv file.
+     * @return list of groups 
      * @throws Exception
      */
     public static Group[] loadGroups(String filename) throws Exception {
+        return loadGroups(filename, new Group[0]);
+    }
+    
+    /**
+     * Merge/Load groups into this class
+     * @param filename name of groups.tsv file.
+     * @param authoritativeGroups an array of groups (from contest/model)
+     * @return list of groups
+     * @throws Exception
+     */
+    public static Group[] loadGroups(String filename, Group[] authoritativeGroups) throws Exception {
+        List<Group> groupList = new ArrayList<Group>();
         
         String[] lines = CCSListUtilities.filterOutCommentLines(Utilities.loadFile(filename));
         
@@ -288,10 +302,6 @@ public final class ICPCTSVLoader {
 //        }
 
         i++;
-
-        groups = new Group[lines.length - 1];
-
-        int groupAccount = 0;
 
         for (; i < lines.length; i++) {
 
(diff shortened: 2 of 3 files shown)


src/edu/csus/ecs/pc2/core/imports/LoadICPCTSVData.java
```diff
@@ -227,6 +225,8 @@ protected void updateGroupsAndAccounts(IInternalContest inContest, List<Group> g
             
             accountList.set(i, existingAccount);
             i ++;
+            
+
```
clevengr, line 229: I can understand the addition of a _single_ blank line for improved readability, but personally I think adding TWO blank lines just makes it _harder_ to read the code.   [This is not a request for a change; just an observation...]

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
```diff
@@ -319,17 +329,57 @@ public static Group[] loadGroups(String filename) throws Exception {
...
-            groups[groupAccount] = group;
-            groupAccount++;
+            
+            groupList.add(group);
         }
-
+        
+        /**
+         * Add groups from authoritativeGroups if missing from groupList.
+         */
```
clevengr, line 346: I do not understand the purpose of, or the need for, the following for-loop.  The comment says that it's to add to `groupList` any groups that were in `authoritativeGroups` but are missing from `groupList`.  However, the new code added to the previous for-loop (just above here) obtains a group from the input file, then looks to see if that group is in `authoritativeGroups` (using `Group group = findGroupById(authoritativeGroups, number);`).  It then adds the group (either the one found in `authoritativeGroups` or else a newly-created group) to `groupList`.   Therefore, at THIS point in the code _every_ group found in `authoritativeGroups` MUST also now be in `groupList` -- so what's the point in the entire for-loop following this comment?
Or am I missing something?

clevengr: BTW, I would not block approval for this -- but I don't like having chunks of code that are useless.   (Unless I'm wrong about what's going on, in which case please explain how I'm wrong...)

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
```diff
@@ -319,17 +329,57 @@ public static Group[] loadGroups(String filename) throws Exception {
...
         return groups;
     }
-    
+
+    /**
+     * Search and return group if found.
+     * 
+     * @param authoritativeGroups list of groups
+     * @param groupCMSId group (CMS
+     * @return null if not found, else the group
```
clevengr, line 372: What happened to the rest of the comment here?

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
```diff
@@ -376,4 +427,6 @@ public static HashMap<Integer, String> loadPasswordsFromAccountsTSV(String filen
         }
         return passwordMap;
     }
+
+
```
clevengr, line 431: Again... why two blank lines?

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
```diff
@@ -111,5 +115,88 @@ private void checkIfGroupsAssigned(List<Account> accountList) {
...
+    private String getTestSampleContestConfigFile(String contestName, String filename) {
+        return getTestSampleContestDirectory(contestName) + File.separator + IContestLoader.CONFIG_DIRNAME + File.separator + filename;
+    }
+
+    /**
+     * Test loading a teams.tsv twice.
+     * 
+     * The 2nd load of teams and groups, in particular groups, caused all accounts' groups to be "empty"
+     * on the Accounts tab.  See 
+     *
```
clevengr, line 144: See.... what?

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
```diff
@@ -111,5 +115,88 @@ private void checkIfGroupsAssigned(List<Account> accountList) {
...
+
+        IInternalContest contest = loadSampleContest(null, "mini");
+        assertNotNull(contest);
+        IInternalController controller = new SampleContest().createController(contest, true, false);
+        
+        LoadICPCTSVData loader = new LoadICPCTSVData();
+        loader.setContestAndController(contest, controller);
+        
+        boolean loaded = loader.loadFiles(groupsFilename, false, false);
+        assertTrue("Expecting min contest loaded", loaded);
```
clevengr, line 160: I think maybe the word in the message was supposed to be "mini", not "min"?

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
```diff
@@ -111,5 +115,88 @@ private void checkIfGroupsAssigned(List<Account> accountList) {
...
+         * Groups from contest/model, the authoritative groups from model.
+         */
+        Group[] modelGroups = contest.getGroups();
+
+        assertEquals("Number of groups", 12, modelGroups.length);
+        assertEquals("Number of accounts", 151, accounts.length);
+
+        /**
+         * Second load of groups from groups.tsv
+         */
```
clevengr, line 183: This is a repeat of a comment above; that doesn't seem accurate.

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
```diff
@@ -111,5 +115,88 @@ private void checkIfGroupsAssigned(List<Account> accountList) {
...
+        // Load groups from file, merge with modelGroups list.
+        Group[] mergedGroups = ICPCTSVLoader.loadGroups(loader.getGroupsFilename(), modelGroups);
+
+        Account[] accounts2 = ICPCTSVLoader.loadAccounts(loader.getTeamsFilename());
+
+        assertEquals("Number of groups", 12, mergedGroups.length);
```
clevengr, line 189: This doesn't seem like a valid (or at least, not a very good) test of merging -- the group sizes (12) are the same before and after the merge. Unless I've missed something which makes them not the same?  If so, please illuminate me...

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
```diff
@@ -111,5 +115,88 @@ private void checkIfGroupsAssigned(List<Account> accountList) {
...
+        assertEquals("Number of accounts", 151, accounts2.length);
+
+        for (int i = 0; i < modelGroups.length; i++) {
+            assertEquals("(" + i + ") Expecting same group element id for group " + modelGroups[i], modelGroups[i].getElementId(), mergedGroups[i].getElementId());
+        }
+
+        // check that all accounts assigned groups in the model/contest
+        for (Account account : accounts2) {
+            Group group = contest.getGroup(account.getGroupId());
```
clevengr, line 198: `account.getGroupId()` can return null (if no Group is assigned).  I don't know if that's a problem here or not...

clevengr (commented):
I've made a number of comments.  None are overwhelming, but I think it's worth at least having you review them, and "Resolve" them if you think appropriate.

src/edu/csus/ecs/pc2/core/imports/LoadICPCTSVData.java
lane55: noted.

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: This test ensures that after the merge there are the proper, 12, number of groups.
If there are other than 12 then the merge has failed.

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: It can, and if it does then this unit test will fail... and the cause can be fixed/researched.

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
lane55: java doc updated.

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
lane55: The comment is correct.     Merge may imply that the number of groups in the input file are identical to
the number/data for the existing authoritativeList.   That would be a bug.   There may be more or different
groups in the authoritative list than there are in the input file.    The merge is a union of groups in
the input file and the authoritativeList (model).   For example there could be 12 authoritativeList
groups and only 4 groups in the input file.  There should be a minimum of 12 groups output.  Without
the second loop there would be 4 groups output.

Logic is:
The top loop merges the authoritative/existing group data and the groups from the input file.  The comment is:
   "Add groups from authoritativeGroups if missing from groupList."
The bottom loop adds any missing groups from the authoritative list into the method's output groupList

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
lane55: Stuff happens.   I would like to focus on non-whitespace and non-format related actual code work/review.

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: fixed comment, added issue url

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: updated to use variable for contestName, pushed.

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
clevengr: Ok, it took me a while but now I get why the second loop is needed.  However, it took me a LONG time to understand what the second loop is actually doing, due to very bad naming (see below).  I believe the code works -- but it's really confusing to the reader (again, see below).

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
```diff
@@ -319,17 +329,57 @@ public static Group[] loadGroups(String filename) throws Exception {
...
+        /**
+         * Add groups from authoritativeGroups if missing from groupList.
+         */
+        for (Group aGroup : authoritativeGroups) {
+            
+            Group missingGroup = null; 
+            
+            for (Group glGroup : groupList) {
+                if (aGroup.getGroupId() == glGroup.getGroupId()) {
+                    missingGroup = glGroup;
```
clevengr, line 349: `missingGroup` is a VERY bad/misleading name for this variable.  In fact, when a value is assigned to `missingGroup`, that  value is a Group which is NOT missing (from `groupList`).  The reversed logic of this is quite confusing (although I'll concede that the binary works...)

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
clevengr: `I would like to focus on non-whitespace and non-format related actual code work/review.`  Sure, I get that.  But if I'm reviewing code I'm going to comment where I think there's an issue even if it's trivial; failure to do so means things creep into code that make it harder in the future (in some, very weak in this case, sense).  I did note that I certainly wouldn't refuse to approve such code...

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
clevengr: Right, but my point was:  what about the case where 12 is NOT the proper number?  E.G. when a new Group is being added?  It seems like THAT case is a much more substantial test of whether a "merge" worked correctly.

clevengr (approved):

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: removed.

test/edu/csus/ecs/pc2/core/imports/LoadICPCTSVDataTest.java
lane55: You asked about why it was valid, not what a better test would be.   Sure there are bunches of tests
that could be added, I added that test for the reasons above.

src/edu/csus/ecs/pc2/imports/ccs/ICPCTSVLoader.java
lane55: Suggest a less misleading name, rather than just comment.   I used that name thinking at the time that
it made sense (or more likely the code section was rewritten keeping that name).

clevengr: Well, since what this loop is doing is checking, for each Group in `authoritativeGroups`, whether that group is found in `groupList` (and if NOT, it is adding the Group from `authoritativeGroups` into `groupList`, a "better" name might be "found".

However, the real problem is the confusion in what this loop is trying to accomplish vs. the code it uses to accomplish that.  What it is trying to accomplish (according to the comment at the top, and, unless I'm mistaken, in the code itself) is to add to `groupList` any Group which is currently in `authoritativeGroups` but is NOT already in `groupList`.  I think a better solution would be to replace the entire loop with something like:
```
       for (Group aGroup : authoritativeGroups) {
           if (! groupList.contains(aGroup) {
              groupList.add(aGroup);
              break;
          }
       }
```
This is much simpler and clearer, and I believe it accomplishes what is intended (although I would certainly expect you to review that assertion :)  )

lane55: So no change in the name of the variable, ok.

lane55: pushed fix.

PR #371, opened 2022-04-07
I 295 develop work

clevengr:
  Updates a number of classes in support of partial work on #295.  Updated classes include:  Problem, ProblemsPane, EditProblemFrame, EditProblemPane, and EditProblemSandboxPane.   The result is (what I believe to be) a working GUI for supporting Memory Limits.  (Note however that this doesn't mean Memory Limits now work; that requires additional updates for invoking a sandbox shell.  This PR is strictly about (hopefully) completing the GUI support for specifying Memory Limits.)

### Issue which the PR fixes
None; it is a *partial* fix for Issue 295.

Windows 10.1, Eclipse Version: 2019-12 (4.14.0), Java 1.8_201.

- Start a server with a contest with at least one problem.
- Start an Admin.
- Select the "Problems" tab.
- Select a problem and click the "Edit" button.
- Select the "Sandbox" tab.
- Test the following:
  - ability to select either "No Sandbox" or "PC2 Sandbox"
  - *inability* to select "Custom Sandbox"
  - ability to enter an arbitrary memory limit value when the PC2 Sandbox button is selected
  - *inability* to enter a non-numeric memory limit value
  - entering a non-negative memory limit value and then pressing "Update" saves the memory limit value in the problem
    - the new memory limit value displays on the "Problems Pane" in the "Memory Limit" column
    - re-editing the same problem causes the new memory limit value to appear in the "Memory Limit" textbox on the edit screen
  - entering a negative memory limit value causes the memory limit to be set to zero
  - Pressing each of the "Information Question Mark" icons causes an appropriate and helpful explanatory message to be displayed.

### Additional Information
  The code in EditProblemSandboxPane includes declarations for two "debug variables":  *debugShowPanelBorders*, which is useful when debugging GUI layout issues, and *debugAllowSandboxSelectionOnWindows*.  This latter variable allows testing the GUI selection capabilities on a Windows system even though eventually we are going to disallow such selection on Windows (until we get a working Windows sandbox).    Currently, *debugShowPanelBorders* is set false while *debugAllowSandboxSelectionOnWindows* is set true (to allow reviewers to test the GUI).  The *debugAllowSandboxSelectionOnWindows* variable needs to be set false in a production distribution...

### DIFF
--- a/src/edu/csus/ecs/pc2/core/model/Problem.java
+++ b/src/edu/csus/ecs/pc2/core/model/Problem.java
@@ -1195,7 +1195,22 @@ public class Problem implements IElementObject {
                 return false;
             }
             
+            //check for equivalence in Sandbox configuration
+            if (this.getSandboxType() != otherProblem.getSandboxType()) {
+                return false ;
+            } else if (this.getSandboxType() == SandboxType.PC2_INTERNAL_SANDBOX) {
+                if (this.hasMemoryLimit() != otherProblem.hasMemoryLimit() ) {
+                    return false;
+                } else {
+                    if (this.getMemoryLimitMB() != otherProblem.getMemoryLimitMB()) {
+                        return false;
+                    }
+                }
+            }
+            
+            //all comparisons pass; problems are equivalent
             return true;
+            
         } catch (Exception e) {
             StaticLog.getLog().log(Log.WARNING, "Exception comparing Problem "+e.getMessage(), e);
             e.printStackTrace(System.err);
--- a/src/edu/csus/ecs/pc2/core/model/Problem.java
+++ b/src/edu/csus/ecs/pc2/core/model/Problem.java
@@ -14,6 +14,7 @@ import edu.csus.ecs.pc2.core.log.Log;
 import edu.csus.ecs.pc2.core.log.StaticLog;
 import edu.csus.ecs.pc2.core.model.inputValidation.InputValidationResult;
 import edu.csus.ecs.pc2.core.model.inputValidation.VivaInputValidatorSettings;
+import edu.csus.ecs.pc2.ui.EditProblemSandboxPane;
 import edu.csus.ecs.pc2.validator.clicsValidator.ClicsValidatorSettings;
 import edu.csus.ecs.pc2.validator.customValidator.CustomValidatorSettings;
 import edu.csus.ecs.pc2.validator.pc2Validator.PC2ValidatorSettings;
@@ -36,7 +37,9 @@ public class Problem implements IElementObject {
 
     public static final int DEFAULT_TIMEOUT_SECONDS = 10;
     
-    public static final int DEFAULT_MEMORY_LIMIT_MB = 2048 ;
+    public static final int DEFAULT_MEMORY_LIMIT_MB = 0 ;   //zero memory limit = "none", i.e. the problem can use all available memory
+    
+    public static final SandboxType DEFAULT_SANDBOX_TYPE = SandboxType.NONE ;
     
     /**
      * Problem title.
@@ -211,8 +214,6 @@ public class Problem implements IElementObject {
     private VivaInputValidatorSettings vivaSettings = null;
 
 
-    
-
     /**
      * Use international judgement method.
      */
@@ -332,9 +333,8 @@ public class Problem implements IElementObject {
         EXTERNAL_SANDBOX 
     }
     
-    private boolean hasMemoryLimit = false;
     private int memoryLimitMB = DEFAULT_MEMORY_LIMIT_MB;
-    private SandboxType sandboxType = SandboxType.NONE;
+    private SandboxType sandboxType = DEFAULT_SANDBOX_TYPE;
     private String sandboxCmdLine = Constants.PC2_INTERNAL_SANDBOX_COMMAND_LINE;
     private String sandboxProgramName = Constants.PC2_INTERNAL_SANDBOX_PROGRAM_NAME;
     
@@ -354,8 +354,6 @@ public class Problem implements IElementObject {
         this.customInputValidationStatus = InputValidationStatus.UNKNOWN;
         this.customInputValidationResults = new Vector<InputValidationResult>();
         this.vivaSettings = new VivaInputValidatorSettings();
-        this.hasMemoryLimit = false;
-        this.sandboxType = SandboxType.NONE;
     }
 
     public Problem copy(String newDisplayName) {
@@ -455,6 +453,9 @@ public class Problem implements IElementObject {
             clone.addGroup(group);
         }
         
+        clone.setSandboxType(this.getSandboxType());
+        clone.setMemoryLimit(this.getMemoryLimitMB());
+        
         return clone;
     }
 
@@ -543,6 +544,9 @@ public class Problem implements IElementObject {
         retStr += "; usingExternalDataFiles=" + usingExternalDataFiles;
         retStr += "; externalDataFileLocation=" + externalDataFileLocation;
         retStr += "; state=" + state;
+        
+        retStr += "; sandboxType=" + this.getSandboxType();
+        retStr += "; memoryLimit=" + this.getMemoryLimitMB();
       
         retStr += "]";
         return retStr;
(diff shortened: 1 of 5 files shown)


src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -1992,31 +2008,11 @@ public void setVivaInputValidatorHasBeenRun(boolean vivaInputValidatorHasBeenRun
         vivaSettings.setVivaInputValidatorHasBeenRun(vivaInputValidatorHasBeenRun);
     }
 
-    /**
-     * Returns the flag indicating whether this Problem has a runtime memory limit associated with it.
-     * 
-     * @return the flag indicating whether or not this Problem has been configured with a memory limit.
-     */
-    public boolean hasMemoryLimit() {
-        return hasMemoryLimit;
```
lane55, line 2001: Can we remove the field hasMemoryLimit and change this method just...
return memoryLimitMB  > 0;
?

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -211,8 +214,6 @@
     private VivaInputValidatorSettings vivaSettings = null;
 
 
-
```
lane55, line 214: Unable to leave a comment near:   public SandboxType getSandboxType() {
For backwards compatibility and to avoid NPEs should change to:
    public SandboxType getSandboxType() {
        if (sandboxType == null) {
            sandboxType  = SandboxType.NONE;
        }
        return sandboxType;
    }

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -690,232 +700,258 @@ public void enableUpdateButton() {
...
+                    } else {
+                        updateToolTip += ", Output Validator";
+                    }
+                }
 
-                    // see if the choice of which output validator (if any) to use has changed;
-                    // enable the Update button and update the tooltip if so
-                    if (problem.getValidatorType() != changedProblem.getValidatorType()) {
+                // see if the PC2 Validator options have changed; enable the Update button and update the tooltip if so
+                if (problem.getPC2ValidatorSettings() != null && changedProblem.getPC2ValidatorSettings() != null) {
```
lane55, line 721: Is this related to a bug fix?   Validator field are not in the scope of sandbox changes.

src/edu/csus/ecs/pc2/ui/EditProblemFrame.java
```diff
@@ -30,7 +30,7 @@
 
     private IInternalController controller;
 
-    private EditProblemPane problemPane = null;
```
lane55, line 33: Confirm that the only change in this file was a rename of problemPane, thanks.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -690,232 +700,258 @@ public void enableUpdateButton() {
...
                         }
+                    }
 
-                        // check for changes in the specified interface being used
-                        if (problemSettings.isUseClicsValidatorInterface() != changedProblemSettings.isUseClicsValidatorInterface()
-                                || problemSettings.isUsePC2ValidatorInterface() != changedProblemSettings.isUsePC2ValidatorInterface()) {
+                    // check for changes in the actual validator program file
+                    String changedProblemValidatorFileName = changedProblemSettings.getCustomValidatorProgramName();
+                    if (changedProblemValidatorFileName != null && changedProblemValidatorFileName.length() > 0) {
+                        // TODO: there could be a problem here; pdf.getValidatorFile() seems to sometimes return null even though the problem has a validator...
```
lane55, line 783: for this TODO, is the code updated to avoid this potential bug ?

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -5492,8 +5615,20 @@ private void logWarning(String message, Exception e) {
...
-        frame.getContentPane().add(new EditProblemPane());
+
+        //build an EditProblemFrame containing an EditProblemPane containing an EditProblemSandboxPane
+        EditProblemFrame frame = new EditProblemFrame();
+        
+        //put a contest and controller in the frame (which also puts it in the panes)
+        IInternalContest contest;
+        IInternalController controller;
+
+        SampleContest sampleContest = new SampleContest();
```
lane55, line 5626: BUILD FAILURE.   This cause a build failure, SampleContest is not part of the src/ tree it is part of the test/ tree.
Error is: symbol:   class SampleContest
[javac]   location: class EditProblemPane

lane55 (changes requested):
Need to fix the build failure using SampleContst.
Discuss other smaller changes.

src/edu/csus/ecs/pc2/core/model/Problem.java
clevengr: Sure, we COULD do that.  I tend to want to isolate functions with accessors.  What if we later decide we have a different definition of what it means to "have a memory limit"?   If we keep it surrounded by an accessor -- and all other code always uses the accessor (which I think I tried to ensure), we only have to change the accessor to make such a change.  I think it's worth the one-line accessor to provide this improved code structure.

src/edu/csus/ecs/pc2/core/model/Problem.java
clevengr: Agreed; fixed.

src/edu/csus/ecs/pc2/ui/EditProblemFrame.java
clevengr: Confirmed.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: I don't know.  This TODO was already in the code from long ago; I didn't add it and it has nothing to do with Memory Limits.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: Hmmm.... I'm not clear what the correct fix for this is.   Can you suggest a fix?   Or shall I just remove the test function (main)?

src/edu/csus/ecs/pc2/core/model/Problem.java
lane55: Since I did not suggest removing or changing the sig,   My suggestion was to change the code inside
the accessor which is supported by your comment.   Per the comment for Problem:40
 public static final int DEFAULT_MEMORY_LIMIT_MB = 0 ;   //zero memory limit = "none", i.e. the problem can use all available memory
that indicates to me that memoryLimitMB > 0; is the condition for hasMemoryLimit.

lane55:
Suggested Fix for the SampleContest unresolved external/class in EditProblemPane.
Move the main method into the unit test class EditProblemPaneTest, main will be ignored
and not cause a build failure.

src/edu/csus/ecs/pc2/core/model/Problem.java
clevengr: Ok, I misinterpreted your comment.  But:  there is something wrong.  The entire method `hasMemoryLimit()` was removed entirely.  You are commenting about code that was removed.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: No, there is no change (or intent to change) something related to validators.  The "code change" you are seeing is a result of the "diff" routine being unable to deal with a different change:  altering the scope of this test.  That change was necessary in order to place the code which tests for changes in the sandbox in the correct level of scope.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: I've commented out the offending code and will push an update.

lane55 (approved):
Approved.

PR #372, opened 2022-04-15
I 295 develop work invoke sandbox

clevengr:
  Adds support for sandbox invocation (on Linux) from the Executable class.  Includes updates to several classes to provide this support; also includes a few instances of "Continuous Improvement (CI)" fixes to things along the way.

### Issue which the PR fixes
None; it is a partial step toward #295.

Windows 10.1, Eclipse Version: 2019-12 (4.14.0), Java 1.8.0_201.

- Start a server loaded with CDP "sumitHello"
- Start an Admin, use it to:
  - Edit C/C++ languages so "executable file" produces "./{basename}.exe"  (instead of ".\\{basename}.exe" -- different slashes (the default language configurations are Windows-based).
  - Edit Problem "Hello":  set time limit to 10 seconds; select "PC2 Sandbox" with MemLimit 1024.
- On my machine I had to change the file format of pc2judge and pc2env from DOS to Unix.  This may just be a local configuration issue...
- Start a judge, login as J3 (since Judge3 is configured to AutoJudge the sumitHello contest problems).
- Submit the following samples for Problem Hello, Language Java:
  - samps/hello.java (gets back "Yes").
  - samps/NoOutput.java (gets back "Incomplete Output").
  - samps/RuntimeError.java (gets back "No - Run-time Error").

- Examine the JUDGE3.Site1-0.log file in the executeSite1Judge3 directory; observe that each of the three submitted runs has "time" values in the log, generated by the pc2sandbox.sh script.

This verifies that invocation of the sandbox works.

### Additional Information
There are (new) samples for causing TLE and MLE errors.  However, they currently do not work (that is, the existing sandbox is not correctly enforcing TLE and MLE violations).  (Actually, it does enforce the TLE, but it returns the wrong error message).

These are different problems from just verifying that invocation of the sandbox works correctly (which is what this branch/PR is about); those issues will be addressed on a subsequent branch/PR.


samps/src/MLE.c
```diff
@@ -0,0 +1,23 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define PAGE_SZ (1<<12)
+
```
lane55, line 6: Suggest at least a one line comment about what this program does/performs.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2230,18 +2251,14 @@ private boolean copyPC2Sandbox() {
...
         
-        boolean success;
         try {
             //copy the PC2 internal sandbox program into the execute directory
-            success = ExecuteUtilities.copyFile(srcFileName, targetFileName, getLog());
+            ExecuteUtilities.copyFile(srcFileName, targetFileName, getLog());
             
         } catch (Exception e){
             log.severe("Exception copying PC2 Internal Sandbox to execute directory: " + e.getMessage());
-            success = false ;
```
lane55, line 2240: Fix this.  This exception should be added to executionData,  as others are, by adding this line:
executionData.setExecutionException(e);

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
```diff
@@ -877,10 +877,14 @@ protected void executeRun() {
...
             Language lang = getContest().getLanguage(run.getLanguageId());
 
+            //TODO: the following assignment and 'if' don't make sense (the assignment inside the 'if' is 
+            //     the same as the one preceding the 'if').   jlc
             String command = lang.getJudgeProgramExecuteCommandLine();
             if (command == null) {
                 command = lang.getJudgeProgramExecuteCommandLine();
             }
+            
+            //TODO:  the command line needs to be adjusted to reflect the (possible) presence of a sandbox.  jlc
```
lane55, line 887: Fix this. This substitution should not be done here because the auto judge will never execute this code.
All substitutions should be done in Executable for that reason.

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -2105,6 +2105,9 @@ public boolean isUsingSandbox() {
      * @return an element of {@link Problem.SandboxType}.
      */
     public SandboxType getSandboxType() {
+        if (sandboxType == null) {
```
lane55, line 2108: Fix this.  For good design, there should be only one getSandboxType (place where the sandbox type if fetched/got) 
that code/method should include the Problem sandbox type field.
Code was added into Problem so getSandboxType would never be null, we should use/take advantage of that.

src/edu/csus/ecs/pc2/core/execute/JudgementUtilites.java
```diff
@@ -196,7 +196,9 @@ public static void dumpJudgementResultsToLog(Log log, ClientId judgeId, Run run,
 
             int numberOfTestCases = problem.getNumberTestCases();
             log.info(prefixString + " Dumping for " + run + " judge id=" + judgeId);
-            log.info(prefixString + " There are " + numberOfTestCases + " for problem " + problem.getDisplayName() + " '" + problem.getShortName() + "'");
+            String verb = numberOfTestCases > 1 ? " are " : " is " ;
+            String pluralform = numberOfTestCases > 1 ? " testcases " : " testcase ";
```
lane55, line 200: For future pluralizing should use the Pluralize.pluralize method.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1827,33 +1832,37 @@ protected boolean executeProgram(int dataSetNumber) {
             
             if (usingSandbox) {
```
lane55, line 1833: For robustness, useSandbox should be replacd with an accessor method like isUsingSandbox rather than a field variable.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -212,6 +212,11 @@ public class Executable extends Plugin implements IExecutable {
     private String packageName = "";
 
     private String packagePath = "";
+    
+    //setting this to True will override the prohibition on invoking a Sandbox when running on Windows.
```
lane55, line 216: For items like this consider adding a TODO: to easily identify what should be removed.

lane55 (changes requested):
Three fixes, other suggestions.  All fix comments have phrase "fix this"
- add executionData.setExecutionException(e);
- move substitution into Executable so AJ will also use substituttioin
- modify getSandboxType so that value is only set/read in one place, possible remove field sandboxType

samps/src/MLE.c
clevengr: Done, and pushed.

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: Agreed; fixed and pushed.

src/edu/csus/ecs/pc2/core/execute/JudgementUtilites.java
clevengr: Huh.  Didn't know (or at least, didn't remember) that class existed.  Nice.

src/edu/csus/ecs/pc2/core/model/Problem.java
clevengr: Hmmm...  I'm sure what you're asking for. You said "Fix this", and then "For good design there should be only one getSandboxType()..."   But there IS only one such method -- this one.  What exactly are you suggesting needs to be fixed -- that is, what fix are you expecting to be done?

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
clevengr: This is not something I'm prepared to fix.  I'm not saying I disagree about where substitutions should be done.  I'm saying I didn't add the call to "substituteAllStrings()" in this code (which I believe is what you are objecting to).   The ONLY thing I added was the comment (TODO) pointing out that currently, SelectJudgementPaneNew is not coded to properly handle the presence of a sandbox.  I agree that it needs to be fixed.  I'm not at all sure what that fix should be; it's a design problem that I simply discovered.  Fixing it should be handled by creating a new Issue.

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: There is nothing here which "should be removed".   This is a debugging variable intended to remain in the code permanently.   However, I have added a TODO indicating that the variable needs to be changed to False prior to generating a final production/distribution version.

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
lane55: I agree.   A new issue will be created to ensure that both AJ and UI use the same code/data to
execute then run.   I am pretty sure there is another part of the code with the same problem,
the AJ uses different code than the UI to read the validator output and create the judgement.
Issue #373 has been opened.

src/edu/csus/ecs/pc2/core/model/Problem.java
lane55: As stated the method should use the Problem sandbox type field via Problem.getSandboxType() not a local class field.
The sandbox type in the Problem should be authoritative.

Lots of moving parts, I assumed that getSandboxType() was already changed, but I remembered it wrong.
In #361 the method that was changed was ContestInformation.getTeamScoreboardDisplayFormat().
The same bug condition (NPE) will happen with the new sandbox Problem fields if the get accessors
are not updated to handle if the field is null (due to Serialization of values).    The following Problem
methods should be updated to handle if the field == null:  getSandboxProgramName, getSandboxCmdLine,
and getSandboxType so no NPEs can occur.
This can be in scope for this issues as a robustness fix, or a new issue can be created.

clevengr: Doug agreed this comment is moot (he was looking at the wrong file).

lane55 (approved):
All fixes done.

PR #382, opened 2022-04-22
I 380 fix wti scoreboard use scoring properties

### LINKED ISSUE #377: JSONToolTest fails sometimes, because ended date is wrong in JUnit

testtoStateJSONEnded test fails.   The JUnit has a bug, the code works properly.   The JUnit
actual time is the current time, whereas the ended time for the JSON is the contest end date/time,
that is a difference of 5 hours (contest length).

Merge/Push code and cause a build on gitLab, when current time is less than 5 hours from midnight.   

test should pass.

See more below, or gitLab build console output.
    [junit] junit.framework.ComparisonFailure: Expected ended value expected:<2022-04-1[7]> but was:<2022-04-1[6]>

All.

**Log Info**:  PC^2 generates logs (in the **logs** folder for client modules; under the **profiles** folder for the PC2 Server).  Please review the relevant logs and provide copies of them as attachments -- particularly if the logs show Exceptions being thrown.  

From gitLab console output:
    [junit] Testcase: testtoStateJSONEnded(edu.csus.ecs.pc2.services.core.JSONToolTest):	FAILED
    [junit] Expected ended value expected:<2022-04-1[7]> but was:<2022-04-1[6]>
    [junit] junit.framework.ComparisonFailure: Expected ended value expected:<2022-04-1[7]> but was:<2022-04-1[6]>
    [junit] 	at edu.csus.ecs.pc2.services.core.JSONToolTest.testtoStateJSONEnded(JSONToolTest.java:1090)
    [junit]

lane55:

The WTI scoreboard is incorrect if the Penalty For Nos is changed from 20 (default) to -5 (or any other non-20 value)

### Issue which the PR fixes

#377

java version "1.8.0_121"

Unpack and start contest, use tenprobs sample contest

Create contest with the following runs and judgements

  Team 1: submitted a NO at time 2
  Team 1: submitted a YES at time 36.
  Team 3: submitted a YES at time 44 

On Admin: change Scoring Properties Point for No to -5

Unpack and Start WTI

Add scoreboard2 account

Login into WTI at: http://localhost:8080/  as team1 / team1 
Click Scoreboard

**Expected behavior**: 

Scoreboard shows

1 team1 1 31
2 team3 1 44
3 team2 0 0

**Actual behavior**: 

Scoreboard shows

1	team3	1	44
2	team1	1	56
3	team2	0	0


clevengr (commented):
It's not clear to me why it was necessary to make a copy of ScoreboardUtilities in the WTI-API project.  That class is already available from the PC2v9 project, via the pc2.jar.   Now, we have two copies of the same file which have to be maintained/kept in symc.  Is there an important reason for the copy?

clevengr:
Also, it's minor but: the description in the PR includes "Expected" and "Actual", and they are different.  I would think that a PR should be producing expected results that are the SAME as actual results...

troy2914:
> Also, it's minor but: the description in the PR includes "Expected" and "Actual", and they are different. I would think that a PR should be producing expected results that are the SAME as actual results...

Expected is the intended results post the PR.  Actual is the broken bits the pre-PR code generates.

lane55:
> It's not clear to me why it was necessary to make a copy of ScoreboardUtilities in the WTI-API project. That class is already available from the PC2v9 project, via the pc2.jar. Now, we have two copies of the same file which have to be maintained/kept in symc. Is there an important reason for the copy?

I agree it should not be necessary.   When I attempted to reference ScoreboardUtilities in the WTI-API eclipse project,
there was a compilation error.    If there is a way to not have to duplicate the code - let me know.

lane55:
> Also, it's minor but: the description in the PR includes "Expected" and "Actual", and they are different. I would think that a PR should be producing expected results that are the SAME as actual results...

I agree with Troy.   How would you like it to be formatted/phrased that there is actual (pre-fix) values and
expected (post-fix) values.

troy2914:
> > It's not clear to me why it was necessary to make a copy of ScoreboardUtilities in the WTI-API project. That class is already available from the PC2v9 project, via the pc2.jar. Now, we have two copies of the same file which have to be maintained/kept in symc. Is there an important reason for the copy?
> 
> I agree it should not be necessary. When I attempted to reference ScoreboardUtilities in the WTI-API eclipse project, there was a compilation error. If there is a way to not have to duplicate the code - let me know.

You probably need to checkout WTI-API as a separate project with a class path that includes pc2.jar

lane55:
WTI-API already has a copy of pc2.jar in the path, it is relatively old (1096 days)

A reason I added ScoreboardUtilities  to WTI-API is because there is a eclipse error because ScoreboardUtilities class
is not in the (old,  9.6.0, April 22, 2019, 1096 days ago) pc2.jar pushed as part of the WTI-API.
There is also another eclipse there is a error in WTI-API
> The method getInternalContest() is undefined for the type Contest	ContestController.java	/WTI-API/src/main/controllers	line 556	Java Problem
During the build process new pc2.jars are created but the old pc2.jar is still checked in, so the error in eclipse goes
away because a new pc2.jar is put in place.  
But that also means that in eclipse it looks like there is a bug/error.
I cannot remember what we decided or if we even discussed when to push a new pc2.jar into the sub projects.
The current pc2.jar was pushed Sat Apr 11 18:22:50 2020 -0700
That would be a way to not require a partial copy of ScoreboardUtilities - but that is scope creep for the issue
which we try to avoid.

clevengr:
One solution to the problem of avoiding duplication of code (ScoreboardUtilities) in WTI-API is to change the fix (in `ContestController`) to instead say:
```
		Properties props = internalContest.getContestInformation().getScoringProperties();
		String xmlStandings = dsa.getStandings(internalContest, props, logger);
```
This generates no errors in Eclipse, and it seems totally correct -- the DSA is being given the Scoring Properties defined in the current contest, which fixes the bug.   What this  _doesn't_ do is also obtain properties which are "added" by the `ScoreboardUtilities.getScoringProperties()` method.   I think it's totally reasonable that the DSA should be getting passed to it the properties defined in the current contest...

clevengr (approved):

Issue #373, opened 2022-04-15
Ensure that design flaws are fixed relating to execute and assigning judgements


lane55:
**Describe the issue**: 

There is no known bug in executing and assigning judgements, there may be a design
flaw that should be fixed (asap)

This is a bug about two different sets of code doing the identical
action.  It is a design flaw where if set of code is changed
then other may not be and would cause a significant execution
or judging bug.   There should be only one method/way to 
perform the actions described below.

The code to execute a run should be identical/the same whether
the UI or an AJ is used.

The code to assign a judgement should be the same whether using
the UI and AJ.

**To Reproduce**: 

At this time there is no known way to reproduce a bug.
The current code seems to be identical, this work is to
avoid a bug due to a design/implementation potential flaw.

**Expected behavior**: 

Judgements should be the same for both AJ and UI run.

**Actual behavior**: 

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

lane55:
There are two substituteAllStrings methods in two different classes, Executable and ExecuteUtilities.  Currently
only the method in Executable is used.    The version in ExecuteUtilities is out of date.

Issue #187, opened 2021-02-28
Add a way to map validator judgement text to a judgement


lane55:
**Is your feature request related to a problem?** 

Yes.

**Feature Description**: 

Provide a judgement mapping feature to match a validator judgement
with an existing judgement. 

When a validator returns a judgement string/name, this feature
would map the validator name to a judgement name.

Example mappings

validator: WA
maps to: No - Wrong Answer

validator: Wrong Results
maps to: No - Wrong Answer

This feature would support many mappings.

**Have you considered other ways to accomplish the same thing?** 

Yes.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** 

Similar to Auto Judge data, add a new tab on the UI to add mappings, load mapping from a file,
save (create a file) mappings to a file, and load mappings from
yaml.

**Additional context**:

PR #396, opened 2022-05-12
i387: remove invalid judgements from JSON event feed

### LINKED ISSUE #387: Event Feed JSON contains judgements with undefined acronyms.
**Describe the issue**: 
In the 2022 PacNW contest, the Event Feed JSON output from PC2 was rejected by the Resolver, whose log showed the following:
```
2022.03.05 16:28.32 E Contest event feed has 31 unjudged submissions, cannot run
2022.03.05 16:28.32 E Submission ids: 40, 87, 146, 170, 197, 498, 508, 522, 575, 611, 872, ...
```
Looking at the Event Feed JSON for the contest shows the following entries related to Submission 40:
```
{"type":"submissions", "id":"pc2-246", "op":"create", "data": 
    {"id":"40","language_id":"C++-3918517525028210700","problem_id":"RiseandFall-1","team_id":"104","time":"2022-03-05T11:19:19.472-08",
     "contest_time":"00:19:19.473","entry_point":"F","files":[{"href":"/submissions/40/files"}]
    }
}
...
{"type":"judgements", "id":"pc2-2621", "op":"create", "data": 
    {"id":"Run-1923957307774372492","submission_id":"40","start_time":"2022-03-05T11:19:19.472-08","start_contest_time":"00:19:19.473",
     "judgement_type_id":"IO","end_time":"2022-03-05T11:19:00-08","end_contest_time":"00:19:00.000"
    }
}
```
So Submission 40, which the Resolver claims has "no judgement", actually DOES have a judgement, as shown above; the `judgement_type_id` for it is "IO" (Incomplete Output).

However, the Event Feed also contains the following:
```
{"type":"judgement-types", "id":"pc2-3", "op":"create", "data": {"id":"AC","name":"Accepted","penalty":false,"solved":true}}
{"type":"judgement-types", "id":"pc2-4", "op":"create", "data": {"id":"CE","name":"Compilation Error","penalty":false,"solved":false}}
{"type":"judgement-types", "id":"pc2-5", "op":"create", "data": {"id":"RTE","name":"Run-time Error","penalty":true,"solved":false}}
{"type":"judgement-types", "id":"pc2-6", "op":"create", "data": {"id":"TLE","name":"Time Limit Exceeded","penalty":true,"solved":false}}
{"type":"judgement-types", "id":"pc2-7", "op":"create", "data": {"id":"WA","name":"Wrong Answer","penalty":true,"solved":false}}
{"type":"judgement-types", "id":"pc2-8", "op":"create", "data": {"id":"EO","name":"Excessive Output","penalty":true,"solved":false}}
{"type":"judgement-types", "id":"pc2-9", "op":"create", "data": {"id":"OFE","name":"Output Format Error","penalty":true,"solved":false}}
{"type":"judgement-types", "id":"pc2-10", "op":"create", "data": {"id":"OCS","name":"Other - Contact Staff","penalty":true,"solved":false}}
```
and these are the ONLY judgement types defined in the Event Feed. So the actual problem is that there is no judgement type "IO" defined (which the Resolver then interprets as an "unjudged submission").

Investigation into the issue found that the PacNW contest was indeed configured with exactly (and only) the judgement types shown above.  However, when the CLICS Validator encounters a submission whose execution produces incomplete output (i.e., output that matches the judge up to a certain point but then stops), or produces NO output (which is just the extreme case of "incomplete output"), the Validator returns a `judgement.txt` string (in the "feedback" folder) of "Incomplete Output".  It returns this string irrespective of the fact that "Incomplete Output" is not one of the judgement types configured in the contest.

There is a class `JSONTool` containing a method `convertJudgementToJSON(Run)` which is used in a number of places to generate JSON representations of judgements associated with a given run (Submission).   Method `convertJudgementToJSON(Run)` invokes an external method `CLICSJudgementType.getCLICSAcronymFromDisplayText(String)` which accepts a judgement (result) string (such as "Incomplete Output", "Wrong Answer", or "accepted") returned by a Validator, and returns a corresponding CLICS acronym for the judgement result string.  

Method `CLICSJudgementType.getCLICSAcronymFromDisplayText(String)` contains a mapping of every known CLICS result string to a corresponding CLICS acronym.  However, _it does not pay attention to whether a result string corresponds to a result which is defined for the current contest_.  The result is that calls to this method can get back acronyms which are defined by CLICS (such as "IO") but which are _not_ defined in the current contest.  This is what caused the above issue at the PacNW contest.

It's not yet completely clear what the proper fix for this issue is (although there is a known workaround; see below).
Among the classes which invoke `JSONTool.convertJudgementToJSON(Run)` (and hence invoke `CLICSJudgementType.getCLICSAcronymFromDisplayText(String)`) are:  `services.core.EventFeedJSON`, `services.web.EventFeedStreamer`, and `services.web.JudgementService`.   Any code change which is an effort to fix this bug need to consider possible ramifications on all of these classes.

**To Reproduce**: 
- Start a server, loading `samps/contests/SumitHello`.
- Submit a program which produces _no output_ for the problem (`samps/src/nooutput.c` or `samps/src/NoOutput.java` will work).
- Generate an EventFeedJSON report; examine the listed "judgement_types" and the "judgement_type_id" for the submitted run in the report.

**Expected behavior**:  
The submission should have an acronym ("judgement_type_id") which exists in the list of "judgement_types" in the Event Feed.

**Actual behavior**: 
The submission has a "judgement_type_id" of "IO", which is not defined in the "judgement_types" for the contest.

**Environment**: 
 - Windows 10.1
 - PC2 Version:   any V9.5+ version up to the date this Issue was filed.

**Additional Information**: 
A workaround for this issue is to ensure that every possible return string from the validator being used is defined in the contest, by including it in the `reject.ini` file.  

Strings which can be returned by the CLICS Validator include:
- "Incomplete output"
- "Excessive output"
- "Incorrect output format"
- "Wrong Answer"
- "Runtime Error"
- "accepted"
- 
Strings which can be returned by the PC2 ("Internal") Validator include:
- "accepted"
- "No - Presentation Error"
- "No - Wrong Answer"
- "No - Run-time Error"
- "No - Indeterminant"

lane55:

Removes invalid judgement acronyms (judgements) from event feed JSON.

remove judgement (acronym) mapping from JSONTool to fix invalid
judgement acronyms (acronyms not found in model).

This is the first step in standardizing and formalizing judgement mapping.
More details/step can be found in #373 (only one place to assign judgement and acronym)
and #187 (manage mapping)

### Issue which the PR fixes

#387

Microsoft Windows [Version 10.0.19044.1645]
java version "1.8.0_121"

Start a server, loading samps/contests/SumitHello.
Submit a program which produces no output for the problem (samps/src/nooutput.c or samps/src/NoOutput.java will work).
Generate an EventFeedJSON report; examine the listed "judgement_types" and the "judgement_type_id" for the submitted run in the report.

Expected behavior:
The submission should have an acronym ("judgement_type_id") which exists in the list of "judgement_types" in the Event Feed.

Actual behavior:
The submission has a "judgement_type_id" of "IO", which is not defined in the "judgement_types" for the contest.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/util/JSONTool.java
+++ b/src/edu/csus/ecs/pc2/core/util/JSONTool.java
@@ -14,8 +14,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
-import edu.csus.ecs.pc2.clics.CLICSJudgementType;
-import edu.csus.ecs.pc2.clics.CLICSJudgementType.CLICS_JUDGEMENT_ACRONYM;
 import edu.csus.ecs.pc2.core.IInternalController;
 import edu.csus.ecs.pc2.core.StringUtilities;
 import edu.csus.ecs.pc2.core.Utilities;
@@ -26,6 +24,7 @@ import edu.csus.ecs.pc2.core.model.ClarificationAnswer;
 import edu.csus.ecs.pc2.core.model.ClientType;
 import edu.csus.ecs.pc2.core.model.ContestInformation;
 import edu.csus.ecs.pc2.core.model.ContestTime;
+import edu.csus.ecs.pc2.core.model.ElementId;
 import edu.csus.ecs.pc2.core.model.Group;
 import edu.csus.ecs.pc2.core.model.IInternalContest;
 import edu.csus.ecs.pc2.core.model.Judgement;
@@ -476,55 +475,12 @@ public class JSONTool {
 
             JudgementRecord judgementRecord = submission.getJudgementRecord();
             
-            //commented out; only fetch the judgement if it's needed (see below)
-//            Judgement judgement = model.getJudgement(judgementRecord.getJudgementId());
-            
             // only output its judgement and end times if this is the final judgement
             if (!judgementRecord.isPreliminaryJudgement()) {
 
-                //****************
-                //new code, copied from RunsPane.getJudgementResultString().  
-                //This code has the effect of attempting to use the Validator Result String, 
-                // if it exists, to determine what judgement type (acronym) to put into the JSON Event Feed
-                // rather than defaulting to using the JudgementType out of the Judgement Record directly.  
-                // This was added because PC2 assigns a default judgement type
-                // of "RTE" to any run which is neither a Yes nor a CE.  See Bug xxx.
-                //
-                //Note that this change does not take into account the possibility that a judge changed the judgement
-                //  manually but didn't cause the validator judgement to change.  If there is a validator
-                //  result in the judgementRecord, this code will put into the returned JSON the judgement type corresponding to 
-                //  the validator (not that from the current judgement result).
-                
-                String resultString = null;
-                boolean usingValidator = false;
-                
-                //check if there's a validator result
-                if (judgementRecord.isUsedValidator() && judgementRecord.getValidatorResultString() != null) {
-                    
-                    resultString = judgementRecord.getValidatorResultString();
-                    if (resultString!=null) {
-                        //try to convert the validator string to a known acronym
-                        CLICS_JUDGEMENT_ACRONYM acronym = CLICSJudgementType.getCLICSAcronymFromDisplayText(resultString);
-                        //check if we got back a judgement acronym
-                        if (acronym!=null) {
-                            resultString = acronym.name();
-                            usingValidator = true;
-                        }
-                    }
-                } 
-                
-                if (!usingValidator) {
-                    //no validator result; get the string representation of the judgement itself
-                    Judgement judgement = model.getJudgement(judgementRecord.getJudgementId());
-                    if (judgement != null) {
-                        resultString = judgement.getAcronym();
-                    }
-                }
-                //***************
-                
-                
-//                element.put("judgement_type_id", getJudgementType(judgement));
-                element.put("judgement_type_id", resultString);
+                // Fetch judgement_type_id from judgement acronym
+                String judgmentAcronym = getJudgementAcronymn(judgementRecord);
+                element.put("judgement_type_id", judgmentAcronym);
                 
                 Calendar wallElapsed = calculateElapsedWalltime(model, judgementRecord.getWhenJudgedTime() * 60000);
                 if (wallElapsed != null) {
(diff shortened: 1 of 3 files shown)


johnbrvc (approved):
Tested the code before and after this commit was applied.  Works as described.
The commit description/comment is a bit lacking in explaining how the code that was
changed (removed) actually fixes the problem.  Unless, of course, it was left as an exercise
to the reader, in which case it worked - because I had to figure out how it all worked to
understand WHY the fix addresses the issue #387.  Part of the comment in in the original JSONTool.java was helpful.  Perhaps put part of that comment back to explain what was done and why?  Just a suggestion - shouldn't prevent the merge though.

lane55:
By way of explanation and history...

The fix (for #387) was based on code inserted into the JSONTool
to generate the JSON event feed that replaced existing Judgement
instances/acronyms that potentially would cause a NPE because the selected
acronym might not be in the model (InternalContest) list of Judgements.
When the JSON list of judgements are added to the JSON they are only from
the model.  The particular judgement from the validator was "Incomplete
Output" acronym IO.  When the new code assigned a non-model Judgement
(IO) then the (Resolver) code that loaded the JSON, would not find IO
in the JSON list of judgements because IO was not in the model (list
of judgements) therefore not output as one of the judgements near the
top of the JSON.    On loading a run that was judged IO, the resolver
would throw an NPE (because there was no IO in its list of judgements).
Removing that acronym replacement code avoids the NPE.

The current work around is to load a whole bunch of judgements
(reject.ini) and judgement acronyms to ensure that any replaced acronym
is found in the list of judgments in the model.    The workaround will
not be needed when the PR for this issue is merged.

Like a growing number of re-visited initial requirements/implementations
this fix was made in a good intention fix to standardize the judgements
for runs.

The follow up effort is to revise how validator results are converted
to standardized judgements.  Where nearly everything in that previous
sentence is TBD (needs discussion and decisions about how that will work)

A second and related refactor is that the code to assign a run a judgement
is duplicated.   See #373
There should be only one method coded to assign judgments, but
since the AJ was introduced years ago, never got around to refactoring that.
The solution/implementation should be updated whatever the standardization
of judgements code ends up being.

And some history….  one of the as-is requirements was to present to
the human judges a judgement name/comment from the validator.  So a
Judgement instance had to be used if the name/comment did not match any
existing Judgement name.   Initially the first No was selected which is
CE, well that messed up scoring when CEs were no longer penalized so it
was assigned a non-CE Judgement instance.

John C did a bunch of code work to standardize judgment (acronyms)
remnants of that effort are in class CLICSJudgementType.

clevengr (approved):

PR #403, opened 2022-05-14
i267: additional Q&D patch:  skip delay on "event type == runs".

### LINKED ISSUE #267: Shadow GUI updates take excessively long times to refresh
**Describe the issue**: 
  The `ShadowControlPane` class has multiple Swing buttons, including `Test Connection`,  `Compare Runs`, and `Compare Scoreboards`.  Clicking on either of the "Compare" buttons displays a corresponding Frame -- either a `ShadowCompareRunsFrame` or a `ShadowCompareScoreboardFrame`.  Following the initial button press, it takes an unacceptably long time (as much as 10+ minutes) for the corresponding frame to be displayed.

Similarly, each of the "Compare Frames" has a `Refresh` button which, if pushed, can take up to 10+ minutes to generate an updated screen.

Originally, this delay was being attributed to the separate `RemoteEventFeedMonitor` thread overwhelming the AWT Event Thread.  (The original description of, and extensive discussion about, this suspected cause is retained below. )  

Now however, further testing appears to indicate that the problem is not related to (just) the `RemoteEventFeedMonitor`, because if you start the `RemoteEventFeedMonitor` (by pressing the `Start Shadowing` button) and then do NOTHING until the entire Remote Event Feed has been read and consumed, **subsequently pressing the `Compare Runs` or `Compare Scoreboards` buttons will _still_  demonstrate an unacceptably long delay**.  (Note: determining when the Event Feed has been totally consumed can be accomplished by doing a `tail -f  FEEDER1@site1-0.log` in the Shadow Client's `logs` directory.)

This seems to point at the GUI code for the `Compare Runs` and `Compare Scoreboard` frames/panes.  Using the JVisualVM profiler shows that the AWT EventQueue thread continues to run, with no break, for up to five minutes after the `RemoteEventFeedMonitor` thread has completed the fetching/processing of the remote Event Feed _and has been terminated_.

**To Reproduce**:
- Either use a distribution prior to the inclusion of the fix for Issue #259, or open the source for the current distribution in Eclipse and comment out the line `Thread.sleep(10)` in class `RemoteEventFeedMonitor` (around line 159).
- Start a PC2 server loaded with the NADC21 contest configuration, including Remote CCS information pointing it to the Kattis NADC21 URL:  `https://nadc21.kattis.com/clics-api/contests/nadc21/event-feed`.  It is also necessary to obtain a Kattis account/password which has permission to access this event feed.
- If desired, start one or more AutoJudges judging the NADC21 problems (this is not necessary to demonstrate the issue; not doing this step simply means you'll never see any Shadow comparison judgements from PC2 other than "Pending").
- Start a Feeder client and select the `Shadow Mode` tab.
- Press `Start Shadowing` on the `Shadow Mode` pane.
- Press a `Compare` button on the `Shadow Mode` pane.
- Press the `Refresh` button on the Compare Pane.

**Expected behavior**: 
Pressing either of the `Compare` buttons should display a new frame showing the results of comparing the remote CCS results received so far with the corresponding results computed so far by PC2, and pressing `Refresh` should continually update the compare pane.

**Actual behavior**: 
The Compare Pane will freeze (if it even gets displayed at all), and pressing `Refresh` will have no effect -- not even to highlight the button as an indication that it has been pressed.  If you wait LONG ENOUGH (which in the case of this test is defined as something like 1/2 hour) THEN the GUI will suddenly update properly.

**Environment**: 
 - OS name and version under which the issue was detected: Windows 10 
 - PC2 Version:   see "Steps to reproduce", above.

**Additional Information**
The following facts have been observed using the JVisualVM profiler:
- When the Feeder Client GUI first starts, the AWT EventQueue runs briefly (as expected, to initialize the GUI), but then becomes quiescent (either "Parked" or "Monitor" state; it's not clear which from the coloring in VisualVM).
- Performing innocuous GUI actions, such as switching between GUI tabs, hovering over button components (which causes them to "highlight"), and so forth, DOES wake up the AWT EventQueue thread -- but it only runs for a second and then goes back to quiescent state.
- Pressing the `Start Shadowing` button _and then doing nothing else_ does TWO things:  (1) it creates a `RemoteEventFeedMonitor` thread, which immediately starts fetching events from the Remote CCS; and (2) _it wakes up the AWT EventQueue_ thread.  The former result is expected; the latter is not.
- When the `RemoteEventFeedMonitor` thread receives an "End-of-Updates" state event from the Remote CCS, it dies (assuming we're running updated code which causes it to terminate after getting End-of-Updates).
- The AWT EQ thread _continues to run_ long after the RemoteEFM has terminated.  This is unexpected, and it needs to be investigated.
- The AWT EQ thread _does_ eventually stop running and goes back to the quiescent state (this takes about five minutes, during which time no external inputs to the system were provided).
- Clicking on `Compare Runs` or `Compare Scoreboard` _after_ the AWT EQ thread has gone back to quiescent state causes an _immediate_ display of the corresponding frame.

Taken together, the above seems to indicate that "Start Shadowing" somehow puts a large load on the AWT Event Queue thread, and that this load is not directly related to the RemoteEventFeedMonitor thread's processing.  This, then, is the likely issue with the unacceptable delays described by this Bug/Issue.

## Original Issue Description (maintained here for completeness/posterity) ##

If the event feed being supplied by the Remote CCS comes too fast, the `RemoteEventFeedMonitor` class which reads that event feed overwhelms the JVM thread scheduler, keeping the AWT Event Dispatch thread from getting sufficient cycles to keep the corresponding GUIs (for example, the Test Connection dialog, the `ShadowComparePane`, and even the `ShadowControlPane` itself) updated.

Note:  this is _not_  an issue of putting these GUI update operations at the end of the AWT event queue using `SwingUtilities.invokeLater()`;  all such Shadow GUI operations are _already_ being spawned using `invokeLater()`.  This appears to be an issue of the relative scheduling of JVM threads -- that is, of the AWT Dispatch thread not being able to get enough time to complete its GUI updates in a responsive manner even when using `invokeLater()`.

A partial fix for this issue was implemented in the fix for Issue #259 : a short `Thread.sleep()` was introduced at the start of the `while` loop which processes incoming remote CCS events.  This allows the AWT thread to gain sufficient cycles to keep up with GUI updates (at least for the test case, which was the NADC21 event feed from Kattis, which has over 53,300 events in it).  

However, this fix is inefficient, in that it is causing undesirable delayed processing of remote events just to allow GUI updates.  A more comprehensive fix needs to be found.

**Additional context**: 
Substantial investigation and testing was done regarding this issue.  

The fundamental (root cause) of this issue is probably the fact that the Shadow GUI operations are taking place in the same JVM as the RemoteEventFeedMonitor operations.  This was initially done because the GUI needs access to the Remote Judgements coming from the Remote Event Feed, and the only way to support this _without storing those judgements on the PC2 server_ is to have the GUI components running in the same JVM client as the Event Feed Monitor.  Note however that _storing the remote judgements on the PC2 server would require implementing a new (set of) packets_, and that early in the design of the PC2 Shadow system, _strong objections were voiced to doing this_.  As a result, the entire Shadow system was implemented in a single client (which in retrospect was probably a mistake).

A reasonable guess as to why this is happening is that the `RemoteEventFeedMonitor` thread essentially has a tight "while-loop" processing incoming (remote CCS) events.  If this thread is higher priority than the AWT thread, it starves the AWT thread.  Even if it is lower priority than the AWT thread (see below), it still seems to be the case that the `RemoteEventFeedMonitor` thread is able to _effectively_ starve the AWT thread.  This in turn probably happens because the `RemoteEventFeedMonitor` thread is _always ready to run_ (state==`Runnable`); if the AWT thread blocks for _any reason_ while doing GUI updates then the Monitor thread will be allowed to run (even though it's lower priority), and will at least get an entire timeslice.  

Further, I'm pretty sure there is something in the Java specification which says that this lower-priority thread _may be allowed to keep running even though a higher priority thread is Runnable_ (that is, the Java Language Specification does NOT require "preemptive thread scheduling".  This may be relevant in this case.

I looked at various alternatives (short of redesigning the entire Shadow system with new PC2 packets, which I wasn't going to undertake at this time).

- I tried adding a `Thread.yield()` each time around the `RemoteEventFeedMonitor` while-loop.  This did not help.

- I tried adding code to increase the priority number (so, lower the priority) of the `RemoteEventFeedMonitor` thread so that it was lower priority than the AWT Event Dispatch thread, using code like the following at the top of the `RemoteEventFeedMonitor` thread:
``` java
    private static int awtDispatchThreadPriority = 0;
   ...
        //find the priority of the AWT Thread
        EventQueue.invokeLater(() -> {
            int dispatchThreadPriority = Thread.currentThread().getPriority();
            awtDispatchThreadPriority = dispatchThreadPriority;
        });        
        
        while (awtDispatchThreadPriority==0) {
            try {
                Thread.sleep(200);
            } catch (InterruptedException e) {
            }
        }

        //give our thread lower priority than the AWT Event Dispatcher, so AWT can keep ahead when updating GUIs...
        int newPriority = awtDispatchThreadPriority + 1;
        Thread.currentThread().setPriority(newPriority);
```
  This seemed to _help_, but it did not solve the problem.

- A more comprehensive solution might be to maintain a "moving average" of the rate of remote event processing, and introduce variable sleeps in the Remote Event Feed Monitor -- but only when the average event processing rate exceeds some threshold.  I even developed some simple code to support something like this:

```java
package edu.csus.ecs.pc2.shadow;

/**
 * This class supports maintaing a "moving average" of integer values.
 * It maintains a circular buffer of a user-specified size, and provides for adding values
 * to the buffer and for obtaining the (current) average value in the buffer.
 * When the buffer is full, adding a new value replaces the oldest value in the buffer
 * and updates the average.
 *  
 * @author John Clevenger, PC2 Development Team (pc2@ecs.csus.edu)
 */
public class MovingAverage {
    
    private int [] buffer;
    private int currentCount, insertPos;
    private long sum;

    /**
     * Constructs a MovingAverage holding "bufferSize" unique values.
     * 
     * @param bufferSize the number of values considered for averaging.
     */
    public MovingAverage(int bufferSize) {
        buffer = new int[bufferSize];
        insertPos = 0;
        sum = 0;
    }
    
    /**
     * Adds a new value to the buffer.  If the buffer currently has fewer than bufferSize entries,
     * the new value is added at the end of the buffer; otherwise, the new value replaces the oldest
     * value in the buffer.  Adding a new value automatically updates the value returned by method
     * {@link #add(int)} appropriately.
     * 
     * @param val the new value to be added to the buffer.
     * 
     * @see MovingAverage#getAverage()
     */
    public void add(int val) {
        //keep track of how many values are currently in the buffer
        if (currentCount < buffer.length) {
            currentCount++;
        }
        //remove the oldest value from the sum (note that prior to the buffer filling up, the buffer positions all contain zero)
        sum -= buffer[insertPos];
        //add the new value into the running buffer sum
        sum += val;
        //put the new value into the next buffer position
        buffer[insertPos] = val;
        //increment the buffer pointer, wrapping around at the end of the buffer array
        insertPos = (insertPos + 1) % buffer.length;
    }
    
    /**
     * Returns the current (moving) average of the buffer values.  
     * 
     * @return a double containing the average of the values currently in the buffer.
     */
    public double getAverage() {
        return (double)sum / currentCount;
    }

}
```
and prior to the start of the while-loop:
``` java
//                //create a moving-average queue for events.  This queue allows throttling the Remote Event Feed Monitor's handling
//                // of events when the input event rate gets too high.
//                MovingAverage eventRateAverage = new MovingAverage(EVENT_RATE_WINDOW_SIZE);
//                
//                //create a Timer Task to sample the input event rate
//                TimerTask eventRateSampler = new TimerTask() {
//                    
//                    public void run() {
//                        System.out.println ("EventRateSampler invoked...");
//                        
//                        //TODO: add code below to count events
//                        //TODO: add code here to reset event count and to check if rate threshold has been exceeded; if so, set a flag
//                        //TODO: add code below to check the flag and insert a delay (then clear the flag)
//                    }
//                };
//                
//                Timer eventRateTimer = new Timer("EventRateTimerThread");
//                eventRateTimer.scheduleAtFixedRate(eventRateSampler, 1000L, EVENT_SAMPLE_RATE_SECS * 1000L);
//                
//                //TODO: add code (somewhere) to CANCEL the eventRateTimer when done...
```

However, I didn't pursue this avenue any further, choosing instead to implement the simple addition of a sleep() in the `RemoteEventFeedMonitor` (see Issue #259).

clevengr:

Skips event feed delay for events of type "runs".

### Issue which the PR fixes
Partial attempt to fix #267

### DIFF
--- a/src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
+++ b/src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
@@ -167,7 +167,7 @@ public class RemoteEventFeedMonitor implements Runnable {
                     //See also GitHub Issue 267:  https://github.com/pc2ccs/pc2v9/issues/267
                     // Thread.sleep(10);
                     
-                    if (!event.contains("\"type\": \"organizations\"")) {
+                    if ((!event.contains("\"type\": \"organizations\"")) && (!event.contains("\"type\": \"runs\""))) {
                         Thread.sleep(500);
                     }
                     


lane55 (approved):
looks good

PR #411, opened 2022-05-20
I 275 write tsv from event feed new tool

### LINKED ISSUE #275: Update team numbers in pc2 from a CLICS event feed

More of an inefficiency

**Feature Description**: 

Update team numbers based on team numbers in a CLICS event feed.

The CMS Team Id (aka external team id) will be used to match
the team account, then the team number can be updated/assignged.

Yes, there is no workaround.   

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, 
Create a command line tool that will update an existing teams.tsv
and update the team numbers.

Another option is to add a UI feature that will read the event feed and
update the team numbers.

**Additional context**: 

Note that connection information can be read from CDP/config files or
from the ContestInformation class.

lane55:

Adds tool that will extract from an event feed json a new.teams.tsv or
new.accounts_load.tsv.  Can be used to update team name and other data.

### Issue which the PR fixes

#275 partial

Microsoft Windows [Version 10.0.19044.1706]
java version "1.8.0_121"

To create a new.teams.tsv file use:
pc2tools --efx --file event-feed.json

To create a new.accounts_load.tsv use: (--offal option)
pc2tools --efx --offal --file event-feed.json

test using \testdata\EventFeedLoaderTest\partial-event-feed-0519-1231.json
or any other event feed json.


src/edu/csus/ecs/pc2/core/imports/clics/EventFeedLoader.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class EventFeedLoader {
+
+    private static ObjectMapper mapper = new ObjectMapper();
+    
+    static {
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    }
+
+    public static List<TeamAccount> createTeamAccounts (String [] lines) throws JsonParseException, JsonMappingException, IOException {
```
clevengr, line 27: I suggest this method be renamed to "getTeamAccounts".  I think "createTeamAccounts" makes it easy to misinterpret that this method will somehow actually create accounts in PC2.

src/edu/csus/ecs/pc2/core/list/TeamAccountComparator.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class TeamAccountComparator implements Comparator<TeamAccount>, Serializable {
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = 6930011395916416715L;
+
```
clevengr, line 22: This is just a personal opinion, but I think randomly-generated serialVersionUIDs are at best useless and at worst misleading.   I think the first version of a class should have serialVersionUID 1; if the class is changed so that it is no longer serialization-compatible the version should be bumped to 1, etc.   (With a serialVersionUID like 6930011395916416715L. what would be the appropriate "next serialVersionUID" value when the class is upgraded?)

src/edu/csus/ecs/pc2/exports/ccs/CLICSEventFeedExract.java
```diff
@@ -0,0 +1,415 @@
...
+import edu.csus.ecs.pc2.shadow.IRemoteContestAPIAdapter;
+import edu.csus.ecs.pc2.shadow.RemoteContestAPIAdapter;
+
+/**
+ * Extract data from event feed, output data to file
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class CLICSEventFeedExract {
```
clevengr, line 40: Please correct the spelling of this class name ("...Extract", not "...Exract").

src/edu/csus/ecs/pc2/core/list/TeamAccountComparator.java
lane55: A serialVersionUID  is not useless, it is critical to support non-system-breaking changes in classes.
A UID is not necessarily a sequential id.   If is a different Id.   If there were no use for a random id why would
eclipse provide a eclipse Fix with a random UID?

src/edu/csus/ecs/pc2/core/imports/clics/EventFeedLoader.java
lane55: renamed.

src/edu/csus/ecs/pc2/exports/ccs/CLICSEventFeedExract.java
lane55: fixed name/typo

johnbrvc (approved):
Tested.  Worked for me.  I'm not in love with the extremely long, arcane option name -offal, which I typed wrong twice, even after I looked at the help: two o's, then one f.  I might have just added a "--al".  Or, an even more generic output format option for possible future use as well: "--of"  and that could take an arg of "t[eams]" or "a[ccountload]", eg  --of t, or --of a.  Or, just use the "posix" way:  --of=t  or --of=a  or even for those verbose posix people: --output-format=teams or --output-format=account-load
(not a fan of the last two, but posix is).  But, the current args are not posix complaint anyway, so...

Also, am I correct in assuming that scope of usefulness of these new options are primarily for  loading shadows?

src/edu/csus/ecs/pc2/core/list/TeamAccountComparator.java
clevengr: I didn't say that a serialVersionUID is useless.  I said that a _**randomly-generated**_ SerialVersionUID is useless (or worse, misleading.  I totally agree that having a SVUID is important; I just think that using random ones obfuscates one of the purposes, which is to help the programmer keep track of "versions".   If I have a SVUID of "1", and I upgrade the class and raise the SVUID to "2", that makes if very easy for others to know "which came first".   If I have a SVUID of 6930011395916416715L and I upgrade the class and change it to 6730411565912316751L, a programmer looking at those two numbers learns _nothing_ about the history of the revision sequence.  That's all I'm saying.  You said "A UID is not necessarily a sequential id. If is a different Id. "  That's my point:  making it different _and sequential_ provides additional information that's lost if it's _just different_.

As for why Eclipse provides a random UID, who knows?   I certainly would be hesitant to use "because Eclipse does it that way" as a justification for _anything_  :)

In any case, as I said in the opening line, "this is just a personal opinion".  If you agree, please change the UID and mark this comment resolved.  If you disagree, please leave the UID unchanged and likewise mark the comment resolved.

src/edu/csus/ecs/pc2/exports/ccs/CLICSEventFeedExract.java
clevengr: Hmmm...I'm not seeing any change.  Maybe I'm using GitHub's facilities wrong...

In any case, if the requested change has been made, please go ahead and mark this comment "resolved".

lane55:
no plans to update this PR, it has not been a priority for 8+ months.

PR #433, opened 2022-06-05
i431: convert connections and logins panes to use JTable

johnbrvc:
Before submitting your Pull Request, please make sure you have read the [Guidelines for Submitting Pull Requests](https://github.com/pc2ccs/pc2v9/wiki/Guidelines-for-Submitting-Pull-Requests).  Then please provide the following information:

### Description of what the PR does
Adds 2 new classes: ConnectionsTablePane and LoginsTablePane to ulitmately replace ConnectionsPane and LoginsPane.
The new classes make use of JTable to allow a scroll pane.  May also be more efficient than MCLB.  This is part of the
effort to remove MCLB from the code.  In addition, this was added in an attempt to help the ACPC alleviate their
issues with thousands of logins.

### Issue which the PR fixes
#431 
### Environment in which the PR was developed (OS,IDE, Java version, etc.)
Application     : CSUS Programming Contest Control System
Version         : 9.8build 20220510
Build Date      : Tuesday, May 10th 2022 17:29 UTC
Build Number    : 6023

OS              : Windows 11 10.0 (amd64)
Java Version    : 1.8.0_321

Today's Date    : Sun Jun 05 11:25:24 EDT 2022

Version 9.8build 20220510 (Tuesday, May 10th 2022 17:29 UTC) Java ver 1.8.0_321 build 6023 Windows 11 10.0 (amd64) 

Eclipse: Version: 2021-12 (4.22.0)

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)
From the PC2 server (or administrator), view the Connections and/or Login tabs.  These should be scrollable (using the mouse wheel.).

I would have like to do more testing with more logins (I only tested with 10 logins).

PR #434, opened 2022-06-06
I432 command line suppress connections and logins grids

### LINKED ISSUE #432: Temporary support for command-line suppression of Connections and Logins grids on Admin.
**Describe the issue**: 
The ACPC 2022 Kickoff Contest died because of issues with PC2.  The contest had 3300+ teams and had received 6700+ submissions by the time it was called off.  The root issue seems to have been that the server crashed (probably due to OOM) and failed to write a large number of RunFiles to disk.  This in turn caused errors when attempts to fetch those runs were made.  At the same time, the Admin became totally unusable ("frozen") so no investigation or recovery attempts were possible.

A detailed investigation by a number of people spread across 9 time zones seems to indicate that the issue with the Admin freezing was the MCLB grids failing to handle the large number of WTI connections and logins.  Removing those two grids from the AdminView allowed a recreation of the contest to function smoothly (i.e., without the Admin locking up).

A longer-term solution needs to be found -- for example, replacing all uses of MCLB with JTables.  However, in order to get the next ACPC contest running in a few days, a workaround is to provide the ability to start the Admin without those two grids present.

**To Reproduce**: 
- Start up the ACPC 2022 Kickoff Contest on AWS, then start an Admin.  

**Expected behavior**: 
The Admin GUI should be responsive.

**Actual behavior**: 
The Admin GUI locks up (doesn't respond to input).

clevengr:
Adds two command-line options:  `--noconnectionspane` and `--nologinspane`.  When one (or both) of these options are specified on starting an Admin from the command line, the appearance of the corresponding grid (_Connections_ and _Logins_ respectively) is suppressed.  This provides a (brute-force) workaround for issues where these grids take so long to load that they effectively freeze the Admin.

Note: it is possible that PR #433, which replaces the old MCLB grids in AdminView with newer "JTable" versions, obviates the need for the fixes in this PR.  However, if it turns out that #433 does _not_ completely eliminate the problem described in Issue #432, this PR is needed to provide a rapid-response workaround for upcoming contests such as ACPC.  In any case, the PR is benign if the new options are not used.

### Issue which the PR fixes
Fixes #432, which was filed in an effort to provide the ACPC contest with a viable workaround when running extremely large contests (>3000 teams).

The code in the PR passes all JUnit tests, and a local build generated from the PR works as intended.

Windows 10.1; Eclipse Version: 2020-12 (4.18.0) Build id: 20201210-1552; Java 1.8.0_271.

- Start a server loading a contest, e.g. "SumitHello".
- Login a team (e.g. team1).
- Start an Admin using defaults.
- Observe the presence of the "Connections" and "Logins" tabs on the "Run Contest" tab.
  - Note the presence of both Connection and Login information (on the respective tabs) for "team1".
- Shut down the Admin.
- Restart the Admin, specifying either `--noconnectionspane` and/or `--nologinspane` as arguments on the command line.
  - Note the _absence_ of either or both of the "Connections" and/or "Logins" tabs on the "Run Contest" tab.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/model/IFileImpl.java
+++ b/src/edu/csus/ecs/pc2/core/model/IFileImpl.java
@@ -2,8 +2,6 @@ package edu.csus.ecs.pc2.core.model;
 
 import java.util.Base64;
 
-import edu.csus.ecs.pc2.core.model.IFile;
-
 /**
  * This class is a concrete implementation of a file from the point of view of the PC<sup>2</sup> API.
  * 
--- a/test/edu/csus/ecs/pc2/tools/PasswordGeneratorTest.java
+++ b/test/edu/csus/ecs/pc2/tools/PasswordGeneratorTest.java
@@ -4,8 +4,6 @@ package edu.csus.ecs.pc2.tools;
 import java.util.List;
 
 import edu.csus.ecs.pc2.core.util.AbstractTestCase;
-import edu.csus.ecs.pc2.tools.PasswordGenerator;
-import edu.csus.ecs.pc2.tools.PasswordType2;
 
 /**
  * Unit test.
--- a/src/edu/csus/ecs/pc2/core/IInternalController.java
+++ b/src/edu/csus/ecs/pc2/core/IInternalController.java
@@ -292,6 +292,18 @@ public interface IInternalController {
      * @return true if using GUI, false if using text only
      */
     boolean isUsingGUI();
+    
+    /**
+     * Is this controller suppressing display of Connections grids.
+     * @return true if GUIs using this controller should suppress display of Connections grids.
+     */
+    boolean isSuppressConnectionsPaneDisplay();
+
+    /**
+     * Is this controller suppressing display of Logins grids.
+     * @return true if GUIs using this controller should suppress display of Logins grids.
+     */
+    boolean isSuppressLoginsPaneDisplay();
 
     /**
      * Update site in contest, send site update packet to other servers and admins.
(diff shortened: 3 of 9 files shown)


src/edu/csus/ecs/pc2/core/model/IFileImpl.java
```diff
@@ -2,8 +2,6 @@
 
 import java.util.Base64;
 
-import edu.csus.ecs.pc2.core.model.IFile;
```
troy2914, line 5: the changes to this file seem unrelated to this ticket.

test/edu/csus/ecs/pc2/tools/PasswordGeneratorTest.java
```diff
@@ -4,8 +4,6 @@
 import java.util.List;
 
 import edu.csus.ecs.pc2.core.util.AbstractTestCase;
-import edu.csus.ecs.pc2.tools.PasswordGenerator;
-import edu.csus.ecs.pc2.tools.PasswordType2;
```
troy2914, line 8: more unrelated changes.

troy2914 (approved):
Approving, though confused about unrelated cleanup in a couple of files.

src/edu/csus/ecs/pc2/core/model/IFileImpl.java
clevengr: The relationship of the changes in this file to the ticket is simply that the unchanged version of this file generates Errors in Eclipse -- and I didn't want to submit a PR with code that has Errors flagged in it.  I suspect the underlying issue might be that I'm developing in a different Eclipse environment -- my laptop, in Sharm -- from what I normally use.  However, I tried but failed to figure out what is different, and given the urgency for ACPC of getting this ticket resolved I had two choices -- check in code with Errors, or fix the item causing the Error.  (Yes, I could have filed a separate ticket just for this -- but I didn't have time to do that and get the code working and tested.

test/edu/csus/ecs/pc2/tools/PasswordGeneratorTest.java
clevengr: Same response as to the previous comment.

PR #440, opened 2022-06-07
i438: Autojudges hang if a run's related files are missing

johnbrvc:
If an exception occurs fetching a run's files, then set the run's status to "NEW" so AJ will not pick it up again.  Also notify the the client that requested the run that it is unavailable.

Handle the read-only case for this as well.

Catches any exception that getRunFiles() may throw (like file not found) and changes the run status to NEW (so no AJ's
will pick it up), and then notifies the client that the run is unavailable.
### Issue which the PR fixes
Autojudge's hang fetching run files for runs that don't have any since the status is QUEUED_FOR_COMPUTER_JUDGEMENT, 
all AJ's will attempt to check out the run, but hang, waiting for the files if they  don't exist.

Java Version    : 1.8.0_321

(From @ashoos72 ):
Start a server
Start admin
Configure a few teams/judges
Configure 1 hello world problem with an answer file and computer judging (using pc2 diff or whatever)
Configure java
Configure 1 aj for the problem
start the contest

Start a team
submit three hello world solutions

go to a server db.1 folder and delete s1r2.files (the run files for submission 2)

start the AJ
You will see the AJ judge the first submission and then not hang on submission 2 but mark it as NEW, and continue on to submission 3.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/PacketHandler.java
+++ b/src/edu/csus/ecs/pc2/core/PacketHandler.java
@@ -3545,11 +3545,11 @@ public class PacketHandler {
                     } catch (Exception e) {
                         controller.getLog().warning("contest.getRunFiles (R/O) can not get files for run " + run.getNumber() + ": " + e.getMessage());
                         
-                        try {
-                            contest.cancelRunCheckOut(run, whoRequestsRunId);
-                        } catch (Exception e1) {
-                            controller.getLog().severe("Problem canceling run checkout after error getting run " + run.getNumber() + " files." + e1);
-                        }
+                        // set status to NEW indicating there was a failure and it has to be manually taken care of
+                        // the judges will be notifed of a new run.
+                        theRun.setStatus(Run.RunStates.NEW);
+                        Packet availableRunPacket = PacketFactory.createRunAvailable(contest.getClientId(), whoRequestsRunId, theRun);
+                        controller.sendToJudgesAndOthers(availableRunPacket, true);
                         
                         Packet notAvailableRunPacket = PacketFactory.createRunNotAvailable(contest.getClientId(), whoRequestsRunId, theRun);
                         controller.sendToClient(notAvailableRunPacket);
@@ -3579,12 +3579,13 @@ public class PacketHandler {
                             controller.getLog().warning("contest.getRunFiles can not get files for run " + run.getNumber() + " (settng to status NEW): " + e.getMessage());
                             
                             try {
+                                // cancel the checkout and set run state to NEW to notify judges
                                 contest.cancelRunCheckOut(run, whoRequestsRunId);
                                 theRun.setStatus(Run.RunStates.NEW);
                                 Packet availableRunPacket = PacketFactory.createRunAvailable(contest.getClientId(), whoRequestsRunId, theRun);
                                 controller.sendToJudgesAndOthers(availableRunPacket, true);
                             } catch (Exception e1) {
-                                controller.getLog().severe("Problem canceling run checkout after error getting run " + run.getNumber() + " files." + e1);
+                                controller.getLog().severe("Problem cancelling run checkout after error getting run " + run.getNumber() + " files." + e1);
                             }
                             
                             Packet notAvailableRunPacket = PacketFactory.createRunNotAvailable(contest.getClientId(), whoRequestsRunId, theRun);
@@ -3597,7 +3598,7 @@ public class PacketHandler {
                             try {
                                 contest.cancelRunCheckOut(run, whoRequestsRunId);
                             } catch (UnableToUncheckoutRunException e) {
-                                controller.getLog().severe("Problem canceling run checkout after error getting run files.");
+                                controller.getLog().severe("Problem cancelling run checkout after error getting run files.");
                             }
                             throw new RunUnavailableException("Error retrieving files.");
                         }
--- a/src/edu/csus/ecs/pc2/core/PacketHandler.java
+++ b/src/edu/csus/ecs/pc2/core/PacketHandler.java
@@ -3543,7 +3543,7 @@ public class PacketHandler {
                     try {
                         runFiles = contest.getRunFiles(run);  
                     } catch (Exception e) {
-                        controller.getLog().info("contest.getRunFiles (R/O) can not get files for run " + run.getNumber() + ": " + e.getMessage());
+                        controller.getLog().warning("contest.getRunFiles (R/O) can not get files for run " + run.getNumber() + ": " + e.getMessage());
                         
                         try {
                             contest.cancelRunCheckOut(run, whoRequestsRunId);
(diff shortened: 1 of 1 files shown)


src/edu/csus/ecs/pc2/core/PacketHandler.java
```diff
@@ -3536,7 +3536,26 @@ private void checkoutRun(Packet packet, Run run, ClientId whoRequestsRunId, bool
...
                     theRun = contest.getRun(run.getElementId());
-                    RunFiles runFiles = contest.getRunFiles(run);
+                    
+                    RunFiles runFiles = null;
+                    
+                    // in case GetRunFiles throws an exception we want to deal with it separately
+                    try {
+                        runFiles = contest.getRunFiles(run);  
+                    } catch (Exception e) {
+                        controller.getLog().info("contest.getRunFiles (R/O) can not get files for run " + run.getNumber() + ": " + e.getMessage());
```
clevengr, line 3546: I don't think this should be an "info()" log message; the existence of a Run with no RunFiles is clearly an error of some kind.  I would suggest at least type "warning()".

src/edu/csus/ecs/pc2/core/PacketHandler.java
```diff
@@ -3551,7 +3570,29 @@ private void checkoutRun(Packet packet, Run run, ClientId whoRequestsRunId, bool
...
 
-                        RunFiles runFiles = contest.getRunFiles(run);
+                        RunFiles runFiles = null;
+                        
+                        // in case GetRunFiles throws an exception we want to deal with it separately
+                        try {
+                            runFiles = contest.getRunFiles(run);  
+                        } catch (Exception e) {
+                            controller.getLog().info("contest.getRunFiles can not get files for run " + run.getNumber() + " (settng to status NEW): " + e.getMessage());
+
```
clevengr, line 3580: As above, I think this is ALWAYS indicative of problem (can't getRunFiles() for some reason) so this should be (at least) a `warning()`, not just `info()`.

src/edu/csus/ecs/pc2/core/PacketHandler.java
```diff
@@ -3536,7 +3536,26 @@ private void checkoutRun(Packet packet, Run run, ClientId whoRequestsRunId, bool
...
+                    RunFiles runFiles = null;
+                    
+                    // in case GetRunFiles throws an exception we want to deal with it separately
+                    try {
+                        runFiles = contest.getRunFiles(run);  
+                    } catch (Exception e) {
+                        controller.getLog().info("contest.getRunFiles (R/O) can not get files for run " + run.getNumber() + ": " + e.getMessage());
+                        
+                        try {
+                            contest.cancelRunCheckOut(run, whoRequestsRunId);
```
clevengr, line 3549: For the "non-readOnly" case (below), after calling `cancelRunCheckOut()` the code sets the run Status to `NEW` (and sends an "available NEW run" packet to judgedAndOthers.  This was what I think we all agreed should happen (for the non-readOnly case.  Is there a reason for NOT doing that here?

clevengr (commented):
I've made a couple of comments (regarding Logging Type) that could arguably be ignored (declined).  However, I'd like someone to comment on the one note about whether the "readOnly" case should (or should not) include changing the Run State and notifying "judgesAndothers".

clevengr:
> There was some discussion on this on the slack channel between me and Sam. Here it is: John Buck 12:26 PM BTW in the case of R/O I took out the code to set the status to NEW, since, the AJ won't ever ask for a R/O run, only an interactive user would. so the run status, whatever it is, should probably stay as it is. If someone requests a R/O (bad) run with status "queued for computer judgement", it's fine to leave it as is since the AJ will eventually set the status to NEW.
> 
> Samir Ashoo 12:28 PM That works, but I disagree :) I think the run should be flagged as bad "NEW" the instant know about it.
> 
> John Buck 12:29 PM Maybe if the current status is "queued for computer judgement"... 12:31 Ill do a PR when I get back from lunch.
> 
> Samir Ashoo 12:31 PM No. Again, why delay to alert the sysadmin that there may be an issue. It may be something they can deal with and if we wait until someone checks out the run, it may be too late to recover. 12:31 But again, for now this works.
> 
> John Buck 12:37 PM Ok, but what other status can it be? It's either been judged (probably not possible), or "NEW".
> 
> Samir Ashoo 12:38 PM I would just put it to "NEW". This is the catchall something went wrong for a problem that is configured for Auto Judging. I think eventually we need to add a new status like MISSING_DISK_FILES and use that instead.

I think I agree with Sam; there doesn't seem to be any good reason NOT to change the status as soon as we know there's a problem with the run.  Eventually we need a new/different status -- something like "BAD_RUN" or Sam's "MISSING_DISK_FILES" (although I think the latter might be misleading because we don't know that missing disk files is the ONLY reason for an exception at this point).  In any case, unless you (JohnB) wish to make a strong argument disagreeing, I suggest we have the code change the "readOnly" case exactly the same way it changes the "non-readOnly" case.  

Can you either reply with a counter-argument or else push a change to the PR?  (And change the logging types while you're at it :)

johnbrvc:
My only concern was losing a run status that was already set.  ex. a run WAS valid and judged "Accepted".  Then, the admin goes to look at the run.  For some reason, the run files got deleted/corrupted/protected or are otherwise unavailable.  The judgment would then be changed to "NEW", losing the "Accepted".  

Other than that I don't have any objections, as long as everyone knows the implications are.   

I already changed the ".info" to ".warning".  I don't think it warrants a "severe" since things can "go on", unless there are objections?  "severe" is used in the same area code if the run checkout can't be cancelled, which is probably a very bad thing.

I will add the "NEW" code to the R/O case there.  

If that's all, I'll push the update.

clevengr:
> My only concern was losing a run status that was already set. ex. a run WAS valid and judged "Accepted". Then, the admin goes to look at the run. For some reason, the run files got deleted/corrupted/protected or are otherwise unavailable. The judgment would then be changed to "NEW", losing the "Accepted".

I don't think we would lose anything, because the (prior) _judgement_ is retained in the `RunResultsFiles` which are included in the processing of the run.  Changing the _state_ to NEW doesn't affect the fact that the `RunResultsFiles` still have the (hypothetical) "Accepted".

There IS a possibility that the `RunResultsFiles` could be _missing_ even though the `RunFiles` were present; we probably should be surrounding the calls to `getRunResultFiles()` with similar try/catch blocks.  But I think the likelihood of that happening is small enough that we should go ahead with the fix as proposed, in order to get it done for ACPC.

> I already changed the ".info" to ".warning". I don't think it warrants a "severe" since things can "go on", unless there are objections? "severe" is used in the same area code if the run checkout can't be cancelled, which is probably a very bad thing.

I agree; `warning()` (but not `severe()`) is the appropriate change.

> If that's all, I'll push the update.

yes, please do!

clevengr (approved):

Issue #445, opened 2022-06-10
AJs are churning trying to fetch next run


lane55:
**Describe the issue**: please provide a clear and concise description of what the bug is.  

When many AJs running, need to keep AJs from constantly all requesting the same run.

Need to find a better way to avoid AJs requesting to judge the same run

**To Reproduce**: please describe, in order, the steps to reproduce the behavior:

**Expected behavior**: please provide a description of what you *expected* to happen when you took the above steps:  

**Actual behavior**: please describe what _actually_ happened when you took the above steps:

**Environment**: please complete the following information:
 - OS name and version under which the issue was detected (e.g. Windows 10 or Ubuntu 18.04):
 - PC2 Version (e.g. 9.7.0):   
 - Browser (if the bug is related to web-based team clients) (e.g. _Chrome Version 75.0.3770.142_ or _Safari Version 11.1.2_):

**Log Info**:  PC^2 generates logs (in the **logs** folder for client modules; under the **profiles** folder for the PC2 Server).  Please review the relevant logs and provide copies of them as attachments -- particularly if the logs show Exceptions being thrown.  

**Screenshots**:  if applicable, please add attachments showing screenshots to help explain the issue.

**Additional context**: please add any other context about the issue here.

clevengr:
There are a large number of comments which relate to THIS issue (#445 - the AJ's "churning" when fetching runs) that were made under PR #447 (Change AJ sort order) which fixed somewhat-related-but-not-identical Issue #444.  Since that PR was approved and merged (fixing JUST the sort order, not the churning) and the corresponding issue was closed, I'm copying those comments which really should be under this issue, to this issue.  This will help keep the discussion about this issue in one single place.

From JohnC:
```
the current code has every client (judges, plus any other client such as a "queue display" fetching every single run 
every time a new run is needed, sorting those (maybe THOUSANDS) of runs, 
and selecting the best candidate from the sorted results. 
That is LOT of overhead; we might want to consider alternatives.

For example, runs could be maintained on the Server in a standard Priority Queue based on some filter criteria; 
a request for the "next available run" would then be a simple removal from the head of the Priority Queue. 
This is just one idea; maybe there are better ones. But I think we're should talk about it.
```

From Sam:
```
I think we should change the whole concept of AJ fetch.

We have two different issues that need to be addressed: [#1](https://github.com/pc2ccs/pc2v9/pull/1) -- multi-site &
[...]

PR #447, opened 2022-06-10
i444: change AJ fetch sort order

### LINKED ISSUE #444: Change AJ fetch next unjudged run order

AJs should request runs based on earliest TIME, not site or runId.   Currently the AJ fetches all runs in model,
sorts them by site # and run id then fetches the first 

**Feature Description**: please provide a clear and concise description of the proposed feature:  

Change the sort to by run elapsed time (ms), then run id, then site id.   It is very likely that the elapsed ms will
always be different and the secondary and tertiary comparisons will not be necessary

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, please describe.  

Add new Comparator to sort runs.

**Additional context**:

lane55:

AJ will now fetch runs ordered by run elapsed ms, then run number, then site number

### Issue which the PR fixes

#444 

All.

Start 2 servers/contest clocks.
Submit runs to site 2 first
pause for a couple of minutes
Submit runs to site 1.

Start a single AJ to judge runs

Before
Site 1 runs would be judged first.

Expected
Runs from site2 should be selected/judged first.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/list/RunComparatorByElapsedRunIdSite.java
+++ b/src/edu/csus/ecs/pc2/core/list/RunComparatorByElapsedRunIdSite.java
@@ -1,0 +1,35 @@
+// Copyright (C) 1989-2022 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.core.list;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import edu.csus.ecs.pc2.core.model.Run;
+
+/**
+ * Run Comparator, Order the runs by elapsed time (ms), run id, then site.
+ *
+ * @version $Id$
+ * @author pc2@ecs.csus.edu
+ */
+
+public class RunComparatorByElapsedRunIdSite implements Comparator<Run>, Serializable {
+
+    private static final long serialVersionUID = 1629965664816187817L;
+
+    public int compare(Run run1, Run run2) {
+
+        if (run1.getElapsedMS() == run2.getElapsedMS()) {
+            
+            if (run1.getNumber() == run2.getNumber()) {
+
+                return Long.compare(run1.getSiteNumber(), run2.getSiteNumber());
+
+            } else {
+                return Long.compare(run1.getNumber(), run2.getNumber());
+            }
+        } else {
+            return Long.compare(run1.getElapsedMS(), run2.getElapsedMS());
+        }
+    }
+}
--- a/src/edu/csus/ecs/pc2/ui/AutoJudgingMonitor.java
+++ b/src/edu/csus/ecs/pc2/ui/AutoJudgingMonitor.java
@@ -15,7 +15,7 @@ import edu.csus.ecs.pc2.core.IInternalController;
 import edu.csus.ecs.pc2.core.execute.Executable;
 import edu.csus.ecs.pc2.core.execute.ExecutionData;
 import edu.csus.ecs.pc2.core.execute.JudgementUtilites;
-import edu.csus.ecs.pc2.core.list.RunComparator;
+import edu.csus.ecs.pc2.core.list.RunComparatorByElapsedRunIdSite;
 import edu.csus.ecs.pc2.core.log.Log;
 import edu.csus.ecs.pc2.core.model.ClientId;
 import edu.csus.ecs.pc2.core.model.ClientSettings;
@@ -187,7 +187,7 @@ public class AutoJudgingMonitor implements UIPlugin {
         }
 
         Run[] runs = contest.getRuns();
-        Arrays.sort(runs,new RunComparator());
+        Arrays.sort(runs,new RunComparatorByElapsedRunIdSite());
 
         for (Run run : runs) {
             if (run.getStatus() == RunStates.QUEUED_FOR_COMPUTER_JUDGEMENT) {


johnbrvc (approved):
I assume the third compare test (for site number) is there only so the result is deterministic; that is, if we need to figure out what happened, we know what order things were sorted in.  The third compare in terms of "which run to take" is not really needed, but I can understand a case for it.

clevengr:
I will review this as soon as I can (I'm still in transit).  I certainly have no problem with the general approach as described, especially as an "immediate fix" for the ACPC problem.  However,  I think we may also want to have a discussion about a possible longer-term solution as well.
For example, the current code has every client (judges, plus any other client such as a "queue display" fetching every single run every time a new run is needed, sorting those (maybe THOUSANDS) of runs, and selecting the best candidate from the sorted results.  That is LOT of overhead; we might want to consider alternatives. 

 For example, runs could be maintained on the Server in a standard Priority Queue based on some filter criteria; a request for the "next available run" would then be a simple removal from the head of the Priority Queue.   This is just one idea; maybe there are better ones.  But I think we're should talk about it.

But we should also implement a "quick fix" for the ACPC contest in two weeks.

ashoos72:
I think we should change the whole concept of AJ fetch.

We have two different issues that need to be addressed: #1 -- multi-site &
#2 everyone selecting the same run.

I'm addressing #2, I think the AJ should do a similar thing as it does now
-- look at its list, pick a run and send a request off to its server. The
"new" feature would be that in this request is included this judges AJ
criteria (ok fine, the server knows my AJ judging criteria we don't need to
send it -- but we need some flag turned ON). If the server gets the request
and can't fulfil it because someone else already grabbed it and the flag is
ON, the SERVER will now (using the AJ's criteria) go ahead and pick the
next run that meets the criteria and return that different run to the AJ
for judging. This will require changes on the AJ side to deal with getting
a different run. But I hope you get the basic concept.

As for #1 -- Single site is simple, run id order. For multisite, I think we
should use epoch time across all sites :) or something like that to select
the order for judgement across sites. So runs will be fetched in real world
order, no matter what time your contest was started.

Just throwing these thoughts out there...
--Sam

lane55:
All,

                One problem is solved, change sort of runs AJ will pick from sort by site, runid.  There
                is a PR to sort by elapsed ms.



re: server selecting runs for AJs

                The server has enough load already, let’s focus on changing the way the
                AJ/client selects its runs.

                Also there were 4 active servers, they would end up processing all the AJ
                requests (20+ of them) adding to the processing/memory instead of
                sending back a run not available package.

re: using epoch time

                The epoch time depends on the server’s time which can be incorrect,
                and if is incorrect would favor the runs on the site with the earliest/incorrect date/time.
                Whereas the elapsed time is relative time to the start of the contest.
                Since all sites will usually start at the same time and we base score
                on elapsed time, we should continue to rely upon that.

re: optimizing selection of runs

                The AJ fetches runs in a brute force method, rather than using a sub list of only
                the runs that the AJ can select.

                Current Code:   AutoJudgingMonitor.findNextAutoJudgeRun

                get all runs from model
                sort runs (by site 1, run id …. until PR pushed/fix)

     for all runs
          if ( run state is QUEUED_FOR_COMPUTER_JUDGEMENT )
          {
               if ( AJ assigned problem )
               {
                  return run
               }
          }

                This could easily be optimized to create a new list with only the runs
                that the AJ can select from, rather than say the thousands of runs in the model.

re: the bear of a problem

                Once we get past the selecting all Site 1’s runs first sort order problem, https://github.com/pc2ccs/pc2v9/issues/444,
                there is the all AJs requesting the same run, https://github.com/pc2ccs/pc2v9/issues/445.

                One idea is a load balance-ish module (Chief/Master judge) which would be responsible for
                assigning/re-assigning a AJ a run.   That would require a bunch of new code and testing to go
                along with it… so not a solution for now.

                Another thought is if the AJ has a list of 10+ runs that the AJ would not select the first run
                but select the second or third.   Alas, same problem of contention.

                Another thought is to pause the time between receiving run not available and selecting
                the next run, say 40ms.
                And to randomize it a little take the AJ# and make that a factor
                int sleepMS = 40 * ( contest.getClientId().getClientNumber() % 10 ); // so max would be 9*40 aka 540ms
                the 40 is arbitrary so it could be 10 or 60…


                Thoughts ?

Doug

johnbrvc:
I think we have to put the sort by _time_submitted_ thing (#444) ASAP so they can use it for their upcoming contest in a few days.

As a longer term solution (#445), I agree with Sam and JohnC in that I think it's up to the server to return the next available run when a client auto-judge) asks for one (either because it asked for one that was not available, or, it just asks for the next run).  The server already has a hashtable of runs (RunList), so it is already keeping track of runs isn't it?  Just add another list (or queue) that has QUEUED_FOR_COMPUTER_JUDGEMENTs.  We really don't care about NEW (manual judging), since there is little chance of run request collisions for manual judging.  The time and overhead saved by not having to service run requests for the same run would be worth it.  In fact, we observed during the ACPC, that the more auto-judges they created, it seemed the more contention we observed.  Basically, 20+ judges asking for the same run instead of 1.  As it stands now, it's better to have fewer AJ's servicing a few (or one) problems, it seems.  Just my $0.02.

src/edu/csus/ecs/pc2/core/list/RunComparatorByElapsedRunIdSite.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import edu.csus.ecs.pc2.core.model.Run;
+
+/**
+ * Run Comparator, Order the runs by elapsed time (ms), run id, then site.
+ *
+ * @version $Id$
```
clevengr, line 12: For CI/Cleanup, this line should be omitted.

clevengr (approved):
The comments I've made are minor bordering on trivial; the intent of the PR is totally adequately met.

PR #457, opened 2022-06-14
I 407 add output/custom validator load from yaml

### LINKED ISSUE #407: During yaml load problem custom output validator not loaded
**Describe the issue**: 

If an output validator is specified in problem.yaml, the validator
is not loaded/used.

**To Reproduce**: 

Create the output validator file/executable.

Define a (custom) output validator in problem.yaml

validator: 
    validatorProg: 'config/bells/output_validators/bells_validator/validator'
    validatorCmd: '{:validator} {:infile} {:ansfile} {:feedbackdir} '
    usingInternal: false
    validatorOption: 0

    use-internal-validator: false
    use-clics-validator: false
    use-custom-validator: true
    use-clics-custom-validator-interface: true

Use --load to load yaml/CDP.

**Expected behavior**: 

For the problem, on the OutputValidator tab,

Selected radio: Use Custom (User-supplies) Validator
Validator program: config\bells\output_validators\bells_validator\validator
Validator Command Line: {:validator} {:infile} {:ansfile} {:feedbackdir} 
Validator interface, selected radio: Use CLICS Standard Interface

In Problem report:
   Validator name : config\bells\output_validators\bells_validator\validator

**Actual behavior**: 

For the problem, on the OutputValidator tab,

Selected radio: Use CLICS Validator

In Problem report:
   Validator name : edu.csus.ecs.pc2.validator.clicsValidator.ClicsValidator

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

lane55:

Loads custom output validator specified in file/path in problem.yaml file,

Samples for testing have been added, some CI (javadoc) and improved error messages were updated/added.

See https://github.com/pc2ccs/pc2v9/issues/407#issuecomment-1155405410 for how to specify
validators (there are 3 ways to specify)

### Issue which the PR fixes

#407

Specify output validators, test using validators

Use examples load ccs1 sample which contains:
samps/contests/ccs1/config/castles/output_validators/yes_always.sh
samps/contests/ccs1/config/channel/output_validators/yes_always_2.sh

examples specified in:
samps/contests/ccs1/config/castles/problem.yaml
samps/contests/ccs1/config/channel/problem.yaml


johnbrvc (commented):
As a point of information, the contest configuration provided by the DOMJudge folks for the Dhaka WF pretest contest does not contain the `validatorProg` key in the `validator` group. Problems that use an output validator simply say:  `validator: custom` and that is all.  Unless the DOMJudge team changes their **problem.yaml** files to comply with the format suggested above, there is no point in merging this PR prior to WF Dhaka.  Ex. [https://github.com/icpcsysops/ccsconfig-dhaka/blob/master/contests/pretest/config/ixthproblem/problem.yaml](url)

I guess it is assumed that the if `validator` is custom, the reader should know to look in the `output_validators/validate` folder.  8 of the 12 problems for the pretest contest have output validators defined; the validators are all C++ programs.  Sometimes they are called "validator.cpp" other times "output_validator.cpp".  I'm not sure how a reader is supposed to know exactly what the validator program is.

lane55:
The validatorProg is a solution to the problem of custom validators, during contests we update the problem.yaml
with that key in order to specify the name of the custom validator, so we can load it into pc2.

I assume that most CCSs find all source files for validators and then compile and use the validator applications.
I have also assumed that there can be more than one validator per problem.   I have seen where there 
were two validators in a problem, each validator source in a different directory.

At this time pc2 does not support multiple validators per problem.

It is our practice to do that manually for each output validator
* identify the validator source
* create the validator application
* add the validatorProg key/value
* push the validator into git repo

lane55:
> there is no point in merging this PR prior to WF Dhaka.

I disagree the reason has not changed from a recent meeting.   This change supports pc2 automatically
loading a validator, which is highly desired during the WF.   If not then we must still create the validator
application and manually update the problem picking the validator.

johnbrvc:
> > there is no point in merging this PR prior to WF Dhaka.
> 
> I disagree the reason has not changed from a recent meeting. This change supports pc2 automatically loading a validator, which is highly desired during the WF. If not then we must still create the validator application and manually update the problem picking the validator using the Admin. Without this it increases the human error.

Agreed.

johnbrvc (approved):

clevengr:
> > there is no point in merging this PR prior to WF Dhaka.
> 
> I disagree the reason has not changed from a recent meeting. This change supports pc2 automatically loading a validator, which is highly desired during the WF. If not then we must still create the validator application and manually update the problem picking the validator using the Admin. Without this it increases the human error.

So that I'm clear:   we are still going to have to manually (externally) compile any custom validator source to generate the custom validator executable... right?    That is, this PR is not about attempting to automatically generate a custom validator executable from the specified source (right?)

johnbrvc:
> > > there is no point in merging this PR prior to WF Dhaka.
> > 
> > 
> > I disagree the reason has not changed from a recent meeting. This change supports pc2 automatically loading a validator, which is highly desired during the WF. If not then we must still create the validator application and manually update the problem picking the validator using the Admin. Without this it increases the human error.
> 
> So that I'm clear: we are still going to have to manually (externally) compile any custom validator source to generate the custom validator executable... right? That is, this PR is not about attempting to automatically generate a custom validator executable from the specified source (right?)

Yes.  Compiling the validator has to be done by hand, since, the compilation process varies depending on what language the validator is written in: shell script, python, c, c++, java, kotlin(?) . I write a shell script (https://github.com/icpcsysops/ccsconfig-dhaka/blob/pc2-config/tools/djprob2pc2) that takes the original DOMJudge supplied problem.yaml and massages it so PC2 can digest it.  It also adds the output validator keys and values as described in the issue #407 .  It also builds the validator and puts the executable in the "shell-script" defined location.  Currently, it only knows about C++ validators, but a switch/case can make it build for java or C if necessary.  Maybe I'll add those to the script before we run into a case where we need them.  I also have to change the default output folder for output validator executables since currently I have it as an absolute path with is lovely on my machine, but not Doug's (for example).

lane55:
> So that I'm clear: we are still going to have to manually (externally) compile any custom validator source to generate the custom validator executable... right? That is, this PR is not about attempting to automatically generate a custom validator executable from the specified source (right?)

Yes, this improves the code that searches for the output validators, esp to support relative paths.
Before this PR we have to use absolute path to find output validators, which is messy.

Issue #458 is the issue/enh to compile the output validators.  That is out of scope for this PR/Issue.

PR #485, opened 2022-07-12
I 484 improvement with run processing and saving run list

### LINKED ISSUE #484: Improvements/optimizations for run list saving

Yes, potential inefficiencies in adding/storing run list

**Feature Description**: 

Improve write runlist to disk by removing unneeded synchronize
and by writing runlist on a seperate (single) thread.

Sorts

Add single thread used to write run list.

**Additional context**:

lane55:

Now runlist is run on a thread

Two unit tests are failing because not all runlist files are written.  Sometimes the unit tests pass, sometimes they don't.

for testBackupStressUsingThreads
org.junit.ComparisonFailure: Expecting dir entries  in C:\repos\lane55fork\.\testout\RunListTest\runlistbackupStress expected:<[101]> but was:<[5]>

for testBackup
org.junit.ComparisonFailure: Expecting runlist files count  in C:\repos\lane55fork\.\testout\RunListTest\runlistbackup expected:<[8]> but was:<[2]>

### Issue which the PR fixes

#484 

### DIFF
--- a/src/edu/csus/ecs/pc2/core/list/RunList.java
+++ b/src/edu/csus/ecs/pc2/core/list/RunList.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2022 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.list;
 
 import java.io.File;
@@ -297,19 +297,35 @@ public class RunList implements Serializable {
      * @throws IOException
      * 
      */
-    private synchronized boolean writeToDisk() throws IOException, ClassNotFoundException, FileSecurityException {
+    private boolean writeToDisk() throws IOException, ClassNotFoundException, FileSecurityException {
+
         if (!isSaveToDisk()) {
             return false;
         }
         
         boolean stored;
         String fileName = getFileName();
-        String backupFilename = getBackupFilename();
         
+        String backupFilename;
+        
+        Hashtable<String, Run>copyofRunHash;
         synchronized (runHashLock) {
-            stored = storage.store(fileName, runHash);
-            storage.store(backupFilename, runHash);
+            backupFilename = getBackupFilename();
+            
+            try {
+                File currentRunList = new File (fileName);
+                File backupRunList= new File (backupFilename);
+
+                currentRunList.renameTo(backupRunList);
+            } catch (Exception e) {
+                StaticLog.getLog().log(Log.WARNING, "FAILED to rename current " + fileName + " to " + backupFilename, e);
+            }
+            
+            copyofRunHash = new Hashtable<String, Run>(runHash);
+            
+            stored = storage.store(fileName, copyofRunHash);
         }
+        stored =  true;
         
         backupList.add(backupFilename);
         while(backupList.size() > 100) {
@@ -319,7 +335,7 @@ public class RunList implements Serializable {
                 file.delete();
             }
         }
-
+       
         return stored;
     }
 
@@ -400,8 +416,7 @@ public class RunList implements Serializable {
     }
     
     private void logException(String string, Exception e) {
-        //TODO:  huh?  the following two lines say "if X is null then call a method on X".  Doesn't make sense. Maybe should be "if != null"?  jlc
-        if (StaticLog.getLog() == null) {
+        if (StaticLog.getLog() != null) {
             StaticLog.getLog().log(Log.WARNING, string, e);
         } else {
             System.err.println(string + " " + e.getMessage());
--- a/src/edu/csus/ecs/pc2/core/list/RunList.java
+++ b/src/edu/csus/ecs/pc2/core/list/RunList.java
@@ -35,7 +35,7 @@ public class RunList implements Serializable {
      * List of runs.
      */
     private Hashtable<String, Run> runHash = new Hashtable<String, Run>(200);
-    
+
     /**
      * Lock for accessing list of runs.
      */
@@ -52,9 +52,14 @@ public class RunList implements Serializable {
 
     private LinkedList<String> backupList = new LinkedList<String>();
     
+    private RunListWriteThread runListWriteThread = null;
+    
+    private Object runListWriteLock = new Object();
+
     public RunList() {
         saveToDisk = false;
         nextRunNumber = getINIBaseRunNumber();
+        createAndStartRunListWriteThread();
     }
 
     private int getINIBaseRunNumber() {
@@ -79,7 +84,7 @@ public class RunList implements Serializable {
         } else {
             retVal = 1;
         }
-        
+
         return retVal;
     }
 
@@ -110,7 +115,7 @@ public class RunList implements Serializable {
         add(run);
         return run;
     }
-    
+
     /**
      * Add a run into the run list, no changes.
      * @param run
@@ -126,7 +131,7 @@ public class RunList implements Serializable {
             writeToDisk();
         }
     }
-    
+
 
     /**
      * Get a run from the list.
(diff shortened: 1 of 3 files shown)


clevengr (approved):

clevengr:
Regarding writing the RunList on every notification:  the potential problem I see is that if we SKIP some notifications then we don't know for sure whether we've got a valid RunList on disk.  I agree that it would be better if we could find a way to avoid writing it every time -- but what criteria can we use to do that?  I'm not sure that "every X times" or "every Y seconds" will ensure a valid system state.  (Maybe there's a way to consider what's _causing_ the update; some packets might be triggering an update when in fact there's not really a significant change? This however would be much more complicated to implement...)

Regarding `diskWriteLock`:  we noticed that in the  original version, method `backupAndWriteRunList()` had the call to `storage.store()` inside the `synchronized (runHashLock)`.  That `synchronized` block is necessary to ensure that the RunList (hashtable) doesn't get modified while `copyOfRunHash` is being created.  However, once the copy is made, there's no need to keep the hashtable locked -- the call to `storage.store()` is writing the _copy_ and we no longer care if the original gets modified by some (other) thread.  In fact, we _don't want_ the hashtable to remain locked the entire time we're actually doing `storeage.store()`; there are potentially lots of other threads waiting for the **runHashLock**.  So, we moved the end of the `synchronized (runHashLock)` block up above the call to `storage.store()`, to allow those other threads to continue accessing the hashtable.

However, this introduced a new potential problem:  two threads could both call `storage.store()` at the same time (since it was no longer protected within the `synchronized (runHashLock)` block.  So we introduced a _new_ lock:  `diskWriteLock`.  This lock keeps threads from simultaneously performing the actual disk write, while at the same time allowing `runHashLock` to be released while the disk write it occurring.

Please let us know if you think the above logic is flawed, or if there's a better way to handle it.

johnbrvc:
Regarding diskWriteLock.   Since backupAndWriteRunList is only called from the new thread (of which there is a single instance), I'm not sure I see the need for diskWriteLock at all, unless, as I said, some other routine is calling backupAndWriteRunList, but I don't see anyone else calling it (it's a private method).    Since it is synchronizing on a private member Object, even if there are multiple RunList objects, they'll each have their own lock.

clevengr:
By moving the call to `storage.store()` outside of the `synchronized(runHashLock)`, we have allowed the possibility that some other thread which was currently blocked on `runHashLock` can now continue (which is what we want).  However, this hypothetical other thread now has the possibility of also calling `backupAndWriteRunList()`, getting the (free) `runHashLock`, making it's own copy of the RunList, and then calling `storage.store()` -- while the original thread is still inside `storage.store()`.  Thus we do have the possibility that two (or more) threads could be inside `storeage.store()` at the same time. We added the `diskWriteLock` t protect against that.

Do you disagree that the above scenario is possible if we don't have `diskWriteLock`?

johnbrvc:
I do disagree.  There is only a single thread that calls the backupAndWrite routine... unless I'm missing ssomething.

lane55:
From johnbrvc> In the new thread, do we want to write the runlist every single time we're notified? 
Note sure what notified means, if it means every time a run is added or changed (update judgement/state=JUDGED).
Each event that would change the scoreboard/scoring the run list is written.   But there are many states NEW, 
BEING_JUDGED, etc. where the run list is not written because on recovery/restart most of those states will
be set to the essentially to-be-judged status.

lane55:
re: using runHashLock and diskWriteLock  some rationale.
Before this change any update to the runHash (add/udpate/etc.) would block the during I/O;store.
The idea was to synchronize as tightly as possible, first was to remove the synchronized from 
the method sig (done).
Then we looked at the two tasks in the now named backupAndWriteRunList,
make a copy/snapshot of the runHash and write the runHash to disk.
The time to write the runHash to disk is the most expensive and all other
runHash changes were blocked while waiting for runHash to be written
to disk.   That is why we split the old block synchronized on runHashLock
into two synchronized task - clone the runHash and write the runHash.

lane55:
> I do disagree. There is only a single thread that calls the backupAndWrite routine... unless I'm missing something.
I think you have limited your thought to the backupAndWriteRunList synchronizing, the change we made was about not blocking
the other 10+ methods that access or update the runHash, esp. while doing I/O writing the runHash.

johnbrvc:
I'm still missing something.  The only place in the entire codebase that uses diskWriteLock is line 440 of RunList.java.  The routine this is in, backupAndWriteRunList, is only called in one place, from the single thread that does all the writing.  
The 10+ other methods that access runHash all synchronize on runHashLock, NOT diskWriteLock.  My problem is with diskWriteLock.  I'm just wondering what it's there for.  I fully understand runHashLock.

clevengr:
I now agree with JohnB; I think the `diskWriteLock` synchronization is unnecessary, because there is only one thread (the new thread which calls `backupAndWriteRunList()`) which ever actually writes the runfile (hashtable) to disk -- so there's no danger of two threads colliding on that write.

However, this discussion has led me to believe there is another, much more significant, possible design flaw:  what happens if, when the backupToDisk thread gets a `notify()` and hence starts working (calling `storage.store()`), another thread calls `notify()`.  That is, what happens to a Java monitor `notify()` call when there are no threads currently blocking/waiting on the monitor?  I'll research this -- but I have a bad feeling that it means that there is a possibility that we could lose one or more `notify()` calls.   (If so, this would explain the thing I saw in testing the failing JUnit, where it wanted 8 files but only ever got 6...)

johnbrvc:
I don't think that's a problem since THAT part of the RunListWriteThread is synchronized with runListWriteLock.  So, if the thread wakes up to process the notify (wait() returns), it owns the monitor for the lock and will prevent another notify from sneaking in since the notify() is also protected by the same runListWriteLock in writeToDisk() and should not be able to notify until the RunListWriteThread is back to waiting.  At least that is my understanding of how it works.  

Of course this begs the question, doesn't this put us back in the same boat as before?  Now writeToDisk is waiting on runListWriteLock while the RunListWriteThread does its thing from the previous caller?  Aren't we just kicking the can down the road one block?

lane55:
The introduction of diskWriteLock  was to ensure that the writing of the run list is synchronized.   Before
diskWriteLock  was created both updates of the run list (runHash) and writing the run list to disk was
synchronized using runHashLock.

This meant that every of the other 10+ methods that updated the run list had to wait for the storing
of the run list on disk.

So runHashLock was changed to be limited to the copying of the run list.  And diskWriteLock  was added
to synchronize the storing of the run list.

lane55:
re: synchronizing the write to disk using diskWriteLock

One reason the original writeToDisk method was synchronized was to ensure that after update of the run list
was synchronized.   

The RunListWriteThread was created to avoid the bottleneck of storing the run list to disk on the same
thread as writeToDisk.

Then for a short period of time the synchronization was achieved by putting both
the creating of the copy of the run list and the storing of it to disk using the runHashLock.
Then diskWriteLock was added to ensure that the run list store was synchronized.

lane55:
re: JUnit method not writing all expected runlist to disk

This is most likely due to the way JUnits methods are run, or more specifically halted/exited.

JUnits do not wait (no pun) for all thread associated with the method to stop before halting
the method.    So if a test takes "too long" then the threads created in the test (method)
will be unceremoniously halted.

As an example I have tested a JFrame in a JUnit method and the JFrame is almost immediately stopped
whereas if we run the same JFrame code in a main method, the JFrame continues to run. 

An ugly change and poor way to avoid this is to effectively add a sleep to allow all the threads to finish their work.

When the tests for storing the run list were written the method was synchronized and run on one thread.
Now the storing of the run list is run an a different thread.   And it is that second thread that may not
have time to complete the writing of the runlist when the JUnit runner effectively stops that new thread
that writes the run list.

lane55:
As a piece of info,  the specific assert that has been temporarily commented out, the assert that counts the number of run list files
on disk is: assertExpectedFileCount in RunListTest.    In an way we didn't have time to dive into why the assert failed
and had integration tested the new writeToDisk enough to move forward with that more efficient code.

PR #487, opened 2022-07-12
remove reference to CSUS

### LINKED ISSUE #486: builds should not reference external sites anymore.

troy2914:
fixes #486

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,15 +55,6 @@ job-build:
 #     - echo "Running test stage..."
 #     - echo "Do a test here!"
 
-push to ECS at CSUS:
-  stage: deploy
-  script:
-    - echo "Running deploy stage"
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -84,7 +84,7 @@
         value="${date.english} ${time.english}" />
 
     <!-- version is used in the jar manifest & in package.xml -->
-    <property name="version" value="9.8build"/>
+    <property name="version" value="9.9build"/>
```
clevengr, line 87: The original request was to fix (remove) the steps which were causing the build to fail because of the push to ECS at CSUS.  Changing the build number wasn't part of that, and we haven't made a decision on what should be "9.9" and what should be part of an updated 9.8 release.

clevengr (commented):
Please remove the change of the build number until we decide how we're going to handle what goes into 9.8 vs. 9.9.

Issue #497, opened 2022-07-17
Can not fetchRun from remote site in multisite contest


johnbrvc:
Thank you for working to help us improve PC^2!  
Please fill in the following fields in order to help us understand what issue you are reporting.

**Describe the issue**: 
Attempting to fetch a run from a remote site does not work.  (Note that this is different from "checking out" a run
from a remote site which DOES work).  Fetching a run is currently only used from the RunsTablePane/RunsPane to
"View Source" of a non-resident run.   NOTE: This is not a problem on a single site contest since it never has to remotely
fetch a run.

**To Reproduce**: 
1) Configure a multisite contest (I used sumit) with two sites.  I ran them both from the same PC.  site1 used port 50002
     and site2 used port 50003.
2) Bring up both sites
3) Configure at least 1 team account on site2 (I generated 5 teams starting at team # 2000)
4) Open a team model to site2 and log in using one of the newly created accounts on site2
5) Submit a run
6) Start an administrator on site1
7) observe the Runs tab and confirm the run submitted from step 5 is there and from site2
8) Select "View Source"
9) No source appears and after 30 seconds the message is updated indicating it was unable to fetch the run

**Expected behavior**: 
After step 8 above, a new window should be created showing the source code submitted for the run. 

**Actual behavior**:
After View Source is selected shows a message:
`   Preparing to display source code for run 1 at site 5`

After a number of seconds displays:
`   Unable to fetch selected run; check logs`

**Environment**: 
Version         : 9.8build 20220717
Build Date      : Sunday, July 17 2022 14:04 UTC
Build Number    : 6083

OS              : Windows 11 10.0 (amd64)
Java Version    : 1.8.0_321

But it will happen on all OS's

**Log Info**: 
The Administrator log on site1 that attempted to View the Source shows this:
```
220717 084945.580|INFO|ViewSourceThread-0|showSourceForSelectedRun|Preparing to display source code for run 1 at site 2
220717 084945.582|INFO|ViewSourceThread-0|sendToLocalServer|Sending packet to server Packet[FETCH_RUN] #2 ADMINISTRATOR1 @ site 1 -> SERVER0 @ site 1
220717 084945.583|INFO|ViewSourceThread-0|sendToLocalServer|Sent    packet to server Packet[FETCH_RUN] #2 ADMINISTRATOR1 @ site 1 -> SERVER0 @ site 1
[...]

PR #503, opened 2022-07-19
i465: Make sure that fetchRun response is for correct run, fixes #465 and fixes #479

### LINKED ISSUE #465: Incorrect Run ID and Site ID may be displayed for View Source from Runs tab

**Describe the issue**: 
This bug was reported by Coach from the ACPC.  View Source is pressed from a Runs tab.  Sometimes, the displayed Run ID and/or SIte ID may be incorrect in  the title bar of the displayed source code.  The contents of the window (the source code) may, however, be correct for the selected run.

**To Reproduce**: 
1) Select a previously unselected Run (for which the Runfiles have not been obtained). 
2) Select "Rejudge"
3)  Press "View Source" (on the "Rejudge" window)
4) Close both the "View Source" and "Rejudge" windows (basically, this fetches the runfiles locally)
5) DO NOT select a different run!!  From another team account, submit a run.
6) Do "View Source" from the Admin of the originally selected run.
7) Note that the Title bar displays the wrong run id (and possibly site id if multi-site)

**Expected behavior**:
The correct Run ID (and site) should be shown in the title bar

**Actual behavior**: 
The wrong Run ID (and site) are shown in the title bar.

**Environment**: 
All.

**Log Info**:  
 

**Additional context**: 
In RunsTablePane and RunsPane (both have the same issue), the variable "fetchedRun" is used to get the Run ID and Site ID to display on the viewed source's title bar.  The problem is, fetchedRun can be updated asynchronously in the "runChanged" handler when a run event is received for a different run.  There is a comment in the code:
 //TODO:  this needs to be reconsidered; why are we continuing when the event wasn't sent to us?  (Why wasn't it sent to us?)
That, is a good question.  In any event, when fetchedRun is updated in this case, the original run's ID and site are written over.

Possible solutions:
1) Don't set fetchedRun/fetchedRunFiles in the case where the received event was not one requested by the client
2) Don't use fetchedRun for the MultipleFileViewer title, rather, use 'run'.

### LINKED ISSUE #479: View source not viewing proper source

View source shows source from previous run, should show from the currently
run being judged.

Reported as:
`If I mark a solution and view source it opens another solution with a different id, the only way we see the correct solution is by clicking rejudge and then view source from inside to see the actual solutions.
`

**Expected behavior**: 

View the source for the run.

**Actual behavior**: 

Views source from an earlier run.

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

johnbrvc:

It is possible (and likely in a busy contest) to get the incorrect run back when doing a View Source.  This is because no check was done to make sure that a runChanged event was for the specific run whose source was being requested.  So if a runChanged
event for a different run happens to come in while the View Source was occurring, it is possible to get the wrong run's information - only the title bar will be wrong in this case.  Note that this fixes this bug on a single site contest.  On a multisite contest, the fix for bug #497 is also required or no response will be received from the remote site, and it is possible to get the source for the wrong run in this case.

### Issue which the PR fixes
#465 and #479 
Windows 11
java version "1.8.0_321"
 
1) On Judge, select a previously unselected Run (for which the Runfiles have not been obtained).
2) On a judge, select "Rejudge"
- Note: THE ABOVE STEP DOES NOT WORK.  You cannot select "Rejudge" on a run which has never been judged.  There must be some required precondition which is not specified.  (I loaded the "sumithello" contest, then submitted the "hello.java" source from the samps directory, then performed steps (1) and (2) above; you can't select "Rejudge" on the submitted run.  (jlc 8/21/22)
3) Press "View Source" (on the "Rejudge" window)
4) Close both the "View Source" and "Rejudge" windows (basically, this fetches the runfiles locally)
5) DO NOT select a different run!! From another team account, submit a run.
6) Do "View Source" from the Admin of the originally selected run.
7) Note that the Title bar would display the wrong run id (and possibly site id if multi-site), now it is fixed.
This may require a few times to get it to happen.  It is hard to reproduce on a quiet system.

Updated: Aug 21, 2022:
After testing this for the past couple of weeks, there is a much easier and more reliable way to duplicate the issue.
1) Start sumit server.
2) Use 2 existing runs (any 2 will work), or, if using a clean contest with no runs yet, create 2 new runs using a team account, let's say Run 1 and Run 2.
3) Start a non-AJ judge
4) Select "All Runs" tab
5) Select Run 1
6) View Source (Site 1 Run 1's source will be displayed)
7) Close source window
8) Select Run 2
9) View Source (Size 1 Run 2's source will be displayed)
10) Close source window
11) Select Run 1
12) View Source (Site 1 Run 1's source is displayed, but title bar says: Site 1 Run 2)

Repeat this process using the updated (fixed) code and the correct Title bar is displayed.


lane55:
So the bug is that the wrong run's source code is being displayed.
The code was written expecting that View Source would be rare.   If a judge or admin
does a view source on a number of different runs (esp from remote servers)  then
it is possible for the code to randomly pick a winner and show that source.

So if the requirement is now that each View Source is guaranteed to view the run's souce,
other changes will need to be made.

lane55:
Note in this PR this was added in a catch block
```
  //make sure this is clear in case of exception
            requestedRun = null;

```
IF more than one view source can be pending, that cleanup would remove
that run from the queued runs.

johnbrvc:
> 

Correct.  I added the "requestedRun" member to remember the last run whose source was requested.

johnbrvc:
> The title bar should never be incorrect, the title bar should be based on the fetched run (I assume the title bar is based on the requested Run, which likely would cause an incorrect title bar run info).

The title bar used to be based on the fetchedRun, which could have been wrong, but the source could have been correct.  That is, if the 'run' already had its runfiles loaded (getContest().getRunFiles(run) returns non-null), AND fetchedRun was filled in with some other run's runChanged event information in the mean time, then the title bar would show fetchedRun's information (which would be wrong), but the source would have been from the getContest().getRunFIles(run), which would have been correct.

I changed the title bar to always match the place it got runFiles from ('run') so they will never be out-of-sync.  In fact, now fetchedRun should always match requestedRun anyway, but, I think it makes more sense to use the same information that was used to get the run files.

johnbrvc:
After a discussion with Doug, I think we came to the conclusion that multiple outstanding "View Source" requests are not (and have never been) supported.  Supporting multiple outstanding "View Source" requests it outside the scope of this issue and PR.  That being said, multiple "View Source" windows **can** be open at the same time for different runs.  However, having multiple "outstanding" (unfulfilled) "View Source" requests active (that is, select (**remote site**) run, view source, select another run
before response comes back and "View Source", etc.) may not work as expected.  Only the most recently selected **remote** run will be displayed, in addition to any local runs selected.

Example:  If you get tired waiting for a **remote** run's source to be displayed (from a remote site), and move to another run and "View Source" of that run:
1) if the 2nd run selected is a remote run, the first one is effectively forgotten about and the new one is the "active" one.  The response window created will be for the 2nd run.  This is because the "requestedRun" variable is overwritten with the latest remote run request.
2) if it's a local run, it will be displayed correctly, since it doesn't use "requestedRun" or "fetchedRun", rather it just uses "getRunFiles()" and, if the previous remote run eventually comes in withing the time limit (30 seconds), that too should be displayed (in a separate window).

The important thing is, no incorrect source will be displayed.

Before this fix, it was a crap-shoot about what you got back, and you could get the wrong title displayed as well. 

So, I don't think any code changes to address the original issue are required, unless someone sees a bug...  To focus your review efforts, the routines in question in RunsTablePane.java are showSourceForSelectedRun() and the runChanged() implementation.

clevengr:
So do I understand correctly that (1) "multiple outstanding View Source requests are not supported", but (2) it is possible to make multiple View Source requests?   If this is the case, it seems very "user-unfriendly".  Would it work to simply add a variable "viewSourceRequestInProgress", set that variable when the View Source button was hit, clear that variable when the requested source was received, and check that variable each time a new View Source request was made (i.e., when the View Source button was pushed) and if a View Source operation was already in progress then pop up a dialog like "Another View Source operation is already in progress; please cancel that request before making a new View Source request"?

johnbrvc:
> So do I understand correctly that (1) "multiple outstanding View Source requests are not supported", but (2) it is possible to make multiple View Source requests? If this is the case, it seems very "user-unfriendly". Would it work to simply add a variable "viewSourceRequestInProgress", set that variable when the View Source button was hit, clear that variable when the requested source was received, and check that variable each time a new View Source request was made (i.e., when the View Source button was pushed) and if a View Source operation was already in progress then pop up a dialog like "Another View Source operation is already in progress; please cancel that request before making a new View Source request"?

Yes, your understanding is correct.  I can certainly add a sentinel to guard against multiple "View Source" starts.   I guess it would kind of fit in with being CI, since it's not really part of the scope of this issue, but it's a quick fix that will make the interface more consistent for users.   Implementing a "cancel" operation is a  little more involved because the "View Source" is spawned off in another thread and forgotten about by the main class.  It times out after 30 seconds if no response is received.   A variable can be added that is checked in the wait loop of the thread to see if the user wishes to cancel the operation.   A dialog could be presented that gives the user the option to cancel the outstanding request or wait for it - this way the user knows what is going on.   ("Do you want to wait for the current view source to complete or cancel it?  [Wait] [Cancel]). 

One thing to be aware of, is that you will still be able to have multiple "View Source" windows open at the same time (with
source code displayed in them).  This is useful for judges sometimes to compare two run's source code (what changed between run 100 and run 110,  for example).  The above CI would only addresses the case where there is an outstanding request to a remote site for a run.  (Local runs are populated immediately, since the runfiles are, well, right there already).

Does that  make sense?  Any other suggestions?

clevengr:
I like the idea of adding a sentinel against requesting multiple View Source operations.  It could be viewed as "CI", but it could also be viewed as "within the scope of this bug/PR" because of the fact that the "bug" is that users are getting "wrong information" when View Source is invoked in a certain way; this would stop them from being able to do that.

I also like the idea of the more comprehensive approach of supporting ability to "Cancel" a previous request -- but I do think that one is probably better put off into a separate "Issue".

johnbrvc:
The sentinel will not prevent the original reported bug from happening.  It has to do with the fact another undirected runChanged event came in and caused a variable (fetchedRun) to be set with the wrong run.  This problem can happen on a non-multi-site contest.  The "multiple outstanding" view sources could only happen on a multi-site contest (for all intents and purposes).
Basically, on a single site contest, there are no "outstanding" requests, so the View Source is fulfilled immediately; well, almost immediately - however long it takes to spawn a thread and display a window.

lane55:
I think that the user would expect that if they chose toe View Source that the source should be etched and viewed
and that there would not be an additional requirement of  only one View Source at a time using a the sentinel.
If we restrict with a sentinel, then while waiting for that fetched run one would never receive another fetched
run (having to logoff to clear the sentinel)

The view run should be a async operation.  Send a request for a run, when the corresponding run
is received view it.

lane55:
re: one judge/admin receives another judge's view source run.
While there seems to be code to handle this condition - that condition is a bug.  The fetched run should only
be sent to the client/admin/judge that requested it.   Unlike new run, checked out run, judged run - the view
source run should only be sent and received by the client that requested it.

lane55:
A view source in progress (and cancel) is an enhancement.    I recommend that a new issue is created
and given a lower priority.   Once we agree on the View Source requirement and implement it adding monitoring
and cancel imho is a nice to have.

johnbrvc:
> I think that the user would expect that if they chose toe View Source that the source should be fetched and viewed and that there would not be an additional requirement of only one View Source at a time using via a sentinel. If we restrict with a sentinel, then while waiting for that fetched run one would never receive another fetched run (having to logoff to clear the sentinel). That also means we'd need a way to monitor/cancel that View Source.
> 
> The view run should be a async operation. Send a request for a run, when the corresponding run is received view it.

If a View Source run does not return information in 30 seconds, it will allow you to select another run.  30 seconds has always been the timeout for how long it waits for a View Source.  With the sentinel value, any error (timeout, bad response, missing run files, exceptions, etc, will cause the sentinel to be cleared so you can try a different run. (No logout necessary).  In addition, a message is put in the message text label (like all other errors for fetching a run) that indicates if a view source operation is pending, and to please wait until that finishes.

Making it so the multiple OUTSTANDING requests (and this is for multisite only) work, is not in scope of Coach's original bug (#465).  So, we can make up another issue for that, but is it really necessary, as long as the user is told why he can't request another run at the moment?

johnbrvc:
> re: one judge/admin receives another judge's view source run. While there seems to be code to handle this condition - that condition is a bug. The fetched run should only be sent to the client/admin/judge that requested it. Unlike new run, checked out run, judged run - the view source run should only be sent and received by the client that requested it.

Ideally, yes.  But that would be a change to the way the  updateRun handler works, and changing something that affects so many things doesn't seem to be necessary for this issue (correct me if I'm wrong).  I added a comment in the code that explains this (RunTablePane):

 `                                    // The FETCHED_REQUESTED_RUN reply is sent with a a SentToClientID of null as found in
                                    // InternalContest.updateRun().  the RunEvent's sentToClientId member is only set when the run
                                    // is being checked out for BEING_JUDGED, BEING_RE_JUDGED, CHECKED_OUT, HOLD, otherwise it is
                                    // is not set (null), and winds up here in the case of "View Source" but not from a being-judged dialog.
`

johnbrvc:
> A view source in progress (and cancel) is an enhancement. I recommend that a new issue is created and given a lower priority. Once we agree on the View Source requirement and implement it adding monitoring and cancel imho is a nice to have.

Agreed.  Low priority as you said.  So, let's agree on how Coach's bug #465 and the almost duplicate #479 should be handled for this PR.  Displaying the wrong Title in the title bar but the correct source, or the wrong title and wrong source are caused by the same problem - namely, the way responses to the request run was processed in runChanged, and the strings used for the title bar.  So, fixing #465 also fixes #479.   So for this PR, I propose:

1. Do not allow multiple outstanding View Source requests and show an error message to the user if they attempt to do so
2. Change the strings used for the title bar of the View Source to be the requested run.  Previously, it used the fetched run which may been incorrect if the run files were already obtained.
3. Only accept run files for the requested run, ignoring all others.
4. Create a low priority issue to allow the user to cancel an outstanding run request.
5. Edit the original issue report for #465 to clearly indicate the source of the bug report (Coach and the ACPC). 

This PR addressed all but the last two of those items, which can easily be remedied.

Comments or suggestions?

johnbrvc:
As per the meeting between @lane55, @clevengr and @johnbrvc on 08/16/2022 6:00PM PDT, after some minor changes in error reporting (and a CI), this PR should be ready for re-review and, if approved, merged into develop.

An easy way to simulate a delay in fetching a run for the purposes of simulating multiple outstanding requests (on a single-site system) is to:
1) put a breakpoint in PacketHandler.requestFetchedRun() (the first line of the routine will do).  or a Thread.sleep(20000).  Basically, force the fetch to take a long time...
2) start the server
3) start an admin (or judge). 
4) select a previously submitted run and choose "View Source".
5) The server will break in requestFetchedRun(), and the admin/judge client requesting the source will indicate it is "Preparing to display source for..."
6) on the admin/judge client, select a run (can be the same one) and do View Source.
7) a message indicating that a request is already pending for the previous run will be displayed in the message label area
8) continue past the breakpoint on the server.  if you do this within 30 seconds, the source will be displayed
9) you can now do another View Source on a different run that requires a server fetch

src/edu/csus/ecs/pc2/ui/RunsTablePane.java
```diff
@@ -2146,7 +2224,12 @@ private void showSourceForSelectedRun() {
                         SerializedFile[] otherFiles = runFiles.getOtherFiles();
 
                         // create a MultiFileViewer in which to display the runFiles
-                        MultipleFileViewer mfv = new MultipleFileViewer(log, "Source files for Site " + fetchedRun.getSiteNumber() + " Run " + fetchedRun.getNumber());
+                        // Note: previously used 'fetchedRun' here for site/number; it is possible that those values are not
+                        // correct if the run files are already present; in that case, we would have never asked for them to be
+                        // retrieved, and would have used whatever was in fetchedRun.  An NPE was also possible if fetchedRun was null;
+                        // that is, the runfiles were already present from an edit run or judge run on the current client, not no
+                        // server request was every made for a run's files.
```
clevengr, line 2231: I can't parse the final clause of the above sentence.  Was it intended to say "`so` no server request was `ever` made"?

johnbrvc: Yes.  Precisely what you said.  Basically, View Source never made a request to the server for runFiles, then fetchedRun/fetchedRunFiles would be null.  I will change that when the review process is complete.

src/edu/csus/ecs/pc2/ui/RunsTablePane.java
```diff
@@ -2106,31 +2179,36 @@ private void showSourceForSelectedRun() {
...
                         getController().getLog().log(Log.WARNING, "No response from server to fetch run request after " + waitedMS + "ms");
-                        getController().getLog().log(Log.WARNING, "Unable to fetch run " + run.getNumber() + " from server");
-                        showMessage("Unable to fetch selected run; check log");
-                        return;
+                        bFetchError = true;
                     }
                 }
                 
+                // OK to now start another view source
+                AllowViewSource();
```
clevengr, line 2198: I think there may be a potential synchronization error here (although I could probably be convinced otherwise).  See my comments on the subsequent call to `AllViewSource()`, below, around (new) line 2288 for details.

src/edu/csus/ecs/pc2/ui/RunsTablePane.java
```diff
@@ -2196,8 +2279,13 @@ private void showSourceForSelectedRun() {
...
             getController().getLog().log(Log.WARNING, "Exception logged ", e);
             showMessage("Unable to show run source, check log");
+            
+            //make sure this is clear in case of exception
+            requestedRun = null;
         }
-
+        
+        // OK to now start another view source now
+        AllowViewSource();
```
clevengr, line 2288: This is a followup comment to the one I made above, around (new) line 2198.  Suppose a user clicks the `View Source` button.  That triggers a call to `BlockViewSource()`, followed by a spawning of a Thread (call it Thread1) whose `run()` method does just one thing:  calls method `showSourceForSelectedRun()`.  At some point during the execution of `showSourceForSelectedRun()` on Thread1, the method calls `AllowViewSource()` (line 2198, above).  

Now suppose a user clicks the `View Source` button again, _after_ the execution of the first  `AllowViewSource()` call in Thread1 but _before_ Thread1 completes the execution of `showSourceForSelectedRun()`.  This second button click will call `BlockViewSource()`, and then will spawn a new Thread (call it Thread2) which runs a separate copy of method `showSourceForSelectedRun()`.   So Thread2 starts executing in `showSourceForSelectedRun()`.

Now suppose Thread1 finishes -- the last statement being a call to `AllowViewSource()` (the line to which this comment is attached).  We now have the situation that Thread2 is executing within method `showSourceForSelectedRun()` with the `showSourceActive` variable set to false -- but that's wrong; there **_IS_** an active "showSource" operation in progress.

I'm not positive, but I think that removing the FIRST call to `AllowShowSource()`, leaving just the one at the end of the method, might fix the problem.

Or maybe I'm making this whole thing up in my head?

johnbrvc: Yes, it's actually missing a 'return' at line 2272.  Removing the first AllowShowSource() (line 2198) would break the 2 checks that immediately follow since they could 'return' without doing an AllowViewSource():
```
                // OK to now start another view source
                AllowViewSource();
               
                // if any type of error occurred requesting the run to view, log a summary message and finish
                if(bFetchError) {
                    getController().getLog().log(Log.WARNING, "Unable to fetch run " + run.getNumber() + " at site " + run.getSiteNumber() + " from server");
                    showMessage("Unable to fetch selected run; check log");
                   return;
                }
                
                //if we get here we know there should be RunFiles in the contest model -- but let's sanity-check that
                if (!getContest().isRunFilesPresent(run)) {
                    
                    //something bad happened -- we SHOULD have RunFiles at this point!
                    getController().getLog().log(Log.SEVERE, "Unable to find RunFiles for run " + run.getNumber() + " at site " + run.getSiteNumber() + " -- server error?");
                    showMessage("Unable to fetch selected run; check log");
                    return;

                } else {
```

What may be better is to do something like:

```
                // OK to now start another view source
                AllowViewSource();
               
                // if any type of error occurred requesting the run to view, log a summary message and finish
                if(bFetchError) {
                    getController().getLog().log(Log.WARNING, "Unable to fetch run " + run.getNumber() + " at site " + run.getSiteNumber() + " from server");
                    showMessage("Unable to fetch selected run; check log");
                } else {
           
                    //if we get here we know there should be RunFiles in the contest model -- but let's sanity-check that
                    if (!getContest().isRunFilesPresent(run)) {
                    
                        //something bad happened -- we SHOULD have RunFiles at this point!
                        getController().getLog().log(Log.SEVERE, "Unable to find RunFiles for run " + run.getNumber() + " at site " + run.getSiteNumber() + " -- server error?");
                        showMessage("Unable to fetch selected run; check log");

                    } else {
                        //get the Run files
                        ...
                   }
              }
             return;   // THIS IS THE NEW return STATEMENT
         } else {
                // getContest().getRun() returned null
                getController().getLog().log(Log.WARNING, "Selected run not found");
                showMessage("Selected run not found");
         }

    

```

clevengr: Ok, adding that `return` does seem like it would fix the problem.  Can you push an update to your fork so I can pull it down for testing?

johnbrvc: Fixed comment describing possible NPE (which, by the I caused to happen on purpose).
Fix missing return and optimized code flow.

johnbrvc:
@clevengr asked how to test the fix for the wrong source appearing in a View Source window.  This became very difficult to duplicate once PR #504 for Issue #497 was applied.  What caused the bug was the fact that a slow response server (typically a remote site server) did not respond to the correct client when a remote View Source was requested.  This would cause the client that issued the View Source (be it an admin or judge) to wait for up to 30 seconds for a response.  During that 30 second wait, if another runChanged event came in to the client **_for any run_**, the runFiles from **THAT** run would be used in the View Source window since the requesting client did not perform any checking to make sure the response was for the requested View Source run.  There are a couple of ways to simulate this condition.

1) Using the debugger on the remote site's server, a breakpoint can be added when the request for a remote run comes in.  At that point, on the primary server, an admin client can issue an edit run on some other run, and make a change to that run.  This would cause a runChanged event to come in for a different run than the original View Source was waiting for. 
2) If the remote site does not respond to the request for the run.  Run the remote site on a separate PC.  After both the primary and secondary servers are up and running, unplug the network cable on the remote site's PC.  Now, make the View Source request from a judge connected to the primary.  It will wait for a response to the View Source.  Within 30 seconds, using an admin account, edit run on the primary server (site1), and make a change, and update it.  You will see the source for that run appear in the View Source window for judge's View Source request to the disconnect remote.
3) Use older versions (pre PR #504 ) to perform the test:

> For this test you will need a multi-site contest.  site1 (primary) site2 (secondary)
> 1) Use the sumithello sample contest
> 2) start up both servers and make sure everything is ok
> 3) start up a team client on each server and submit 2 different runs on each so there are a total of
>    4 new runs, 2 from site1 and 2 from site2.  Make sure you use different languages and problems.
>    ex.  site1 run1: C hello.c
>         site1 run2: C++ sumit.cpp
>         site2 run1: C++ hello.cpp
>         site2 run2: Java hello.java   
>    or some such combination so all are different.  They don't HAVE to be different but it makes it much
>    easier to see what is going on and what source is appearing.
> 4) close the team clients (for neatness)
> 5) open a judge on site1
> 6) All Runs->select site2 run2 
> 7) View Source  (preparing to display source code for run2 at site2 appears in the message label)
> 8) Before the message times out in 30 seconds,  select site1 run1 from the All Runs list
> 9) View Source  (2 windows will appear both containing the source code for site1 run1)  Only one
>    of those is correct, obviously.  The second one is the one that waiting for the run files to come
>    back from site2 (which never will in this case).  This simulates a very slow response by a remote server;
>    that is, another runChanged event can come in before you get the response back from the original View Source.
>    We are just doing it all using one client here.
> 
> With the new changes, this does not happen for 2 reasons.
> 1) pr504 fixes the fact that site2 did not respond properly, so you actually get the (correct) source back right away
> 2) even if site2 did not respond (you can leave the old version in there for site2's server), the new code
>    will put a message in the message label saying you can not issue another request while one is pending.

clevengr (approved):
I have thoroughly reviewed the source code changes in this PR; I (finally) understand them all and I believe them to be correct and valid.

I have tested the PR in single-site mode, both in the presence of and in the absence of the fixes for the related Issues and PRs (e.g. PR #504), applying all the various "Steps to test this PR" listed both in the PR descriptions and in subsequent comments.

I have also tested the PR in multi-site mode, again both in the presence and the absence of related Issue/PR fixes and by applying all the tests listed in the PR (as well as every other test I could think of).

I approve this PR.

troy2914 (approved):
Seems reasonable to me.

Issue #507, opened 2022-07-27
switch to use Github Actions


troy2914:
so John doesn't have to worry about gitlab anymore :)

troy2914:
the "or" in the if failed.

troy2914:
push-release failed

troy2914:
getting closer.... to push a release....

PR #508, opened 2022-07-27
switch to use Github Actions

troy2914:
This is based on what icpctools does (minus the docker bit).   And using bits from the former .gitlab-ci.yml here.

fixes #507


.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+          name: build
+          path: dist
+      - name: Push release to website
+        run: |
+          eval $(ssh-agent -s)
+          echo "${{ secrets.SSH_PRIVATE_KEY_WEBSITE }}" | tr -d '\r' | ssh-add -
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          export GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=no'
+          git config --global user.email "pc2@troyspaws.com"
```
clevengr, line 110: Shouldn't this be sending email to pc2@ecs.csus.edu?

.gitlab-ci.yml
```diff
@@ -1,224 +0,0 @@
-# GitLab build configuration
```
clevengr, line 1: Might we want to keep this file around, at least until we have the GitHub version known to be working correctly?
(I realize we can always get it out of the history, but...)

clevengr (approved):
I've reviewed and don't see any major issues, so I'm approving.  However, as I commented in Slack, it's not clear to me what actually _triggers_ the invocation of `ci/yml` on GitHub.  How does that happen?

In any case, I'm approving the PR; I don't think it can cause us any major problems (and we probably need to merge it to verify that it does indeed work, or not...)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+    branches:
+      - main
+      - develop
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: pc2fromecs/jdk8hugo
+    steps:
+      - uses: actions/checkout@v2
```
clevengr, line 13: Why is this step using @v2 when the following steps all use @v3?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14.17.x
```
clevengr, line 20: Is this the same version of Node that we were building with on GitLab?   I know that we've had some problems with Node compatibility between our local machines and what's on GitLab....

.github/workflows/ci.yml
troy2914: I was under the impression that pc2@ecs.csus.edu was associated with pc2ccs account,  where as this address is associated with pc2bot.   This was copied from the .gitlab-ci.yml

.gitlab-ci.yml
troy2914: It either works or needs to be fixed.   Not need to keep clutter around, imo.

.github/workflows/ci.yml
troy2914: we were pinning to node 14.

.github/workflows/ci.yml
troy2914: unclear,  that was copied from icpctools

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          eval $(ssh-agent -s)
+          echo "${{ secrets.SSH_PRIVATE_KEY_WEBSITE }}" | tr -d '\r' | ssh-add -
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          export GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=no'
+          git config --global user.email "pc2@troyspaws.com"
+          git config --global user.name "PC2 bot"
+          website/scripts/populate-releases.py ${{ secrets.RELEASE_TOKEN }}
+          website/scripts/copy-readmes.py dist
```
troy2914, line 113: This script is missing.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+          cp doc/ChangeLog.md website/
+          sed -i 's/^# .*//' website/ChangeLog.md
+          sed -r 's/^## (.*)/#### \1/' -i website/ChangeLog.md
```
troy2914: and this ChangeLog.md as well

PR #512, opened 2022-08-01
fixes from johnb.

troy2914:
fi


johnbrvc (approved):

clevengr (approved):
I'll merge so we can see what happens.

PR #517, opened 2022-08-03
fix pre-release

troy2914:
fixes #507

still looking into the pc&sup2;  issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,14 @@ jobs:
       - name: Create GitHub release
         shell: bash
         run: |
+          eval $(ssh-agent -s)
+          echo "${{ secrets.SSH_PRIVATE_KEY_WEBSITE }}" | tr -d '\r' | ssh-add -
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          export GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=no'
+          ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+          git config --global user.email "pc2@troyspaws.com"
+          git config --global user.name "PC2 bot"
           export VERSION=`echo dist/pc2-*.zip | sed 's#^dist.*pc2-\(.*\).zip*#\1#'`
           export VERSION1=`echo dist/pc2-*.zip | sed 's#^dist.*pc2-\(.*\).zip*#\1#' | sed 's/~.*//'`
           if [ github.ref == 'refs/heads/master' ]; then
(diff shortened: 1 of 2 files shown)


johnbrvc (commented):
Did you test out the push-release step after adding the commitish value?  I have a feeling it isn't going to work (I'm still not clear exactly what github wants to see in this field).  I will try to set that up on my test branch today to test it out.  I'd hate to see it merged and have to make another change.  The first one though for pre-release is OK.

troy2914:
> Did you test out the push-release step after adding the commitish value? I have a feeling it isn't going to work (I'm still not clear exactly what github wants to see in this field). I will try to set that up on my test branch today to test it out. I'd hate to see it merged and have to make another change. The first one though for pre-release is OK.

The issue is where were creating a tag_name in nightly-builds,  but trying to use a branch name that is from pc2v9, or a commit hash from pc2v9.  where nightly-builds only has master.   The git checkout and commit means we can actually get the commit hash for the nightly-builds.

johnbrvc:
> > Did you test out the push-release step after adding the commitish value? I have a feeling it isn't going to work (I'm still not clear exactly what github wants to see in this field). I will try to set that up on my test branch today to test it out. I'd hate to see it merged and have to make another change. The first one though for pre-release is OK.
> 
> The issue is where were creating a tag_name in nightly-builds, but trying to use a branch name that is from pc2v9, or a commit hash from pc2v9. where nightly-builds only has master. The git checkout and commit means we can actually get the commit hash for the nightly-builds.

Understood.  But I'll still test it _**just to be sure.**_..

johnbrvc:
> > > Did you test out the push-release step after adding the commitish value? I have a feeling it isn't going to work (I'm still not clear exactly what github wants to see in this field). I will try to set that up on my test branch today to test it out. I'd hate to see it merged and have to make another change. The first one though for pre-release is OK.
> > 
> > 
> > The issue is where were creating a tag_name in nightly-builds, but trying to use a branch name that is from pc2v9, or a commit hash from pc2v9. where nightly-builds only has master. The git checkout and commit means we can actually get the commit hash for the nightly-builds.
> 
> Understood. But I'll still test it _**just to be sure.**_..

I have tested the push-release step and made some changes.  I sent @troy2914 the link to new ci.yml file in my git branch; I can't create a PR right now using my branch because it's polluted with other changes to my develop that I have to resolve.  Once that's integrated it should work.

troy2914:
I have pulled in all but one removal of a space change that he proposed.

clevengr (approved):
I must admit that I do not understand all the changes herein.  However, it appears to me that both Troy and JohnB agree that these changes are appropriate, so I'm approving and then merging so we can see what happens.

johnbrvc:
The process of "debugging" the workflow is tedious and slow.  Turn-around to test on production is about 9+ minutes.   I created a complete test bed on my fork that has nightly-builds, and now the website - I removed a lot of steps from the workflow that take time - like building and testing and packaging all of PC2 every time.  This cut the test turnaround time to about 2 min or so.  To put things in perspective, Take a look at my workflow summary: https://github.com/johnbrvc/pc2v9/actions  It took us 56+ workflow runs to get something that is a _candidate_ to be pushed to the real pc2 repo!  And, even then, we sometimes would "miss" something that we changed to make it work on my repo.  It's a process.  But... I think the push release part is working!

PR #520, opened 2022-08-11
add the Dockerfile to rebuild pc2fromecs/jdk8hugo

### LINKED ISSUE #519: container needs new curl for proper retries.
The container was last built on pc2aries.ecs.csus.edu on May 12th of 2020.
Add the updated Dockerfile to the repo, along with a README

With the upgrade to ubuntu 22.04 we need a `git config --global --add safe.directory ${GITHUB_WORKSPACE} ` 
 in the  "Create GitHub release" stage of ci.yml

troy2914:
with ubuntu22.04 which has the curl version that
supports --retry-all-errors

Also note the /root/node_modules with all the dependecies
for WTI-API are installed on this container.
It would be time safer to "cp -r /root/mode_modules projects/WTI-UI/"
effectively in the projects/WTI-API/buildWTI.xml if /root/node_modules
exists.

fixes #519

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       - name: Create GitHub release
         shell: bash
         run: |
+          git config --global --add safe.directory ${GITHUB_WORKSPACE}
           eval $(ssh-agent -s)
           echo "${{ secrets.SSH_PRIVATE_KEY_WEBSITE }}" | tr -d '\r' | ssh-add -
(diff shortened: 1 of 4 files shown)


troy2914:
grumble, adding python2 and node 14 bits seems to have made it happy.

johnbrvc (approved):
I looked it over, pulled the docker image and verified the utilities appear to be the right versions.
I guess we won't know if this works until it is put on the develop branch since all the changes to ci.yml are only done in that step.

Issue #506, opened 2022-07-20
Invalid command line arguments are silently ignored (no error message or log message)


johnbrvc:
**Describe the issue**: 
If an invalid command line option is supplied to a model at invoke time, it is silently ignored

**To Reproduce**: 
Just supply an invalid command line option to any pc2 command, eg:
pc2admin --bad-option --login a1

**Expected behavior**:
An error message indicating that "--bad-option" is invalid should be displayed and logged and execution should stop.
An invalid option could indicate that the user mistyped something and unexpected results may occur.  Better to let the
user know so it can be corrected. 

**Actual behavior**: 
No message or log message is generated for an invalid option.

**Environment**: please complete the following information:
 All.

**Log Info**:  
No log messages generated on bad options.

**Screenshots**:  
**Additional context**:

johnbrvc:
I looked over Starter.java and ParseArguments.java.  Also, AppConstants.java.  I'm left a bit confused.
It seems that pc2/AppConstants.java has only 5 things it (included here for your convenience):

```
public class AppConstants {
    public static final String HELP_OPTION_STRING = "--help";
    public static final String DEBUG_OPTION_STRING = "--debug";
    public static final String LOGIN_OPTION_STRING = "--login";
    public static final String PASSWORD_OPTION_STRING = "--password";
    public static final String DATETIME_PATTERN_LONG = "E, dd MMM yyyy HH:mm:ss z";
}
```

InternalController.java defines its own identical constant strings for 3 of those (debug, login, password).
It in-lines "--help".  The date pattern is used only by "tools".  It almost looks like at some point, someone
was going to put all the option constant strings in one place (AppConstants - good idea), but it never
materialized.  It seems reasonable (IMHO) to move all the option constants (as a start) from InternalController.java to
AppConstants and use them from there.  Then, a method can be added to AppConstants to verify that
an option is valid, or another class can be created (derived fromAppConstants?) to do the heavy lifting of deciding
if an option is valid. 

Quick hack test attached (rename to .java).  I just did a crude search for all (some regular expression that has "--xxxxx") and dumped everything in the test program - I'm sure there are more options, like the new ones that I have not gotten yet, but
you get the idea.   Optimize: hash table?

[AppConstants.txt](https://github.com/pc2ccs/pc2v9/files/9154441/AppConstants.txt)

johnbrvc:
[...]

PR #524, opened 2022-08-22
Check command line options to make sure they are allowed - Fixed issue #506

johnbrvc:

Centralizes all allowed command line options in the AppConstants class and checks the user supplied command line to make sure any options supplied are allowed.

### Issue which the PR fixes
Fixes issue #506 where unrecognized options were silently ignored.  This could  cause a misspelled option to not be reported to the user, potentially causing undesired operation.

Java ver 1.8.0_321 build 6141 Windows 11 10.0 (amd64) 

From the command prompt, supply an invalid option to any of the pc2 models:
Example 1:
pc2server --server --login s -nogui --contestpassword contest --load sumithello

Before this fix, the "-nogui" option is silently ignored, when the user actually intended "--nogui".

Example 2:
pc2admin -login a1

Before the fix, the pc2admin login screen was presented and no was error reported for the bad "-login" option.

Now, in both cases, the errors are reported and logged to the startup log.

Notes:
This PR makes changes to the ParseArguments class, which is referenced and/or used by 22 other classes.  The changes made to ParseArguments should not affect any of those classes since they use one of the original constructors when allocating a new object.  Only InternalController.java uses the new constructor that takes an additional argument which is the list of allowed options to accept.  By default, this list is null, in which case the checking for allowed options is not done.


src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3020,10 +2954,20 @@ public void start(String[] stringArray) {
             startLog(null, "pc2.startup."+System.currentTimeMillis(), null, null);
         }
         
+        // make sure supplied arguments are recognized (allowed)
+        try {
+            // throw away object to validate allowed options only
+            @SuppressWarnings("unused")
+            ParseArguments checkAllowedArgs = new ParseArguments(stringArray, requireArgumentArgs, AppConstants.allOptionStrings);
```
clevengr, line 2961: It's not really clear what this new constructor call is intended to do; it would be helpful if the comment lines above gave a more detailed explanation of the purpose of making this call.

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3251,19 +3196,28 @@ private Profile getCurrentProfile() {
...
-                    "[" + REMOTE_SERVER_OPTION_STRING + " <remoteHostname> --proxyme]" + //
-                    "[" + MAIN_UI_OPTION + " classname]", // 
+                    "Usage: Starter " + //
+                    "[" + AppConstants.HELP_OPTION_STRING + "] " + //
+                    "[" + AppConstants.FILE_OPTION_STRING + " filename] " + //
+                    "[" + AppConstants.SERVER_OPTION_STRING + "] " + //
+                    "[" + AppConstants.FIRST_SERVER_OPTION_STRING + "] " + //
+                    "[" + AppConstants.LOGIN_OPTION_STRING + " <login> " + //
+                    "[" + AppConstants.PASSWORD_OPTION_STRING + " <pass> ", //
+                    "           " +
```
clevengr, line 3210: What is this series of blanks intended for?  It's not present in the original output.  It seems like it would tend to make the usage string scroll unnecessarily far to the right on the screen (and it's already pretty long...)

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3251,19 +3196,28 @@ private Profile getCurrentProfile() {
...
+                    "[" + AppConstants.LOGIN_OPTION_STRING + " <login> " + //
+                    "[" + AppConstants.PASSWORD_OPTION_STRING + " <pass> ", //
+                    "           " + 
+                    "[" + AppConstants.LOAD_OPTION_STRING + " <dir>|<file> ] " + //
+                    "[" + AppConstants.SKIP_INI_OPTION_STRING + "] " + //
+                    "[" + AppConstants.INI_FILENAME_OPTION_STRING + " INIfilename] " + //
+                    "[" + AppConstants.CONTEST_PASSWORD_OPTION_STRING + " <pass>] " + 
+                    "[" + AppConstants.REMOTE_SERVER_OPTION_STRING + " <remoteHostname> " + AppConstants.PROXY_ME_SERVER_KEY + "]" + //
+                    "[" + AppConstants.MAIN_UI_OPTION_STRING + " classname]", // 
+                    "           " +
```
clevengr, line 3217: Same question as above...

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3251,19 +3196,28 @@ private Profile getCurrentProfile() {
...
+                    "[" + AppConstants.NOLOGGING_OPTION_STRING + "] " + //
+                    "[" + AppConstants.NO_GUI_OPTION_STRING + "] " + //
+                    "[" + AppConstants.NOSTANDINGS_OPTION_STRING + "] ", //
                     "", //
```
clevengr, line 3221: I think there are at least two options missing here (they were missing in the original as well, so it's understandable):

   NO_CONNECTIONS_PANE_OPTION_STRING  and NO_LOGINS_PANE_OPTION_STRING

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3251,19 +3196,28 @@ private Profile getCurrentProfile() {
...
                     "See http://pc2.ecs.csus.edu/wiki/Command_Line for more information", //
```
clevengr, line 3222: This line should be replaced with `https://github.com/pc2ccs/pc2v9/wiki/Command-Line-Arguments`   (I realize it's not part of this Issue/PR, but it's good to replace bad references wherever they pop up...)

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -4120,8 +4074,7 @@ protected void warning(String message, Exception ex) {
     }
 
     private void showErrorMessage(String message, String title) {
-        // TODO 736 fix this to show GUI message
-        System.err.println(title+": "+message); // TODO 736 remove this
+        JOptionPane.showMessageDialog(null,  message, title, JOptionPane.INFORMATION_MESSAGE);
```
clevengr, line 4077: Two comments about this line:

1. Is it guaranteed at this point that noGUI==false?   Do we know for certain that invoking JOptionPane won't throw a HeadlessException?

2. The name of the method is `showErrorMessage()`.  If it is always the case that this method is called with an ERROR, I suggest `JOptionPane.ERROR_MESSAGE` is a more appropriate constant for the last argument in the call to `JOptionPane.showMessageDialog()`.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: The "**usage**" String array [] is a list of strings that are printed to show the command line usage.  Each element is printed using **System.out.println()**.  The idea of the "        " was to indent the second and third lines of options so they are formatted somewhat nicer instead of all at the left margin.  I can remove them and put everything on one line if that is preferred. (You'll notice that the lines preceding the "       " end in a comma, not a plus sign.

Note: I did notice a couple (LOGIN_OPTION_STRING and PASSWORD_OPTION_STRING) of the options in the usage string do not have a trailing ']'.  I have to add those too.

NO_CONNECTIONS_PANE_OPTION_STRING and NO_LOGINS_PANE_OPTION_STRING  were newer options, and I guess they were not added to the original usage message; I can add them.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Each string in the usage array is printed on its own line.  The "     " is meant to indent the following line so it offset from the left of the screen and lines up under the other options.  The comma on the previous line can be changed to a plus to make everything on the same line if that is desired (and, of course, remove the "      ").

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Yes. showErrorMessage is only called if nogui is false.  I checked that.  
I was going to add a test in there for nogui, but checked and everyone who calls it (only a couple of places) checks already.

src/edu/csus/ecs/pc2/core/ParseArguments.java
```diff
@@ -11,6 +11,8 @@
 import java.util.Set;
 import java.util.Vector;
 
+import edu.csus.ecs.pc2.AppConstants;
```
clevengr, line 14: It doesn't appear that this import is used/needed.

src/edu/csus/ecs/pc2/core/ParseArguments.java
```diff
@@ -301,6 +305,24 @@ public boolean isOptPresent(String argKey) {
...
+    public boolean isAllowedOption(String value) {
+        if(allowedOptions != null) {
+            for(String sOpt: allowedOptions) {
+                if(sOpt.equals(value)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+        return true;
```
clevengr, line 323: Is it really correct to say that if `allowedOptions` is null, the input `value` is an "allowed option"?

src/edu/csus/ecs/pc2/core/ParseArguments.java
lane55: agreed, remove unused import (this was not found on develop branch)

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -3251,19 +3196,28 @@ private Profile getCurrentProfile() {
...
-                    CONTEST_PASSWORD_OPTION + " <pass>] [" + NO_GUI_OPTION_STRING + "] "+ //
-                    "[" + NOSTANDINGS_OPTION_STRING + "] "+ //
-                    "[" + REMOTE_SERVER_OPTION_STRING + " <remoteHostname> --proxyme]" + //
-                    "[" + MAIN_UI_OPTION + " classname]", // 
+                    "Usage: Starter " + //
+                    "[" + AppConstants.HELP_OPTION_STRING + "] " + //
+                    "[" + AppConstants.FILE_OPTION_STRING + " filename] " + //
+                    "[" + AppConstants.SERVER_OPTION_STRING + "] " + //
+                    "[" + AppConstants.FIRST_SERVER_OPTION_STRING + "] " + //
+                    "[" + AppConstants.LOGIN_OPTION_STRING + " <login> " + //
```
lane55, line 3208: Add ] where missing.
--login and --password lines are missing closing } char.

lane55 (changes requested):
3 small fixes.

In usage, MIssing 2 closing ] at  InternalController:3208
Unused import in ParseArguments 
Dead code found (should be commented out)  ParseArgumentsTest:137

test/edu/csus/ecs/pc2/core/ParseArgumentsTest.java
```diff
@@ -72,13 +93,48 @@ public void testIsOptPresent() {
...
      * main test routine for ParseArgs <br>
      *
      * @param args
      *            java.lang.String[]
      */
     public static void main(String[] args) {
+        ParseArgumentsTest t = new ParseArgumentsTest();
+        t.testIsOptPresent();
+        t.testAreOptsValid();
+        if(false) {
```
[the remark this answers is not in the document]
johnbrvc: All will be addressed in next push.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Will be fixed in next commit.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Same answer as above.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: The usage message will display as follows:
```
Usage: Starter [--help] [-F filename] [--server] [--first] [--login <login>] [--password <pass>]
           [--load <dir>|<file> ] [--skipini] [--ini INIfilename] [--contestpassword <pass>]
           [--remoteServer <remoteHostname> server.proxyme] [--ui classname]
           [--nologging] [--nogui] [--nostandings] [--noconnectionspane] [--nologinspane]

```

johnbrvc: There are a couple of options that do not appear in the usage message.  One is not used anywhere (--id xxx) should it be removed?   The --nosave, --port xxx, and --team1 options are used, but not listed in the help.  Should they be?  I think --team1 is debugging shorthand option.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Will be updated in the next push.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: The showErrorMessage() is a private method to InternalController.  It is called in exactly 2 places in InternalController.java and in both places, "usingGUI" is checked.

johnbrvc: 2. Agreed.  JOptionPane.ERROR_MESSAGE is more appropriate (and puts a red X in the dialog instead of the "i").

src/edu/csus/ecs/pc2/core/ParseArguments.java
johnbrvc: Will be removed in latest push.

src/edu/csus/ecs/pc2/core/ParseArguments.java
johnbrvc: Yes.  If the allowedOptions array is not set, then all options are allowed and no checking is done.  This is the default behavior to maintain compatibility with other classes that use this class.  I clarified the comment for the isAllowedOption method.

src/edu/csus/ecs/pc2/core/InternalController.java
lane55: I think that --port should be added
Do not include the following undocumented option (undocumented debugging options)
The --nosave (do not save anything to disk, only for a server) and 
--team1  shorthand for joe login on team 1

lane55 (approved):
Looks good.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: New usage message changed to read as follows including --port option:

```
Usage: Starter [--help] [-F filename] [--server] [--first] [--login <login>] [--password <pass>]
           [--load <dir>|<file> ] [--skipini] [--ini INIfilename] [--contestpassword <pass>]
           [--remoteServer <remoteHostname> server.proxyme] [--ui classname] [--port <port>]
           [--nologging] [--nogui] [--nostandings] [--noconnectionspane] [--nologinspane]
```

src/edu/csus/ecs/pc2/core/ParseArguments.java
```diff
@@ -471,6 +499,27 @@ public void setRequireArgOpts(String newRequireArgOpts) {
...
+     * Set a single allowed option.
+     *
+     * @param newAllowedOpt
+     *            java.lang.String
+     */
+    public void setAllowedOpt(String newAllowedOpt) {
+        allowedOptions = new String[1];
+        allowedOptions[0] = newAllowedOpt;
+    }
+
```
clevengr, line 522: Unless I'm reading things wrong, if someone first calls `setAllowedOpts(String [])`, and then later makes a call to `setAllowedOpt(String)`, the effect would be to _wipe out_ all the options which were set by the original call to `setAllowedOpts(String [])`.  I would not expect that to be desired/correct behavior; if I set a collection of options with a call to `setAllowedOpts(String [])` and then call `setAllowedOpt(String)`, I would expect the new "allowed option" to be ADDED to the set of (previously-defined) allowed options.  

It seems to me that method `setAllowedOpt(String)` should check for a non-null `allowedOptions` and if present it should ADD the specified option to the existing list.

johnbrvc: For consistency, I made the new methods work the same way that "RequiredArgOpts" methods work.  That is, either "set" method (String[]) or (String) creates a completely new array.  I would expect if you wanted to append an additional option that's allowed, there would be an "addAllowOpt()" method.

clevengr: Ok.  I could argue that the design of the original methods, from which you say these are copied, is also flawed in the same way... but let's not go down that rabbit hole.

I do think it would be helpful if the methods (at least the new ones) had a comment in their header that at least made it clear that calling either one of them has the side-effect of throwing away any previously-set options.  That way at least future developers have some documentation that warns them...

test/edu/csus/ecs/pc2/core/ParseArgumentsTest.java
```diff
@@ -72,31 +92,66 @@ public void testIsOptPresent() {
...
+//      for (String s : reqArgs){
+//          System.out.print(" "+s);
+//      }
+//      System.out.println();
+//      
+//      pa = new ParseArguments();
+//      pa.setRequireArgOpts(reqArgs);
+//      pa.loadArgs(args);
+//      pa.dumpArgs(System.out);
+//      }
```
clevengr, line 154: Personally I would rather see blocks of deprecated code like the removed, rather than just "commented out"; the presence of the old code makes it harder to read the current version.  Especially when the old code is characterized by things like blocks that start with `if (false)...`.   

However, this doesn't rise to the level of needing to deny approval of the PR...

clevengr (approved):
I made a few additional comments which would be nice to have addressed, but none of them rise to the level of blocking this PR.

test/edu/csus/ecs/pc2/core/ParseArgumentsTest.java
lane55: Agreed, old/unused code should be removed.   My comment/problem that 1 - if(false) dead code
2 the dead code would be found if grep'ing the code which requires extra analysis/time to
to identify that it is old/dead code

johnbrvc: Removed vestigial code and pushed.

clevengr (approved):

PR #526, opened 2022-08-28
javac change bootclasspath to classpath

### LINKED ISSUE #525: ant build fails with jdk17
on my mac with jdk17 trying to ant or ant -f package.xml was resulting in a failed build with this error:
```
    [javac] Compiling 904 source files to /Users/boudreat/git/pc2v9/build/prod
    [javac] Fatal Error: Unable to find package java.lang in classpath or bootclasspath
```

fix as documented on https://coderanch.com/t/109305/build-tools/Fatal-Error-Unable-find-package

troy2914:
fixes #525

to test, install a newer jdk 17+ and run 
ant
it should work with this fix, before my fix it would have failed on the first javac.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -175,7 +175,7 @@ TODO consider switching back to this at 9.8.0/9.9build
         <mkdir dir="${build.test.dir}" />
         <mkdir dir="${compilelogs.dir}" />
         <!-- Compile the java code from ${src.dir} into ${build.dir} -->
-        <javac srcdir="${src.dir}" destdir="${build.prod.dir}" failonerror="${javacFailOnError}" verbose="${javacVerbose}" debug="${javacDebugInfo}" includeAntRuntime="no" bootclasspath="${bundleBootClasspath}" source="${bundleJavacSource}" target="${bundleJavacTarget}">
+        <javac srcdir="${src.dir}" destdir="${build.prod.dir}" failonerror="${javacFailOnError}" verbose="${javacVerbose}" debug="${javacDebugInfo}" includeAntRuntime="no" classpath="${bundleBootClasspath}" source="${bundleJavacSource}" target="${bundleJavacTarget}">
             <compilerarg line="${compilerArg}" compiler="${build.compiler}"/>
             <classpath refid="project.classpath"/>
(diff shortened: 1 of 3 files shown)


clevengr:
I have reviewed the code changes in this PR.  However, I confess I do not understand them (which is possibly due to my lack of knowledge).  @troy2914: can you explain why changing the value of `classpath` is being done?    Won't this have an effect on every local installation which pulls this PR?   How is the "new" classpath different from the old one (and don't tell me "the classpath is now the same as `${bundleBootClasspath}`"; that's obvious from the code.  My question is:  WHY is it now appropriate to change the definition of classpath to be equal to `${bundleBootClasspath}`?

troy2914:
> I have reviewed the code changes in this PR. However, I confess I do not understand them (which is possibly due to my lack of knowledge). @troy2914: can you explain why changing the value of `classpath` is being done? Won't this have an effect on every local installation which pulls this PR? How is the "new" classpath different from the old one (and don't tell me "the classpath is now the same as `${bundleBootClasspath}`"; that's obvious from the code. My question is: WHY is it now appropriate to change the definition of classpath to be equal to `${bundleBootClasspath}`?

The issue is javac wants the classpath set, not the bootclasspath.   At least with 
```
javac 17.0.1
```

I guess the real question is does this change break the java8 javac?

troy2914:
John, have you actually tried the changes to make sure it does not break building for you?

troy2914:
Sam had the same error I had with eclipse 2020-03 and 2022-12,  this change fixes it for him too.

johnbrvc:
> John, have you actually tried the changes to make sure it does not break building for you?

I assume you mean JohnC, however, I just tried your build.xml and it works fine with eclipse 2021-12 and JDK 1.8.0_321.
The build.xml is missing changes that have occurred since you did the pr, IE it needs to have new stuff and build
targets merged into it.

If it works for others, it's probably a good thing to merge into the latest build.xml.. IMHO.

clevengr:
No, I have not done anything with this for quite a while...

troy2914:
Sam noted I did not update the create xml,  I have those waiting on my computer to have a stable internet.

ashoos72 (approved):
tested on eclipse 2020-03 verified working.

Issue #496, opened 2022-07-17
Redesign AJ "fetch run" mechanism.


clevengr:
**Is your feature request related to a problem?** 

Yes. The current implementation, where AutoJudges "request" a specific run, has been determined to have numerous flaws (including that multiple AJs configured to judge the same problem(s) frequently end up requesting the same run, causing one of them to get the run and the others to have wasted time and now needing to go back and request a different run -- which might end up also being requested by other AJs, etc.  

**Feature Description**: 

Redesign the "AJ run acquisition" mechanism so that instead of an AJ requesting a specific run, AJs "register" as "AVAILABLE_TO_JUDGE", and on the server side when a run is received the server examines the list of "Available AJs" (restricting itself to only those that are configured to judge that particular problem), and then operate as if a "Run Checkout Request" had been received -- i.e., sending the received run to the selected AJ.

**Have you considered other ways to accomplish the same thing?** 
Other than staying with the current, flawed, implementation, nothing comes to mind.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, please describe.  

- Add a new PacketType AJ_AVAILABLE.
- Have each AJ send such a packet when it starts up (instead of searching for a run to request).
- Have each AJ send such a packet when it _finishes_ judging a run (i.e. after sending a JUDGEMENT packet).
- Have the server maintain a list "AvailableAJs" of AJs which have sent an AJ_AVAILABLE packet.
- Have the server implement a loop such as
```
while (there are runs to be judged) {
  Search the availableAJs list for a judge configured for the problem;
  if (found) {
      Remove the AJ from the availableAJs list;
      Checkout the run and send it to the selected AJ;
  }
}
```
(The above is high-level pseudo-code for the concept; there are lots of details such as synchronization, exactly where the above loop belongs, what to do if there's no AJ available, etc.)

lane55:
I disagree with using a loop.   The processing for a new run or AJ_AVAILABLE should
have individual logic/code, rather than having to debug code within a loop.   These
are isolated events and should not be clumped into a while loop, unless there is a
condition that requires a loop.

lane55:
The solution should include a way for an Admin to view the AJ Lists, either via a live monitoring (unlikely, large LOE)
or a snapshot where the Admin requests the list contents and the server sends a packet with the list contents.

lane55:
On server startup, there would be a loop to go through runs.

lane55:
[...]

PR #530, opened 2022-09-02
I 496 aj run flow redesign

lane55:

Initial work on AJ redesign.   New Data/Lists added.  Minimal code.

Added new report (only works on the Server UI at this time): Autojudge Lists Report
which shows the current AJ list and contents.    Since the AJ lists are only stored
on the server, at this time the Autojudge Lists Report is the only way to examine the lists.

### Issue which the PR fixes

Someday #496

Config server for auto judging contest
Submit runs
Runs should be auto judged.

### DIFF
--- a/src/edu/csus/ecs/pc2/AppConstants.java
+++ b/src/edu/csus/ecs/pc2/AppConstants.java
@@ -6,6 +6,7 @@ package edu.csus.ecs.pc2;
  * 
  * @author Douglas A. Lane, PC^2 Team, pc2@ecs.csus.edu
  */
+// TODO i 496  repleace class with head
 public class AppConstants {
 
     public static final String HELP_OPTION_STRING = "--help";
@@ -15,6 +16,8 @@ public class AppConstants {
     public static final String LOGIN_OPTION_STRING = "--login";
 
     public static final String PASSWORD_OPTION_STRING = "--password";
+    
+    public static final String CONSOLE_OPTION_STRING = "--console";
 
     public static final String DATETIME_PATTERN_LONG = "E, dd MMM yyyy HH:mm:ss z";
 }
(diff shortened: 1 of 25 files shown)


johnbrvc:
Picky, defensive programming things: In AutoJudgeManager::findRunToAutoJudge(), I might protect each step of the list removals with their own try/catch and attempt some kind of recovery if one step should fail.  I also may change the order of the removes.  That is, remove the judge from the judge list availableAJList first, THEN the run from the availableAJRunList.  The run list is more important than the judge list, so if the judge remove causes an exception, the run isn't lost forever.  Similarly, if the contest.getRun() should fail, we'd probably want to re-add the judge back to the AvailableAJList so it's not lost forever; the run in this case isn't as important since it's probably bad anyway if the getRun() failed - no great loss.

johnbrvc:
In AutojudgeListReport, it may be nice to include the problem Letters in addition to the short name in the output.

lane55:
> Picky, defensive programming things: In AutoJudgeManager::findRunToAutoJudge(), I might protect each step of the list removals with their own try/catch and attempt some kind of recovery if one step should fail. I also may change the order of the removes. That is, remove the judge from the judge list availableAJList first, THEN the run from the availableAJRunList. The run list is more important than the judge list, so if the judge remove causes an exception, the run isn't lost forever. Similarly, if the contest.getRun() should fail, we'd probably want to re-add the judge back to the AvailableAJList so it's not lost forever; the run in this case isn't as important since it's probably bad anyway if the getRun() failed - no great loss.

Agreed.  Will updae code and then push.

lane55:
> In AutojudgeListReport, it may be nice to include the problem Letters in addition to the short name in the output.

Pushed fir that adds letter to start of short name, Ex.
There are 1 Judges available in Auto Judge list
    JUDGE4 @ site 1 2 problems: A - hello, B - sumit
There are 0 Runs in Auto Judge list

lane55:
> Picky, defensive programming things: In AutoJudgeManager::findRunToAutoJudge(), I might protect each step of the list removals with their own try/catch and attempt some kind of recovery if one step should fail. I also may change the order of ...
Pushed new re-ordered removes and defensive code.

johnbrvc (commented):
This is more or less a test comment to see if it starts a conversation.  Seems correct.

johnbrvc (changes requested):
This is more or less another test comment (using Request Changes) to see if it starts a conversation.

lane55:
closing pull request.  This was a preview, when more work updated/done will push another PR

PR #531, opened 2022-09-07
Initial changes for supporting 2022-07 CLICS Event Feed - fixes #523 #543 #546

### LINKED ISSUE #523: Update Shadow to read new CLICS event feed

Yes, the CLICS team has changed the event feed format/content.   It may
be incompatible with pc2 parsing/code (in Shadow).

**Feature Description**: 

Update the Shadow code to read from the CLICS Event Feed.

Yes, the lowest LOE would be to update the JSON parsing.

If there is time, refactor and move code that parses input JSON
line into its own method.

The java/OOP way is not to use
(String) submissionEventDataMap.get("id");
it is to use
SubmissionEvent.getId();

Something like the already used class ShadowRunSubmission.

**Additional context**:

johnbrvc:
Preliminary changes that add code to handle the removed "**op**" field in the event feed notification object. Add code to support both old and new CLICS feeds (2020-03 and 2022-07, and presumably 2021-11).

This only applies to event feed notifications of type "**judgements**", since that appears to be the only place the new event feed format affects PC2: If the "**op**" field in the notification object on the event feed is missing (**null**), we assume this is version 2022-07 of the CLICS specification and get the "**id**" of the judgment from the notification object (since there is no "**data**" object), otherwise, if the "**op**" field is present and contains "_delete_", then we use the "**id**" in the "**data**" object.

Note: it would be nice if there was a "version" string in the CLICS feed specifying what version of the specification the feed is.

### Issue which the PR fixes
Hopefully, #523 
Java Version    : 1.8.0_321

Run a shadow using the CLICS event feed version 2022-07 (currently, non-existent as of this writing).  Make sure results are correct.  To fully test the code, the event feed should contain a deleted judgment, or perhaps a hand-craft one?
eg:
**{"type": "judgements", "id": "11416647",  "data": null }**
would presumably delete judgement with "**id**" "11416647" (no example was available)

The 2020-03 version of the message might be:
**{"type": "judgements", "id": "4beeb620-c4fe-4e19-b4c4-7072b10b0a4c", "op": "delete", "data": {"id": "11416647", "end_time": null, "start_time": "2022-05-30T14:06:36.834+00", "max_run_time": null, "submission_id": "9007717", "end_contest_time": null, "judgement_type_id": null, "start_contest_time": "0:06:36.834"}}**
would also delete judgment with "**id**" "11416647"


src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
```diff
@@ -94,7 +96,7 @@ public class RemoteEventFeedMonitor implements Runnable {
      * @param pc2Controller an {@link IInternalController} for passing error handling back to the local PC2 system
      * @param remoteContestAPIAdapter an adapter for accessing the remote contest API
      * @param remoteURL the URL to the remote CCS
-     * @param login the login (account) on the remote CCS
+     * @param k the login (account) on the remote CCS
```
clevengr, line 99: I do not understand this change.  There is no parameter named "k" in the Constructor's argument list, and further it appears that "login" is still the correct parameter name.

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
```diff
@@ -162,7 +162,8 @@ private Account getSubmitterId(IInternalContest internalContest, String clientId
         // search for this site's logins first
         for (Account account : accounts) {
             if (account.getSiteNumber() == thisSiteNumber) {
-                if (account.getClientId().getName().equals(clientIdString)) {
+                if (account.getClientId().getName().equals(clientIdString) ||
+                    account.getAliasName().equals(clientIdString)) {
```
clevengr, line 166: Hmmm.... I'm suspicious of the new call to `getAliasName()`.    Indeed, PC2 provides for an alternative account name, called an "alias"; the "alias" is intended to be used to support the ability to display a meaningful "name" to a judge without allowing the judge to actually know what (real) team they are working with (thus protecting the judge from allegations of "favoring" a particular team because the judge knew which team he/she was working on).  

For example, "UC Berkeley" can be "aliased" to some non-identifying name such as "TeamWithPythonIssue"  or something more generic like "Team B";  thereafter, any time the (human) judge sees a team name on a submission it will appear as "TeamWithPythonIssue" (or "Team B") instead of "UC Berkeley".   

Displaying (or not) of "aliases" is controlled in the Admin>ConfigureContest>Settings tab, under "Judge Settings>Team Information Displayed to Judges"  (by selecting the "Show Alias" radio button).

Note that a team "alias" is NOT the same thing as a team "Display Name".  If you select "Accounts>Edit Account" on the Admin, you'll see that a team has both an "Account"  (e.g. "team500"), a "Display Name" (which _defaults_ to the same thing as the "account" _but can be changed to any value_.)  It is the "Display Name" which appear on the scoreboard.  

But accounts also have a separate "Alias" field, which can also be set to any value.  The "alias" value is only used (as far as I can recall) in the above-described "Judge Settings".

Sooo... I have a strong suspicion that the above addition referencing `account.getAliasName()` is actually intended to call `account.getDisplayName()`.  Please either confirm this and make the change, or explain why `getAliasName()` is being called.

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
```diff
@@ -174,9 +175,10 @@ private Account getSubmitterId(IInternalContest internalContest, String clientId
         // Sort by site, type then id.
         Arrays.sort(accounts, new AccountComparator());
 
-        // serach all site's logins
+        // search all site's logins
         for (Account account : accounts) {
-            if (account.getClientId().getName().equals(clientIdString)) {
+            if (account.getClientId().getName().equals(clientIdString) ||
+                account.getAliasName().equals(clientIdString)) {
```
clevengr, line 181: Same comment as above:  I suspect this was intended to call `account.getDisplayName()` rather than `account.getAliasName()`....  yes?

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
johnbrvc: That is a mistake on my part.  Looks like I pressed a "k" instead of "ctrl-k" when I was searching for "login", and it wrote a 'k' over the login I was on.  I'll put it back to "login".

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
johnbrvc: The call to `getAliasName()` is what was intended.  After a discussion with @clevengr we are on the same page as to how this change works.  The idea is to allow an arbitrary **team_id** (such as '**domjudge**') to be matched.  Normally this is not possible since all team accounts are always **team###**.  By using the _alias_ mechanism, we can set the _alias_ to an arbitrary string and the remote submitter can then lookup either an expected **team_id** (a number such as 521) or an arbitrary string (using the _alias_).  There was some discussion about using the _display name_ instead of _alias_, but the issue there is we wanted to keep the _display name_ as a meaningful string that conveyed what the account was for, eg. "_DOMJudge Jury Test Team_".  Note that all normal team accounts (**team###**) will have an empty _alias_ field, and this change will not affect the lookup of those team accounts.  Only the special test accounts would have an _alias_ manually defined by the administrator based on what the remote CCS uses for team test account(s).

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
johnbrvc: The intention was `account.getAliasName()` as discussed in the previous comment.

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
lane55: > Hmmm.... I'm suspicious of the new call to getAliasName(). Indeed, PC2 provides for an alternative account name, called an "alias"; the "alias" is intended to be used to support the ability to display a meaningful "name" to a judge without allowing the judge to actually know what (real) team they are working with (thus protecting the judge from allegations of "favoring" a particular team because the judge knew which team he/she was working on).

The original intention for alias name was as @johnbrvc and I discussed, it was to
allow a user to login as judge1 or as jo.   Later, it was then used/leveraged as an option
in the Team Information Displayed to Judges.

In both cases the alias is used as an alias for the team.   Change approved.

lane55 (changes requested):
Most changes are fine.

lane55 (changes requested):
Most changes are fine.    

The run method is too long already, the handling of the delete event
should be move to another method.

The run method was already 486  lines (385 lines without huge comment blocks)
methods should almost never (switch statements are one exception) be
that long..

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
```diff
@@ -491,26 +491,42 @@ public void run() {
...
                                         String operation = (String) eventMap.get("op");
-                                        if (operation != null && operation.equals("delete")) {
-                                            
-                                            //it is a delete; remove from the global map the judgement whose ID is specified
-                                            String idToDelete = (String) judgementEventDataMap.get("id");
-                                            synchronized (remoteJudgementsMapLock) {
-                                                getRemoteJudgementsMap().remove(idToDelete);
-                                            }
-
-                                            //TODO: how do we notify the local PC2 system that this judgement should be deleted??
```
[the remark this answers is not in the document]
johnbrvc: Since DOMJudge does not delete judgements, we do not need this for WF.  Perhaps a separate issue should be created for deleting the judgement, and the comment above moved there?

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
```diff
@@ -491,26 +491,42 @@ public void run() {
...
-
+                                        String idToDelete = null;
+                                        boolean isDelete = false;
+                                        
+                                        // check if this is a "delete" event
+                                        // For the 2022-07 (and 2021-11) clics spec, "operation" will always be null since the "op" field was removed
+                                        // For the 2020-03 the "op" field will be non-null.  This is how we determine the feed type.
```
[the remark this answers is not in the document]
johnbrvc: The code in question does not delete a run, rather, it deletes the judgement.  As such if we were to delete runs, we would have to process both delete "submissions" and delete "runs" events - we current do not do this.  Since we do not need this for WF, perhaps this should be rolled into a separate issue as described above for a deleted judgement.

I will, however, create a method to detect if an event is a delete event ("data":null) for new event feed, or op "delete" for the old.  I thought we said we are no longer supporting the old feed for the shadow, so perhaps remove the "compatibility" test for "op" propertery?  For now, it's harmless though.

johnbrvc:
> 

Will create a method to detect if an event is a delete event.
Will create a method to handle a "judgements" delete operation and move the code from run() to that method.

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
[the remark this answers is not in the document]
clevengr: First:  @lane55 :  this section of code is not about deleting RUNS, it is about deleting JUDGMENTS.  

Second, this section of code is not about deleting anything in the PC2 server; it is about deleting a reference to a judgment _in the RemoteEventFeedMonitor's map of remote judgments_ (a static map that maps a _remote system_  judgment id to a record containing the submission id and the judgment type id (i.e. the judgment acronym) separated by a colon (see method `RemoteEventFeedMonitor.getRemoteJudgementsMap()`)).

Third, yes, there is an issue about how to go about informing the PC2 server that a judgment has been deleted.  This was noted in the comment in line 509 in the original code (above). That line has (correctly) been deleted in this PR, and has been (correctly) replaced with a corresponding comment (at line 814) in (new) method `deleteJudgment(eventMap)`, below.     

The issue of how to handle "delete judgment" events from the remote CCS may be important (or maybe not, since as @johnbrvc  points out DOMjudge never sends any such events), but dealing with it is outside the scope of this Issue/PR.

clevengr (approved):
I have reviewed all the changes in this PR and I approve these changes.  I would like to get @lane55 to review the various comments and indicate approval also prior to actually merging.

johnbrvc:
Modified title of PR to include issues resolved: #543 #546 in addition to the original #523

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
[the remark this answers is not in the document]
lane55: understood.

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
lane55: Can this be resolved John C. ?

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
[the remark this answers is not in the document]
lane55: understood.

lane55 (approved):
Previous PR work/comments done.  PR Approved.

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
clevengr: I'm not sure why the fix isn't showing here in the GitHub conversation, but it does look like it was corrected in the updated copy which I pulled from the fork, so I'll mark it resolved (it's certainly not worth holding up the PR for...)

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
clevengr: I still think this is somewhat problematic -- now we have _triply_ overloaded the meaning and use of "alias".  I see the possibility that this use of the same field for three different purposes might come back to bite us later.

However, having said that, after conversations with @johnbrvc I agree that the change will work here as intended, and given the _extremely_ small number of times that `alias` gets used for anything (at least, as far as we're aware), it's probably not that big an issue and certainly can be dealt with later.  In any case, it's certainly not worth holding up the PR (which is why I "approved" the PR earlier, even with this change present).  I'll mark the comment "resolved".

PR #533, opened 2022-09-14
i532: add Scoreboard JSON Report

### LINKED ISSUE #532: Add new report that for CLICS API scoreboard.json

No, just a change in how to get a scoreboard.json file.

At this time one must start the API (via feeder) and authenticate
and fetch the scoreboard.json

Add the scoreboard.json as a report so an Admin can save
the file, rather than having to startup a feeder client.

**Feature Description**: 

Add a new Report "Scoreboard JSON" that will output the CLICS API scoreboard.json

Yes, no others.

Add a new Report "Scoreboard JSON" 

**Additional context**:

lane55:

Adds Scoreboard JSON Report on UI and in CLI Reports

### Issue which the PR fixes

Fixes #532 

Microsoft Windows [Version 10.0.19044.1889]
java version "1.8.0_121"

Use the Reports tab or pc2report to use/print Scoreboard JSON report.

This should match the CLICS API endpoint scoreboard  (via the Feeder)


johnbrvc (commented):
I tried running this on your branch.
The problem I see is that when it brings up the the view of the JSON report, the entire report is on a VERY long single line, making the view essentially useless.  It would be nice if it looked like the scoreboard.json found in results/scoreboard.json (attached in zip), where each record is on its own line in the view.   Looking at the file produced by the Scoreboard JSON in the reports folder (attached here as a zip) it is one single long line.
[report.Scoreboard_JSON.03.20.987.zip](https://github.com/pc2ccs/pc2v9/files/11023524/report.Scoreboard_JSON.03.20.987.zip)
[results_scoreboard.zip](https://github.com/pc2ccs/pc2v9/files/11023527/results_scoreboard.zip)

lane55:
This report is a direct use of the scoreboard end point, which returns a single JSON string on one line.

johnbrvc:
> 

Yes, I saw that.  It generates XML, then converts it to JSON.  I guess the question I'm asking is, what is the purpose of this report if you can't view it?  Is it just to generate a scoreboard json?  isn't that already done by the scoreboard client in the results folder?
If the intent is to simply generate a json, then it seems the header and trailer should be omitted all the time, since the report file is essentially not viewable by a human, so why add the header and trailer?

lane55:
One intent of the report is to provide the same JSON as the scoreboard end point
(ex curl -u admin:admin --insecure https://localhost:50443/contest/scoreboard  )
From the description in

lane55:
In order to format the JSON to be a nljson there are a couple of options:
1 - add an option to the report to output nljson (like for XML)
2 - add an option in the file viewer to display the contents in nljson form
In that way all JSON output (like for other endpoints) would benefit from those options.

johnbrvc:
> 

Agreed.  That can be an enhancement for a future time.
Other than that, I'm fine with the PR.  It does what it is supposed to.

johnbrvc (approved):

Issue #543, opened 2022-09-26
Event type not properly detected on event feed to cause delay


johnbrvc:
**Describe the issue**: please provide a clear and concise description of what the bug is.  
RemoteEventFeedMonitor looks for events of type "organizations" and "runs" and if it is not one of those a delay of 500ms is introduced to pace the Event Feed thread so it does not overload the AWT thread.  This test is insufficient when the EventFeed is coming from DOMJudge and a delay will be performed for EVERY packet.

**To Reproduce**: please describe, in order, the steps to reproduce the behavior:
Connect to a DOMJudge event feed and note that a delay is performed for every event type including "runs" and "organizations".  This causes the processing of the feed to be much slower.

**Expected behavior**:   
For events of type "organizations" and "runs", no delay should be performed.

**Actual behavior**: 
A delay is performed for all events.

**Environment**: 
 OS: Linux 5.4.0-58-generic (amd64)
Java Verson: 11.0.9.1

**Log Info**:  
N/A

**Screenshots**:  if applicable, please add attachments showing screenshots to help explain the issue.

**Additional context**: 
The problem is that the DJ event feed does not insert a space in the JSON after the ":", eg.
`"type":"submissions"`
where as Kattis does.  PC2 currently looks for a space after the ":" and does not properly detect the 2 types it checks for:
`"type": "submissions"`

johnbrvc:
Closed as part of PR #531

PR #547, opened 2022-09-29
I 536 add contest vs event feed model comparison

### LINKED ISSUE #536: Add a report that will compare CLICS API configuration with pc2 model/configuration

Yes, need to compare and verify contest model is the
same for primary and current pc2 config/model.

**Feature Description**: 

Create a tool that will compare another CCS's API event feed with
the pc2 configuration.

Yes, a new tool will be better.

Add a CLI tool that will output a summary of the comparison
between a CLICS API model and the pc2 model.

Including:
judgements
languages
problems
contest information

**Additional context**:

lane55:

Add a report that compares a remote CCS API model (via event feed) with 
the pc2 model.  The report prints results.

Notes:
* The reading of the Primary CCS event feed is not robust, will likely leverage robustness/retry work from Issue #543 
* Note there is an option, below, to read from an event feed file rather than the Primary CCS API
* need a better name for the report, now is 
* There is debug when reading from Primary API to stdout for each event, will be removed in a future commit

Read the event feed from the Primary CCS, print comparison result

### Issue which the PR fixes

#536 

Note: This compares configuration, there is no need to submit runs.

Configure contest (load CDP)
Start Server
Run Report  Compare Primary with model

To use a event feed file rather than reading the API, add the following into pc2v9.ini for file /tmp/event-feed.json
```
[server]
eventFeedFile=/tmp/event-feed.json
```


src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
```diff
@@ -0,0 +1,64 @@
...
+    private String type;
+
+    @JsonProperty
+    private String id;
+
+    @JsonProperty
+    private String op;
+
+    @JsonProperty
+    HashMap<String, Object> data;
```
johnbrvc, line 34: I'm not sure what happens if the 'data' member is null (as in a delete event) on the feed.  Will that cause an exception?

src/edu/csus/ecs/pc2/services/core/JSONUtilities.java
```diff
@@ -337,4 +341,31 @@ public static String getLanguageIndexString(IInternalContest contest, ElementId
...
+     * This method uses the Jackson {@link ObjectMapper} to perform the conversion from the JSON
+     * string to a Map.  Note that the ObjectMapper recurses for nested JSON elements, returning
+     * a appropriate Object in the Map under the corresponding key string.
+     * 
+     * @param jsonString a JSON string to be converted to a Map
+     * @return a Map mapping the keys in the JSON string to corresponding values, or null if the input
+     *          String is null or if an exception occurs while converting the JSON to a Map.
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<String, Object> getMap(String jsonString) {
```
johnbrvc, line 355: Probably should update RemoteEventFeedMonitor to use this method since it has the same method.

lane55: Noted, out of scope though.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: I have not seen any null data in the event feeds, so unlikely to happen.

src/edu/csus/ecs/pc2/AppConstants.java
```diff
@@ -162,4 +162,10 @@ public class AppConstants {
      */
     public static final String DATETIME_PATTERN_LONG = "E, dd MMM yyyy HH:mm:ss z";
     
+    
+    /**
+     * The value is a full path to a event-feed.json file.
+     * 
+     */
+    public static final String OVERRIDE_EVENT_FEED_FILE = "server.eventFeedFile";
```
clevengr, line 170: The comment says "the value is a full path", but the value doesn't appear to be a full path...
I suggest either correcting the comment, or providing a full path in the value.

lane55: will update comment, note there is no check or control over whether uses a full path, partial path or no path,.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
clevengr: I think what @johnbrvc meant (JohnB, correct me if I'm wrong) is that whereas in the old EF format we would have seen an "op = delete" field, in the 2022-07 (and forward) EF formats we don't get an "op" -- rather, a "delete" is indicated by the occurrence of a "data:null" field.  So you may not have seen any, because there (probably) aren't any in the particular 2202-07 event feeds we've had so far (just one, from pretest2).   I *think* what JohnB was asking is, what happens in a HashMap if we get a "data:null" in the event feed (which we might) and we try to put that in the HashMap?  (Again, JohnB: speak up if I've misinterpreted the issue...)

src/edu/csus/ecs/pc2/services/core/JSONUtilities.java
clevengr: I agree, this is something that should be done but is out of scope for this particular issue.   Hopefully, one of us will at some point remember to file a separate Issue for this...

src/edu/csus/ecs/pc2/core/imports/clics/CLICSContests.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @JsonProperty
+    private String scoreboard_freeze_duration;
+
+    @JsonProperty
+    private String formal_name;
+
+    @JsonProperty
+    private String penalty_time;
+
```
clevengr, line 40: The above collection of attributes does not cover everything listed in the CLICS 2202-07 specification.  For example, the spec also defines that a `contest` object may contain `countdown_pause_time`, `scoreboard_type`, `banner`, `logo`, `location.latitude`, and `location.longitude`.  Since we are trying to bring PC2 code up to CLICS specs, I think this class should at least declare all of the objects (fields) defined in the spec.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSEventType.java
```diff
@@ -0,0 +1,84 @@
...
+    STATE("state"),
+    /**
+     * submissions event
+     */
+    SUBMISSIONS("submissions"),
+    /**
+     * teams event
+     */
+    TEAMS("teams");
+
```
clevengr, line 68: The above list of CLICS events type is missing the `awards` event type defined in the CLICS 2202-07 specification.  Since we are trying to bring PC2 code up to  compliance with the spec, I think this class should include all event types which are defined in the spec.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSLanguage.java
```diff
@@ -0,0 +1,45 @@
...
+    @JsonProperty
+    private String id;
+
+    /**
+     * Language title, ex Java or Python 3
+     */
+    @JsonProperty
+    private String name;
+
+    public String getId() {
```
clevengr, line 31: The above set of fields is missing a number of Language attributes defined in the CLICS 2202-07 spec, including `entry_point_required`, `entry_point_name`, `extensions`, `compiler`, and `runner`.    It's also missing definitions for _all _ of the fields defined in the `Command` object set (under Languages) in the spec.  Since_ we are trying to bring PC2 up to CLICS 2202-07 compliance, I think this class should include all the fields defined in that spec.

Note that some of these missing fields would provide support for a related issue: comparing the language arguments between a Shadow and a Primary CCS (e.g. the spec provides support for specifying the command-line arguments to be passed to a compiler).

src/edu/csus/ecs/pc2/core/imports/clics/CLICSProblem.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @JsonProperty
+    private Integer test_data_count;
+
+    @JsonProperty
+    private String rgb;
+
+    @JsonProperty
+    private Integer ordinal;
+
```
clevengr, line 42: I think the missing fields from the CLICS 2202-07 spec (`uuid`, `max_score`, `package`, and `statement` ) should be included in this class.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
```diff
@@ -0,0 +1,64 @@
...
+import edu.csus.ecs.pc2.core.model.JSONObjectMapper;
+
+/**
+ * 
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ */
+
+//@JsonIgnoreProperties(ignoreUnknown = true)
+public class EventFeed {
```
clevengr, line 19: The other classes in this package put "CLICS" as the first thing in the class name, which I think is a good thing since it helps make it clear that the class is defining something out of the (a) CLICS specification.  I think this class should follow the same guideline:  the class name should be something like "CLICSEventFeed".

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: For now pushed code to ignore any event where data:null rather than spend an unknown amount of
time on handling a delete.
If the delete has an impact on the comparison, that will show up in the comparison/scoreboard like any
other unhandled data.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @JsonProperty
+    private String type;
+
+    @JsonProperty
+    private String id;
+
+    @JsonProperty
+    private String op;
+
```
clevengr, line 32: "op" is not defined in the 2202-07 spec (although it was defined in earlier spec versions).  Is this field here for backward-compatibility?  (If so, it seems like there's a lot of OTHER "backward-compatibility" stuff missing...)

src/edu/csus/ecs/pc2/services/core/JSONUtilities.java
lane55: There is a bunch of refactoring to be done.   Some are marked with // TODO REFACTOR

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: When parsing and finds data:null, the EventFeed.getData() returns null.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSLanguage.java
lane55: I agree that those fields should exist, but out_ of scope  for comparing the data from the event feed.
Copied into new issue #559

src/edu/csus/ecs/pc2/core/imports/clics/CLICSProblem.java
lane55: Out of scope, added above notes into #559

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: refactoring  EventFeed to CLICSEventFeedEvent

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: No, this effort was not intended to bring all of pc2 up to spec, any spec.   To note that op should be removed
add that to #559.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: pushed refactor

src/edu/csus/ecs/pc2/core/imports/clics/CLICSContests.java
lane55: Out of scope, added as #559

src/edu/csus/ecs/pc2/AppConstants.java
lane55: pushed fix changed comment to * The value is the name of an event-feed.json file.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
johnbrvc: What @clevengr said is what I was thinking.  The feed we have does not delete anything (that I could find), at least in terms of submissions and judgements, but, that doesn't mean that Kattis doesn't.  But, as long as null is handled, that's all I was concerned about, and it sounds like it is.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSLanguage.java
clevengr: I'm not sure why you think this is "out of scope for comparing the data from the event feed".   What happens if the event feed includes those fields defined in the spec?

src/edu/csus/ecs/pc2/core/imports/clics/CLICSContests.java
clevengr: I disagree that this is "out of scope".  The tool is supposed to read a CLICS 2202-07 event feed.  This event feed could legitimately contain any or all of the fields which are currently missing in the class definition.  It seems to me that including every field that could be in the event feed. is an important component of implementing this class (and all the other similar classes about which I've made equivalent comments.

Said another way:  the event feed COULD contain those fields; what happens in the implementation if it does?

src/edu/csus/ecs/pc2/core/imports/clics/CLICSProblem.java
clevengr: Again, I disagree this is out of scope.  The tool is supposed to read the CLICS event feed; the CLICS event feed could contain those fields.  What happens in that case?

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: pushed code earlier to log when data:null

src/edu/csus/ecs/pc2/core/imports/clics/CLICSContests.java
lane55: Since those fields are not defined in the class, they are ignored.
Those fields can be added later when a review of the actual comparison output is done.
Rather than include those fields in the comparison now, the layout/content of the comparison
should be done first.    After that the fields described in #559 can be added once
the format/content has been agreed upon.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSLanguage.java
lane55: Then for this issue they are ignored.   Waiting for an agreement on the content/layout for the comparison,
after that the out of scope fields can be added.

test/edu/csus/ecs/pc2/core/imports/clics/CLICSEventFeedEventTest.java
```diff
@@ -0,0 +1,64 @@
...
+        ObjectMapper mapper = getMapper();
+        String json = "{\"type\": \"problems\", \"id\": \"542cc6e2-104a-49bc-9fea-04752f9af5ad\", \"op\": \"delete\", \"data\": null }";
+        CLICSEventFeedEvent eventFeedEntry = (CLICSEventFeedEvent) mapper.readValue(json, CLICSEventFeedEvent.class);
+        assertNull(eventFeedEntry.getData());
+
+    }
+
+    public void testTestWithData() throws Exception {
+
+        String eventLines[] = { "{\"type\": \"organizations\", \"id\": \"1add73b6-c66f-46de-a81e-a38773613394\", \"op\": \"delete\", \"data\": {\"id\": \"rsu_ac_in\"}}", //
```
johnbrvc, line 30: I suggest using sample events that 2022-07 compliant.  These samples are old format events: "op", "delete", etc. are part of the old specification.

src/edu/csus/ecs/pc2/core/report/ContestCompareModel.java
```diff
@@ -0,0 +1,821 @@
...
+
+                    CLICSProblem problem = mapper.convertValue(eventFeedEntry.getData(), CLICSProblem.class);
+
+                    // System.out.println("type " + eventType + " data = " + eventFeedEntry.getData().toString());
+
+                    feedProblems.add(problem);
+                } else if (CLICSEventType.TEAMS.toString().equals(eventType)) {
+                    // teams
+
+                    // {"type": "teams", "id": "289de609-b1a0-40d8-a9e0-54a30e752e69", "op": "create", "data": {"id": "169347", "name": "Colin Peppler", "icpc_id": null, "group_ids": [],
```
johnbrvc, line 219: This is a bad example to have in the code as it is for the old feed.  Suggest updating the comment to use a 2022-07 team event.

src/edu/csus/ecs/pc2/core/report/ContestCompareModel.java
```diff
@@ -0,0 +1,821 @@
...
+                    // languages data = {name=C, id=c}
+
+                    CLICSLanguage language = mapper.convertValue(eventFeedEntry.getData(), CLICSLanguage.class);
+
+                    feedLanguages.add(language);
+
+                } else if (CLICSEventType.JUDGEMENT_TYPES.toString().equals(eventType)) {
+                    // judgement-types
+
+                    // {"type": "judgement-types", "id": "a17b7d9d-43fc-43ed-bfd1-ef7fccbc6feb", "op": "create", "data": {"id": "RTE", "name": "Run Time Error", "solved": false, "penalty": true}}
```
johnbrvc, line 201: Use 2022-07 example in the comment, not the old feed.

johnbrvc (commented):
My opinion on adding the extra/missing CLICS 2022-07 fields is that initially, we should just have the fields that are in the event feed we are using for WF, in this case, the DOMJudge event feed.  Any field in that feed we should probably have the field in the class.  Any fields that is in the spec, but DOMJudge does not use, can be delayed until issue #559.  This has to work for WF (or at the very least, BAPC in 2 weeks), so in the interest of time...

src/edu/csus/ecs/pc2/core/imports/clics/CLICSProblem.java
lane55: This PR/change is already 2,500 lines and 23 classes - the missing work is covered by #559.

src/edu/csus/ecs/pc2/core/imports/clics/EventFeed.java
lane55: Since op is not harming anything, if that should be removed - add a new Issue.
Working to not have a single issue that "does it all".

test/edu/csus/ecs/pc2/core/imports/clics/CLICSEventFeedEventTest.java
lane55: Sure, that is also out of scope.  Being compliant with any spec was not in scope for this issue, it  is
to provide a comparison between the event feed and the pc2 model - as such providing a framework
and for future changes, including processing/handling 2022-07 spec.

lane55:
> so in the interest of time...
great, there is now a new issue that depends on this PR.

src/edu/csus/ecs/pc2/core/report/ContestCompareModel.java
lane55: It's what I had 4 weeks ago when I started.   I expected to make these sort of updates nearly 2 weeks ago when the PR was created.   But I have not attempted to track/analyse the content because that can be easily updated.  It was work to get the framework and content/layout done.

src/edu/csus/ecs/pc2/core/report/ContestCompareModel.java
lane55: Noted.  No change at this time.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSEventType.java
lane55: pushed fix, 3 days ago.

test/edu/csus/ecs/pc2/core/imports/clics/CLICSEventFeedEventTest.java
lane55: The samples were what I had 4 weeks ago when I started coding, part of #559 may be to update comments..

johnbrvc (approved):
I believe this should meet the requirements we have for WF Dhaka.  #559 can be handled after that, IMHO.

lane55:
All code/logic issues fixed/pushed.

Issue #429, opened 2022-06-02
If shadow connection is lost to primary, an auto-reconnect should optionally be attempted


johnbrvc:
**Describe the issue**: 
If there is a network disruption, it is possible for the shadow to lose its connection to the primary.  In this case, it should be possible to have the shadow automatically attempt to reconnect to the primary.   This can be a check-box option on the Shadow tab.  See also Issue #422 

Add information on the Shadow tab: the last connection date/time to the Primary CCS and some indication whether it is a connection (via Start Shadowing) or an automatic reconnection.
Naturally when a connection is dropped that should be logged, and when reconnected should be logged.
There should be a constant defined for how long in ms before the Shadow reconnects.

**To Reproduce**: 
Start shadowing to a primary.  Disrupt network connection (pull ethernet cable? change IP address?).  Restore network connection after a few minutes.  Notice exception in Feeder log.  No reconnect attempt is made and error is silently ignored.

**Expected behavior**: 
If (a new) "Reconnect on lost connection" checkbox is checked, the Feeder should attempt to reconnect to the primary to restore the connection. 

**Actual behavior**:
The lost connection is silently ignored an no further shadow updates occur.

**Environment**: please complete the following information:
 All

**Log Info**:  
220530 140811.764|SEVERE|RemoteEventFeedMonitorThread|run|Exception reading event from stream:
|java.io.IOException: Premature EOF
| at sun.net.www.http.ChunkedInputStream.readAheadBlocking (ChunkedInputStream.java:568)
| at sun.net.www.http.ChunkedInputStream.readAhead (ChunkedInputStream.java:612)
| at sun.net.www.http.ChunkedInputStream.read (ChunkedInputStream.java:699)
| at java.io.FilterInputStream.read (FilterInputStream.java:133)
| at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read (HttpURLConnection.java:3510)
| at java.io.BufferedInputStream.read1 (BufferedInputStream.java:290)
| at java.io.BufferedInputStream.read (BufferedInputStream.java:351)
| at sun.nio.cs.StreamDecoder.readBytes (StreamDecoder.java:284)
| at sun.nio.cs.StreamDecoder.implRead (StreamDecoder.java:326)
| at sun.nio.cs.StreamDecoder.read (StreamDecoder.java:178)
| at java.io.InputStreamReader.read (InputStreamReader.java:181)
| at java.io.BufferedReader.fill (BufferedReader.java:161)
[...]

PR #575, opened 2022-10-24
Part 2 Shadow connection logging expands on fixing #429

johnbrvc:
Adds a JTable to the Shadow Mode tab (ShadowControlPane) of the Feeder.  This JTable will show every connect, disconnect and socket error disconnect, as well as other connection related information, along with the time of the occurrence.

### Issue which the PR fixes
Expands on #429 

Java Version    : 1.8.0_321

1) Start a PC2 server; load a CDP for a contest which is configured the same as that provided by the "Primary CCS" (below).
2) Start the feeder1 client
3) Click on the Shadow Mode tab
4) Enter a valid Primary CCS URL (that is currently up and available - you can use any one that is available. It makes no sense to put a specific one in here since it's probably not valid anymore).
5) Provide the user and password information
6) Press the "Test Connection" button.
7) Notice the messages that are shown in the JTable.  If the remote CCS provides an API version (and provider), these will be shown in the JTable as a separate entry.
ShadowModeScreen
8) Press the "Start shadowing" button.
9) You should observe an entry added to the table indicating the disposition of the connect attempt.

Here is a list of all the possible messages that can appear in the table along with a brief description.  The code may be more revealing if further details are desired.

**Test connection to remote CCS successful** - after the test button was pressed, this indicates the test succeeded
**Test connection to remote CCS failed** - after the test button was pressed, this indicates the test did not succeed
**API Version: YYYY-## (unknown provider)** - The API EP returned a valid response, but the optional "name" was provided
**API Version: YYYY-## by (DOMJudge)** - The API EP returned a valid response and provider
**API Version: N/A .....** - The API EP returned JSON, but the (required) version was not included.  This is a server problem.
**No API version available at https://URL** - The web fetch of the EP failed
**No API Version URL available** - The URL specified does not appear to point to a valid CCS server that follows any known spec.
(all the above messages are logged as "info" messages as well)

**Connection failed** - if Start shadowing was pressed and no starting token was specified, and the connection to the primary failed
**Connection failed starting at token TTTTTT** - if Start shadowing was pressed and a starting token (TTTTTT) was specified, and the connection to the primary failed
(the above two messages also cause log file messages to be generated)

**Connected successfully** - if Start shadowing was pressed and no starting token was specified, and the connection to the primary was successful
**Connected successfully starting at token TTTTTT**- if Start shadowing was pressed and a starting token was specified, and the connection to the primary was successful
**Connection closed by PC2** - when the Stop shadowing button is pressed and the connection to the primary is closed
(No log messages are generated for the above 3)

**Unexpected IO exception on Event Feed (record count and/or token)** - If a socket exception occurs after a successful connection is made.
(A log message is generated giving details)

Note: I informed the DOMJudge team that their API endpoint does not currently provide the "provider" information in the returned JSON (ie. the optional "name" property is not present, which is where the provider comes from).


johnbrvc:
This PR would be nice to have, but not required for WF.

lane55:
This seems to be missing examples for testing.   What are examples of the messages added to the JLIst

lane55:
Added requirements and questions into #429

lane55:
John,

                How would this work on a read-only feeder (feeder 2 – N)?

Doug

johnbrvc:
> 

In my testing, it works as expected.  It shows connection results for the read-only feeder.  Why wouldn't it work?

johnbrvc:
> This seems to be missing examples for testing. What are examples of the messages added to the JLIst

I itemized all the messages that could appear in the table, and added another example of how to test.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -713,8 +804,57 @@ public void run() {
                                     boolean isConnected = remoteContestAPIAdapter.testConnection();
                                     if (isConnected) {
                                         showMessage ("Connection to remote CCS is successful");
+                                        addConnectTableEntry("Test connection to remote CCS successful");
                                     } else {
                                         showErrorMessage("Connection to remote CCS failed", "Connection failed");
+                                        addConnectTableEntry("Test connection to remote CCS failed");
+                                    }
+                                    
+                                    // API endpoint is right before /contests/ in the URL, so find that, if it's there
```
lane55, line 813: Does this directly involve the connection logging?
The block staring  // API endpoint is right before /contests/ in the URL, so find that, if it's there
This block is too long for the method, it should be put into its own method, also this likely will
provide a way to unit test this block.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -736,6 +876,58 @@ public void run() {
...
+                    objects[0] = lastDateFormat.format(cal.getTime());
+                } catch(Exception e) {
+                    objects[0] = "Unknown";
+                }
+                if(msg == null || msg.isEmpty()) {
+                    objects[1] = "<Empty Message>";
+                } else {
+                    objects[1] = msg;
+                }
+                connectStatusTableModel.addRow(objects);
```
lane55, line 925: Rather than add the row, the row should be inserted from the top.  
Unless, as entries are added, the scrolling reveals the last connection logged.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -785,23 +977,51 @@ public void updateShadowNumberofRecords(int nRec)
...
+        }
+        addConnectTableEntry(msg);
     }
     
     /**
      * {@inheritDoc}
      */
     public void errorDisconnect(String errMsg)
     {
-        // TODO: Update JTable - future commit/PR
```
lane55, line 804: What is this todo?   there should be some sort of summary description of the next task
or a reference to a github issue comment.

lane55 (changes requested):
A question or two
and more information added/needed.for the TODO comments

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
clevengr: I agree with Doug; IMHO each entry should either be added to the TOP of the list (so, it's a Stack), or else adding an entry should force the scrollpane to focus on the bottom (newest) entry.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
clevengr: @lane55 : I don't understand your comment on this line.  The line is a "deleted" line; it's an old "TODO" that has been "ToDONE" with the following few lines, just like the two previous "TODO" comments which have been removed and replaced by code implementing the "TODO"....

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -503,6 +546,54 @@ private ShadowInformation getCurrentShadowInformation(ContestInformation contest
...
+         * Column headers left justified
+         */
+        ((DefaultTableCellRenderer)connectStatusTable.getTableHeader().getDefaultRenderer()).setHorizontalAlignment(JLabel.LEFT);
+        connectStatusTable.setRowHeight(connectStatusTable.getRowHeight() + VERT_PAD);
+                     
+        int idx = 0;
+
+        // These are in sort order
+        // Time
+        sortList.add(new RowSorter.SortKey(idx++, SortOrder.ASCENDING));
```
clevengr, line 581: The ability to sort the table rows by time is very useful.  However, after testing it I found it confusing because when you click on the column header to reverse the sort order, the vertical scrollbar does not also adjust.  For example, if you have multiple rows (enough to cause a Scrollbar to appear) and you click the column so that the sort order is "lowest time first", the time which appears at the top of the view is _not necessarily the lowest time_  -- it's only the lowest time _within the view indicated by the current position of the scrollbar_.  It seems to me that it would be better if, when the user clicks to "sort lowest to highest" the scrollbar would also adjust (to the top) so that it's clear that the top row _in the view_ is indeed the "lowest"  (and similarly for sorting "highest to lowest" -- I would expect the view in that case to show the highest value at the top).

I grant that other people may have a different preference (may not want the scrollbar to change when a sort is invoked), but it sure seems confusing to me as it is currently.

clevengr (approved):
I've reviewed the code changes in this PR, and I've run the code connecting to a Primary CCS (DOMjudge).  I've observed multiple different connection messages (including both "normal" and "error conditions") displayed in the new "status pane".  As far as I can see the code works as intended.  I've made comments/suggestions on minor changes, including:
- concerns regarding table row sorting (clicking header) vs. scrollbar positioning.
- insertion of new messages at the top of the "stack".
- scrollbar resetting when new messages arrive.

Although I'd like to see these changes made, none of these rise to the level of being important enough to force a request for changes, so I approve the PR.

johnbrvc:
> 

I am going to review the suggestions and make changes as required.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
johnbrvc: Pretty much every table we have operates this way (take Accounts and Runs as an example).  I'm not saying it's correct, but, as you indicate it's a preference thing.  The "Windows" solution is to always jump to the "top" after any sort (in explorer) no matter where you were (if no item is selected.)   If an item is selected, it "makes sure that item is visible" after the sort.   We can do that (if an item is selected, make sure it's visible after the sort), but (and I cringe a little to say this),that's out of scope for this issue and if we do that, ALL tables should be changed to work the same way, which, we can probably do somewhat easily since all JTable's use JTableCustomized as a base class...

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
johnbrvc: It was a tiny bit more complicated than either of those two solutions.  I opted for the "adjustmentlistener", but that didn't work exactly right either (you couldn't scroll back manually) - so I fixed that, and it now seems to always add to the end, and keep the end visible.  The problem with simply scrolling the scroll bar to the max, was that it would always miss the most recent row, and not show the LAST entry until the next row came in (people had various solutions to this, but the listener seems to most reasonable).

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
clevengr: I agree that we can consider this "out of scope".

clevengr (approved):
I reviewed the PR again with respect to the changes to make the JTable scrollbar always move to the bottom (latest entry).  Seems to work great; I (again) approved the PR.

johnbrvc:
Based on feedback from @lane55 the shadow status JTable was changed to have a Status column giving a general indication of success, failure or info (color coded appropriately).  In addition, made the JTable expand as the window is resized so you can see more rows as the windows grows.  Re-reviews have been requested.
image

clevengr (approved):
I reviewed the updated code and it looks reasonable.  I pulled the code from the PR repo and executed it against the DOMjudge BAPC22 contest, observing both "success" and "failure" messages as appropriate.  Nice addition; I approve the PR.

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
johnbrvc: This change has been made and the code has been moved into separate methods.

lane55 (approved):
Approved.

PR #578, opened 2022-10-28
I 383 add awards json export report

### LINKED ISSUE #383: CLICS Add awards.json file/report

The awards.json is used for contest results validation and it part of the CLICS spec.

**Feature Description**: 

Add awards.json file/report

Implement per requirements in https://ccs-specs.icpc.io/2022-07/contest_api#awards 

There are none at this time.

Create a report that will output award(s) .json.

**Additional context**:

lane55:

Creates a awards.json file, via a new report "awards"

### Issue which the PR fixes

#383

java version "1.8.0_121"

Load sample contest: tenprobs

Start contest
Submit and judge runs Yes for the following teams/problems.

   Team ##  Problem
   team 22  A
   team 51  B
   team 29  C, A
   team 32  D
   team 38  F
   team 39  G

Select and view report "awards JSON Report", expected results, note that the citation order may
not be sorted.   The runs can be submitted one per minute to produce the results below.

```
[{"citation":"Contest winner","id":"winner","team_ids":["29"]},
{"citation":"First to solve problem A","id":"first-to-solve-a","team_ids":["22"]},
{"citation":"First to solve problem F","id":"first-to-solve-f","team_ids":["38"]},
{"citation":"First to solve problem C","id":"first-to-solve-c","team_ids":["29"]},
{"citation":"First to solve problem B","id":"first-to-solve-b","team_ids":["51"]},
{"citation":"First to solve problem D","id":"first-to-solve-d","team_ids":["32"]},
{"citation":"First to solve problem G","id":"first-to-solve-g","team_ids":["39"]},
{"citation":"Winner(s) of group North","id":"group-winner-100","team_ids":["29"]},
{"citation":"Winner(s) of group South","id":"group-winner-200","team_ids":["51"]},
{"citation":"Gold medal winner","id":"gold-medal","team_ids":["29","22","51","32"]},
{"citation":"Silver medal winner","id":"silver-medal","team_ids":["38","39"]}]


```


lane55:
The requirements question is now, do we add to every report a default report name in place of using
a combo of the report name and a "time stamp" ?

johnbrvc:
> 

I am thinking for now, just fix the one(s) that were added to include the "json" string so the file gets appropriately named, since there is already a mechanism in place in the code to do this.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -110,22 +117,30 @@ public static void addGroupWinners(IInternalContest contest, Run[] runs, List<CL
             return;
         }
 
+        
         /**
          * Group Id and Client/Team Id
          */
         Map<Group, ClientId> groupWinners = new HashMap<Group, ClientId>();
+        try {
```
johnbrvc, line 125: Perhaps add a comment describing what this block of code is supposed to be doing.

johnbrvc (approved):
Suggested code change below should not block approval.

lane55:
New awards json reports are written as:
reports\report.awards_JSON.06.19.198.json
reports\report.awards_JSON.06.19.442.json
reports\report.awards_JSON.06.19.965.json

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
lane55: Pushed updated commnts about groupWinners map and how group winners are 
added to the map.

johnbrvc (approved):
Reviewed updated comments.
Rebuilt and tested (since stuff was merged too).
Works Ok.   Changes since my previous approval were minimal with no functional changes.

clevengr:
The "Precise steps for testing the PR" are incomplete/insufficient.  They tell me to 
- Load sample contest: tenprobs
- Start contest
- Submit and judge runs Yes for different teams/groups

Then the steps contain some text (a list of problems and what appears to be a file dump), but it doesn't tell me what those things are or what I should be looking for to verify that the PR does what it is supposed to do.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
```diff
@@ -0,0 +1,96 @@
...
+    public CLICSAward() {
+        super();
+    }
+    /**
+     * Award with teams.
+     * @param id  identifier for award, ex winner
+     * @param citation ex. Contest winner
+     * @param team_ids lsit of ids
+     */
+    public CLICSAward(String id, String citation, List<String> team_ids) {
```
clevengr, line 55: I'm mildly surprised that there is no constructor accepting `String []` for team_ids.   (See the subsequent comment below.)

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public String getId() {
+        return id;
+    }
+
+    public String getCitation() {
+        return citation;
+    }
+
+    public String[] getTeam_ids() {
```
clevengr, line 84: The return type of this method seems inconsistent with the structure of the constructors.  Specifically, as pointed out above, there is a constructor that accepts `List<String>` for team_ids, but no corresponding constructor accepting `String []`.  That actually makes sense; external clients (callers) shouldn't know the internal storage type of elements (in this case, the fact that team_ids are stored as an _array_ of Strings).  However, the return type of this method defeats that; it exposes the fact that team_ids are stored as arrays of Strings.  It seems like the return type of this method should be `List<String>`, for consistency with the constructors and to avoid exposing the internal (array) storage type.

lane55:
> The "Precise steps for testing the PR" are incomplete/insufficient. They tell me to

Updated PR testing description to include report name.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
lane55: pushed fix to constructor

lane55:
> . I'm left having to guess the meaning of that additional text and what I'm supposed to do to verify the PR does what it is supposed to do.

Precise steps for testing the PR - were updated last week, awaiting aproval or feedback.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
```diff
@@ -52,11 +50,11 @@ public CLICSAward() {
      * @param citation ex. Contest winner
      * @param team_ids lsit of ids
      */
-    public CLICSAward(String id, String citation, List<String> team_ids) {
+    public CLICSAward(String id, String citation, String [] team_ids) {
```
clevengr, line 53: This change, which apparently was made in response to my previous review comment, is in effect the exact _opposite_ of what I suggested/requested.  Previously, the constructor accepted a `List<String>` for `team_ids`.  As I said in my previous comment, this seems correct; external clients should know only that they are to provide a `List` and should not know how that `List` is to be stored externally. The problem I pointed out previously was in the method `getTeamIds()`, which was (and still is) returning type `String[]`.  This means there are now TWO places where the internal storage type (specifically, an _array_) is exposed to external clients.

The request was to change the return type of method `getTeamIds()` to be `List<String>`.  This commit does the opposite of that.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
```diff
@@ -52,11 +50,11 @@ public CLICSAward() {
...
         super();
         this.id = id;
         this.citation = citation;
-        this.team_ids = (String[]) team_ids.toArray(new String[team_ids.size()]);
+        this.team_ids = team_ids;
```
clevengr, line 57: This change should be reverted as part of fixing the issue referred to in my previous comments.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -278,8 +278,8 @@ public static List<String> getTeamIdsByRank(List<TeamScoreRow> scoreRows, int lo
                 teamIds.add(teamScoreRow.getTeam_id() + "");
             }
         }
-
-        return teamIds;
+        
+        return (String[]) teamIds.toArray(new String[teamIds.size()]);
```
clevengr, line 282: Related to my comments on class `CLICSAward`, this line of code exposes the internal storage type (`String[]`) to external clients.  External clients should know only that they are getting a `List`.

Note that reverting this line also requires reverting all the above changed lines (back to what they were).

src/edu/csus/ecs/pc2/core/report/CLICSAwardsReport.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Print CLICS Awards JSON
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class CLICSAwardsReport implements IReport {
+
+    private static final long serialVersionUID = 6270072776103982061L;
```
clevengr, line 27: This is just a comment, not a request for a change... but:  I wish someone would explain to me what sense there is in assigning a serialVersionUID like 6270072776103982061.   Random sVUID values seem pointless to me; I cannot think of a single instance where a human reading the code could make sense of them.   It seems much more logical to me to assign a sVUID of "1" to the first version, and then increment it each time a breaking change occurs in the class.

test/edu/csus/ecs/pc2/core/report/ContestCompareModelTest.java
```diff
@@ -195,12 +195,16 @@ public void testReport() throws Exception {
             // only test if there is a pretest dir found
```
clevengr, line 195: Three lines above this line there is a definition of a path which contains a hard-coded value of `c:\\repos...`.  (I'm commenting on that line here because for some reason GitHub is not letting me attach a comment to that line above.)

I cannot think of any valid reason why there should be a Windows-specific path (starting with "c:") in the code.  This will especially break because this is in the `test` suite, which will ultimately be run on the GitHub Linux build machine.

test/edu/csus/ecs/pc2/core/report/ContestCompareModelTest.java
```diff
@@ -195,12 +195,16 @@ public void testReport() throws Exception {
...
             return;
         }
+        // Load Accounts
+        String loadAccountFilename = "/repos/icpc/ccsconfig-dhaka/contests/pretest2/config/load_accounts_real.tsv";
```
clevengr, line 199: I do not understand why there is an assumption made that a file from Dhaka exists on the machine where the tests are being run.  Even if such a file/path does exist on the author's machine, this test is going to fail if run on somebody else's machine unless the specified file is committed to the master repository (and I don't see any such thing anywhere...)
Please either remove this, or explain why it is important and how it is supposed to work.

test/edu/csus/ecs/pc2/core/util/AbstractTestCase.java
```diff
@@ -792,6 +792,22 @@ public void editFile(String filename) {
         }
     }
     
+    public void editFile(String filename, String comment) {
```
clevengr, line 795: This adds a method which is never called from anywhere in the project.  What's the point in that?

test/edu/csus/ecs/pc2/core/util/AbstractTestCase.java
```diff
@@ -792,6 +792,22 @@ public void editFile(String filename) {
...
+        
+        System.err.println("editFile: "+comment+" file: "+filename);
```
clevengr, line 797: I thought we agreed that Test methods shouldn't output to stdout/stderr?   (Certainly we have a LOT of places that already do... but why are we adding more?)

test/edu/csus/ecs/pc2/core/util/AbstractTestCase.java
```diff
@@ -792,6 +792,22 @@ public void editFile(String filename) {
...
+
+        String viFile = findViExecutable();
```
clevengr, line 799: Unless I missed something, from this point onward to the end of this (new) method, the code in this method is identical to the existing method `editFile(String filename)`.  Why add a duplicate set of code instead of just calling that existing method?

clevengr (changes requested):
I have reviewed all code changes in this PR, and made some requests for changes (or at least, explanations).  I'm now proceeding to perform an execution test.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
johnbrvc: As of the latest commit, the return type of getTeamIds() and the argument to constructor are the same String [].  I think this is fine and consistent.  IMHO I don't think should block approval.

src/edu/csus/ecs/pc2/core/report/CLICSAwardsReport.java
johnbrvc: https://docs.oracle.com/javase/6/docs/platform/serialization/spec/class.html#4100

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
clevengr: I again fall back to my teaching days, where among other things I tried to pound into my student's heads the notion that a class should not expose to outside clients the types which it uses for internal storage.  For example, a class `Stack` should not allow external users of a `Stack` to know whether the Stack is implemented as an array, a linked list, a hashtable, or using some other data type.  The reason for this is because if clients know the internal data type, they can make code assumptions based on that type; if the original class later decides to change the storage type (for efficiency reasons, for example), that will _break external clients_.  Making internal changes to a class shouldn't do that.

In this case, the use of `String []` parameters and return types is allowing external clients to know that these values are (always) stored as an _array of Strings_.  If we later decide that it would be better to store things internally as (say) a `Vector<String>`, making such a change will break external clients.

This is the whole purpose of the `List<E>` interface in Java -- it allows specification of an ordered list of things without exposing the internal implementation of the list; this in turn makes for more robust code.

I stand by my original request that the constructors and the accessor methods use `List<String>` instead of `String[]` as their parameters and return types.   

Yes, I'm being pedantic -- but with good reason in this case IMHO.

src/edu/csus/ecs/pc2/core/report/CLICSAwardsReport.java
clevengr: Thanks for that reference.  I do now at least see why the use of a long auto-generated sVUID is useful -- it encapsulates into the sVUID a number of characteristics of the class, rather than simply being a "version number".

What I still _don't_ see is how it is useful _in our situation_.  Once the sVUID is auto-generated, _it remains the same no matter what changes are made to the class_ until/unless it is regenerated (which I seriously doubt anyone is ever doing).  Further, the whole point is supposed to be to allow run-time detection of an attempt to use an incompatible class version -- which can only be detected if one first defines/decides/declares what is "compatible" or "incompatible".   This in turn requires a lot of subtle analysis of the potential effects of changes to the class -- again, something I seriously doubt any of our developers put much if any time into doing.

Can anyone point to even one time where they have regenerated the hard-coded sVUID in a class because they've made backwards-incompatible changes?   (Note that Eclipse doesn't regenerate those numbers automatically; you have to explicitly tell it to do that.  I'm basically arguing that doing that is more or less the same as simply bumping a serial number up by one when an incompatible change is made.)

Note also:  I don't consider this a blocking-approval item; I'm just trying to point out where I think there are weaknesses in our process... although I grant that this particular one is pretty far down the list compared to others we have :)

clevengr (changes requested):
I have marked certain comments "resolved", and left additional (non-approval-blocking) comments in other places.  I'm still waiting for responses to the other comments I left earlier.

I have also now completed an execution test of the code.  It fails the "Steps to verify the PR works".  

Specifically, I followed the specified steps, loading the `tenprobs` contest and then submitting the specified runs for the specified teams, and then looked at the Awards JSON report.  It does not match the "expected result" shown in the PR, in numerous regards:

- The format of the actual report elements (the individual JSON array elements) is not the same, in either form or content, as what is given in the "expected results".  For example, the "expected results" tells me to expect an element like:
```
["edu.csus.ecs.pc2.core.imports.clics.CLICSAward",{"id":"Contest winner","citation":"winner","team_ids":["java.util.ArrayList",["29"]]}]
```
whereas the report I get shows me an (apparently corresponding) element like
```
{"citation":"Contest winner","id":"winner","team_ids":["29"]}
```
Notice that, in addition to the removal of the Java "type" information (which is A Good Thing -- but still it doesn't match the "expected results" in the PR), there are important other differences:
  - the contents of the "citation" element and the "id" element are reversed between the "expected output" and the actual output
  - the "expected output" says that each JSON array element should consist of an array containing two elements: a type ("CLICSAward") and a "value"  (a list of elements including "citation", "id", and "team_ids") -- but the actual output has each element consisting of just a list (containing elements "citation", "id", and "team_ids"), not an array.
  - The "expected output" tells me to expect an element of the form
```
{"id":"Winner(s) of group Long22","citation":"group-winner-100","team_ids":["java.util.ArrayList",["22"]]}
```
but the actual output shows an (apparently corresponding) element like
```
{"citation":"Winner(s) of group North","id":"group-winner-100","team_ids":["29"]},
```
(Notice that not only are the values of the "citation" and "id" fields reversed, but they're also not even the same.)

- The actual output contains an array element of the form
```
{"citation":"Gold medal winner","id":"gold-medal","team_ids":["29","22","51","32"]}
```
but the "expected output" makes no mention of any such element.

I have a suspicion that the issue may be that the "Steps to demonstrate the PR works" may have never been updated as continual changes to the code have been made.  However, I can't just assume that; I have to test what I'm given -- and what I'm given doesn't pass.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
johnbrvc: Yeah, I think this is one of those "religious" things... I think we can agree to disagree here.  By using `String[]`, this (IMO) does not conveying any information about how the information is stored in the object.  If, for example, `private String [] team_ids`; was `public String [] team_ids`; then yes, this does expose the internals.  Just because an array of strings is passed in the constructor and returned as an array in what is presumably an accessor, does not necessarily mean it is stored or used that way in the `CLICSAwards` object, right?  Also, if, for some reason, the data that is passed in or returned changes from something OTHER than a "bunch of" of `String` objects, client code will have to be modified _anyway_ to accommodate the new type of object.  `List<String>`  or `String[] `or `Vector<String>`, any of those work for _me_.

lane55:
Re: different results

lane55:
Found a bug where the are a bunch of Silver medal winners when the team solved no problems.

lane55:
Updated the expected results which match DOMjudge JSON, tested in NAC23

Pushed fix for silver medals, for sample/test there should be no silver medalists.

johnbrvc (approved):

johnbrvc:
Retested latest commits.  Works as expected. Tested with NAC2023 contest.

lane55:
Reviewed PR for fixes that block the approval.  I do not see any new comments/corrections since
the last push to the branch.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
clevengr: Ok, I'll say two things.

One:  I still think I'm right and it's more than a "religious thing".  There is a good reason the designers of Java added the `Collection` interface to the language, and added `List` which inherits from (extends) collection, and added `ArrayList` (which implements `List`):  it was to support the ability to create abstractions in the types which are passed between classes.  I think that applies here.

Two:  people can have different opinions on stuff like this, and it's not worth spending more time on in this case.   Accordingly, I withdraw my request that the code be changed, and I will mark this conversation (and the related ones) "resolved" so we can move on.

clevengr (changes requested):
I have re-reviewed the code.  I've marked as "Resolved" several comments I made earlier.  There are still several active comments which have not been addressed -- regarding methods which are never invoked, methods which contain code which duplicates code in existing methods, etc.  While none of the remaining comments rise to the point of my not being willing to approve the PR, I don't care for the fact that we are adding code to the system which is (IMHO) not properly structured; I'd sure like to see these things fixed before merging (because we all know they're never going to get fixed AFTER merging...thus just continuing the code bloat problems).  In any case, they aren't stopping me from Approving the PR.

What IS stopping me from Approving is that _the "Expected Output" as listed in the PR still does not match the actual output I get_ (and this is taking into account that the "Expected Output" was recently updated).

Specifically, there are two issues with the Expected vs. Actual Output:

- the order of Award elements in the actual JSON output is not the same as the order given in the Expected Output.   Specifically, the Expected Output says to expect award elements in the following order:
  - first-to-solve-g
  - first-to-solve-a
  - first-to-solve-f
  - first-to-solve-c
  - first-to-solve-b
  - first-to-solve-d
whereas the Actual Output produces these awards in the order
  - first-to-solve-a
  - first-to-solve-b
  - first-to-solve-d
  - first-to-solve-f
  - first-to-solve-g
  - first-to-solve-c

Since the CLICS specification is silent on what the order of award elements within the JSON should be, I suppose it's reasonable to argue that the above are "equivalent", and I could live with this difference.  However, it certainly makes it harder to evaluate whether the PR is working as intended when the actual output does not match the stated (expected) output.  But it's not stopping me from approving the PR.

- The Expected Output says to expect the JSON to contain:
```
{"citation":"Gold medal winner","id":"gold-medal","team_ids":["29","22","32","38","39","51"]}
```
 However, the actual output instead contains:
```
  {"citation":"Gold medal winner","id":"gold-medal","team_ids":["29","22","51","32"]},
  {"citation":"Silver medal winner","id":"silver-medal","team_ids":["38","39"]}
```
Since these are _clearly_ not the same, I can't approve the PR.  Please either fix the Expected Output listed in the PR, or fix the code so it produces the correct output.

lane55:
Consider next time reviewing the output compared to the contest model/awards.

lane55:
Re-running sample with latest code, the bug fix to not include teams who did not solve a problem

lane55:
Pull and re-test, just pullsed and re-tested and matched the expected output/content

lane55:
Until now there was no specification about how fast the runs should be submitted.
Original expected output results were based on all runs being submitted at the same elapsed time.

Updaetd expected output based on pacing the submission of runs to one run per min.
This matches John C's test results.

lane55:
Added note in description about pacing of the submission of the runs.

clevengr (commented):
I did pull the latest code prior to testing.  As Doug has subsequently noted, the major difference between my output and the "Expected Output" shown in the PR was probably due to the fact that I submitted the specified runs over time.  My output now matches what Doug got (and entered in the PR as new "Expected Output").

I still have a bit of a question/concern regarding the other issue (difference) I noted:  the output which I get lists the "First To Solve" awards _in a different order than what is shown in the Expected Output_.  Since we're (theoretically) running the same code, and since I presume that the "Expected Output" in the PR was copied there from an actual execution, it concerns me that we would get output which differs (syntactically if not semantically).   Doug:  can you offer any suggestions as to why this might happen?  (I don't like haven't "unexplained differences" in the code/results...)

lane55:
>  Doug: can you offer any suggestions as to why this might happen? (I don't like haven't "unexplained differences" in the code/results...)

clevengr (approved):
I have completed my review and am approving the PR.  However, I'd still like to point out that there are several places where I've made comments that haven't been addressed (or even answered):
- JUnit Test methods that write to stdout/stderr
- methods whose code duplicates that in other methods

I don't feel I should withhold approval of the PR since we need to get all these "Next Contest" Issues resolved.  However, I sure hate to see us adding code to the code base which contains things like this (I suspect if this code came from an external contributor, we probably wouldn't be willing to accept things like this...)

lane55:
Two approvals, merging

Issue #584, opened 2022-11-06
Shadow Run Compare missing pending count


lane55:
**Is your feature request related to a problem?** 

On Shadow Run Compare does not show Pending count with
other counts.

**Feature Description**: 

Add Pending: NNN 
to show the number of pending runs. on the same line as the Total Submissions, Matches, etc. counts

**Have you considered other ways to accomplish the same thing?** 

n/a

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?**

**Additional context**:

clevengr:
The code already contains a display of Pending Runs, in method `updateSummary()` of class `ShadowCompareRunsSummaryPane`  (around lines 157-162).   The problem is that this field is not always displayed (I've seen it displayed; I think on a Windows system); I've seen it NOT displayed, on a Linux system.  I think therefore that this is a Layout Manager issue.

In particular, the `ShadowCompareRunsSummaryPane` (which extends JPanel) has a constructor which makes the following call:  `this.setMaximumSize(new Dimension(600,40));`.  This JPanel component is placed inside another JPanel with a maximum size of 700,40.  I think the "Pending" label is probably extending beyond the 600 limit and so the FlowLayout manager is trying to move the Pending label/count to the next line -- but it won't be visible.  I think a fix would be to change the setMaximumSize(600,40) to setMaximumSize(700,40).

PR #617, opened 2022-11-16
i612: expand size of ShadowCompareScoreboardSummaryPane.

### LINKED ISSUE #612: On Shadow Compare scoreboard, sometimes when scoreboard differs the Contest Status is blank
**Describe the issue**: 

On Shadow Compare scoreboard, sometimes when scoreboard differs
the Contest Status is blank

**To Reproduce**: 

**Expected behavior**: 

**Actual behavior**: 

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

clevengr:

Expands the size of the CompareScoreboardSummaryPane.  Previously, the pane had a maximum width of 500.  If the "Scoreboards do not match (xx of yy rows mismatched)" message is too large to fit in this size, it disappears (it gets moved down to the next row by the Flowlayout manager, but the next row isn't visible).   This is similar to the problem which caused the "Pending" field on the RunComparison screen to disappear at certain times -- Issue #584 ).  

This PR expands the SummaryPane width to 800.  The CompareScoreboardSummaryPane is placed in a `ShadowCompareScoreboardPane` which is a `JPanePlugin` (which is a `JPanel`) of width >1000, so expanding the size of the SummaryPane should be fine and should allow larger message strings to display properly.

### Issue which the PR fixes

Fixes #612 

Windows 10.1, Java 1.8.0_271

Run a PC2 Shadow against a Primary CCS running a contest which has many teams (>100) and for which many scoreboard mismatches occur (for example, during the initial phase of the contest when the Primary CCS has judged many runs but the PC2 Shadow is lagging behind).   (An example contest is the ICPC WF Dhaka contest).  Previously, a lengthy message like "Scoreboard do NOT match (xxx of yyy rows mismatched)" could disappear under certain GUI environments (e.g. Ubuntu 20.04).  Now the panel should display properly.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardSummaryPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ShadowCompareScoreboardSummaryPane.java
@@ -33,7 +33,7 @@ public class ShadowCompareScoreboardSummaryPane extends JPanel {
      */
     public ShadowCompareScoreboardSummaryPane() {
         
-        setMaximumSize(new Dimension(500, 100));
+        setMaximumSize(new Dimension(800, 100));
 
 //        JLabel header = new JLabel("Comparison of PC2 vs. Remote Scoreboard");
 //        header.setAlignmentX(Component.CENTER_ALIGNMENT);


lane55 (approved):
Approved

PR #618, opened 2022-11-23
i244: now by default loads sample/ data files before secret

### LINKED ISSUE #244: Provide a way to add sample data files to the test data files

When shadowing there is a step required to create test cases data files.
The sample files should be tested before the judge's data files (in secret/)

Here is the workaround:

Sample and Secret data files need to be renamed and 
after renaming the sample files are copied to the secret directory.

Merge the “sample data” into the “secret data” folder and load ALL data
files into PC2 (Kattis runs against both the samples and the secret data).
There is a Python3 script in the NAC GitHub repo that can be used to do
this;  it takes a single argument -- the name of the “cdp” folder which
contains the config folder which contains the problem package (problem
folders and a single “problem.yaml” file).  Warning:  the script MOVES
the sample data into the secret data folder, prepending “sample__”
to all sample data files and “secret__” to all secret data files.
This produces a single folder, “secret”, containing all the data files in
proper alphabetical order -- but it destroys the original cdp structure.

Explanatory note:  data files are required to be run in alphabetical
order, and Kattis runs all sample files before any secret files

Source: NADC 2021 PC2 Shadow Configuration Verification Checklist

**Feature Description**: 

Using --load (from yaml) automatically load sample files before loading secret files for each problem.

For work arround:

Copy sample files to data/secret directory, rename all data files so that the sample files will
be used before the secret files.

Copy each .in and .ans file from sample data directory  (config\<shortdirname>\data\sample) to secret directory
(config\<shortdirname>\data\secret) with a new name, the new file name should be the sample file name
prepended with sample__ .   Ex. data/sample/sumit.in copy to data/secret/sample__sumit.in

Rename all secret files, prepend secret_ to each secret__, ex data/secret/sumit4.ans to data/secret/secret__sumit4.ans

By default renames and copy files.
Provide two options: 
 - output a script to change files
 - list files to be changed and show counts of files to be created/renamed.

There are scripts to rename and copy samples into the secret directory

Yes, create a new tool that will perform the renaming and copying.

**Additional context**:

lane55:

On --load (contest yaml load) will automatically load sample data files before secret data files.

### Issue which the PR fixes

#244 

java version "1.8.0_121"

Before: Load sample mini contest, for the Sumit (only) problem there are 7 test cases

After: Load sample mini contest, for the Sumit (only) problem there are 10 test cases
and includes the 3 samples


src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2799,6 +2815,31 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
         return problem;
     }
 
+    protected void loadDataFiles(Problem problem, ArrayList<SerializedFile> dataFiles, ArrayList<SerializedFile> answerFiles, String sampleDataDirectory, boolean loadExternalFile) {
+        
+        String[] inputFileNames = getFileNames(sampleDataDirectory, ".in");
+        String[] answerFileNames = getFileNames(sampleDataDirectory, ".ans");
+
```
johnbrvc, line 2822: Shouldn't a check be made to make sure that inputFileNames.length == answerFileNames.length ?

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2740,6 +2752,10 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
 
             ArrayList<SerializedFile> dataFiles = new ArrayList<SerializedFile>();
             ArrayList<SerializedFile> answerFiles = new ArrayList<SerializedFile>();
+
```
johnbrvc, line 2755: Since you wrote a routine to load datafiles from an arbitrary directory, why not use loadDataFiles(problem, dataFiles, answerFiles, dataFileBaseDirectory, loadExternalFile); and fix up loadDataFiles so it can work for both?  I understand there will have to be some small changes to the stuff that sets up the dataFiles and answerFiles arraylists, but it would make it slightly cleaner in the long run.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2799,6 +2815,31 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
...
+        
+        for (int idx = 0; idx < inputFileNames.length; idx++) {
+            
+            problem.addTestCaseFilenames(inputFileNames[idx], answerFileNames[idx]);
+
+            String answerShortFileName = inputFileNames[idx].replaceAll(".in$", ".ans");
+            
+            String dataFileName = sampleDataDirectory + File.separator + inputFileNames[idx];
+            String answerFileName = dataFileName.replaceAll(".in$", ".ans");
+
```
johnbrvc, line 2833: Should a check be made here for idx == 0 to set the problem.setDataFIleName(xxx), problem.setAnswerFileName(xxx) like it is for the secret data, and then, don't do it for the secret data?  Basically, the test should probably be based on the number of test cases already added to the problem (problem.getNumberTestCases()) as opposed to the index of the test case being added.  This way, it will work for both sample and secret data and only add the first one;  the setDataFileName() sets the single name that appears in the problem display as I recall, and if sample data is in the list, THAT should be the file shown for consistency.

johnbrvc (commented):
This is an extremely useful and needed fix, and we should try to get it approved and merged ASAP (maybe for ACPC, assuming they are even using the test-case mechanism that PC2 has).

samps/contests/mini/config/contest.yaml
```diff
@@ -35,6 +35,10 @@ default-validator: /home/judge/cdp/config/validefault {:infile} {:outfile} {:ans
 # Load data files into pc2 as SerializedFiles, not using CDP external files
 load-data-files: false
 
+# load sample files then load secret data files, default is true
+# load-sample-judges-data : false
+# load-sample-judges-data : true
+
```
clevengr, line 41: I do not understand this change.  The text comment (3 lines up) says "default is true", but BOTH of the lines just above are commented-out.  Was the intent that the second line should have been uncommented (making the comment correct)?  Otherwise, I'm totally mystified as to why there are two lines (39 and 40), both comments, giving opposite values.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -266,8 +266,13 @@ private void setCcsTestMode(IInternalContest contest, boolean ccsTestMode) {
         contestInformation.setCcsTestMode(ccsTestMode);
         contest.updateContestInformation(contestInformation);
     }
-    
-    
+
+    private void LoadSampleJudgesData(IInternalContest contest, boolean loadSampleFiles) {
```
clevengr, line 270: Totally minor, picky point:  java method names should start with a lower-case letter.

samps/contests/mini/config/contest.yaml
clevengr: Oh, unless the comment means that in the CODE the "default is true", and this sample YAML provides two options, both of which are currently commented-out but where the user can choose to uncomment one of them to override the code default?   That would make sense... if the comment explained it.  If that's the intent, I think the comment should be expanded to clarify the intent...

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2721,6 +2729,10 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
          * Data files are external so data files should not be loaded into problem data files.
          */
         boolean loadExternalFile = problem.isUsingExternalDataFiles();
+        
+        boolean loadSamples = contest.getContestInformation().isLoadSampleJudgesData();
+        
+        String sampleDataDirectory = dataFileBaseDirectory.replaceAll("secret$", "sample");
```
clevengr, line 2735: These lines of code are confusing; they get the boolean value `loadSamples` and the name of the `sampleDataDirectory`, but then those values are not used in the following lines (which use `dataFileBaseDirectory` instead, and make no use of `loadSamples`).   It would be much clearer if these lines were moved down just in front of line 2756, which is the place where these values are actually used.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2799,6 +2815,31 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
...
+    protected void loadDataFiles(Problem problem, ArrayList<SerializedFile> dataFiles, ArrayList<SerializedFile> answerFiles, String sampleDataDirectory, boolean loadExternalFile) {
+        
+        String[] inputFileNames = getFileNames(sampleDataDirectory, ".in");
+        String[] answerFileNames = getFileNames(sampleDataDirectory, ".ans");
+
+        Arrays.sort(inputFileNames);
+        
+        for (int idx = 0; idx < inputFileNames.length; idx++) {
+            
+            problem.addTestCaseFilenames(inputFileNames[idx], answerFileNames[idx]);
```
clevengr, line 2827: Hmm... this line of code references element `[idx]` in both the `inputFileNames` and `answerFilesNames` arrays.  The first one (`inputFileNames`) has been sorted (line 2823, above), but the second one has not.  Doesn't that mean that the elements at `[idx]` in the two arrays are no longer guaranteed to be corresponding?

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2799,6 +2815,31 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
         return problem;
     }
 
+    protected void loadDataFiles(Problem problem, ArrayList<SerializedFile> dataFiles, ArrayList<SerializedFile> answerFiles, String sampleDataDirectory, boolean loadExternalFile) {
```
clevengr, line 2821: Since this method is being used to load ONLY sample data files, for clarity I would suggest a better method name would be `loadSampleDataFiles`.  Although actually I agree with JohnB's earlier comment that the most logical thing to do would be to use the same method for loading both samples and secrets.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
clevengr: Although I see that this is also the way it was done when only the "secret" files were being loaded, so (presumably?) it works... although I'm not sure how.

clevengr (commented):
I'd like to see responses to the questions/suggestions that both JohnB and I have made.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: Yes, I followed the logic through, and unless I missed something, it would only work if the directory is already in sorted order.  There is also code in that "idx" loop that creates 2 arrays (dataFiles and answerFiles) for use just below in computing "problemDataFiles", which is passed into loadCCSProblemFiles.  It turns out that problemDataFiles is never used by the caller.  Evidently it WAS used at one point for the ClicsOutputValidator, but it is no longer used in there, so that code is dunsel.  Note, however, that the code that calculates dataFiles and answerFiles does go out of its way to compute the CORRECT answer file name based on the input file name. `String answerShortFileName = inputFileNames[idx].replaceAll(".in$", ".ans");`  However, as I said, it is for naught, since those arrays serve no useful purpose.  Similar code should probably be used to form the answer file name before the call to `problem.addTestCaseFilenames()`.  Perhaps:

`problem.addTestCaseFilenames(inputFileNames[idx], inputFileNames[idx].replaceAll(".in", ".ans"));`

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2757,6 +2757,9 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
                 loadDataFiles(problem, dataFiles, answerFiles, sampleDataDirectory, loadExternalFile);
             }
 
+
```
johnbrvc, line 2760: Is there any particular reason we can't do the refactor now since it should be pretty simple to do?

johnbrvc:
There are still a couple of outstanding questions that need answering regarding this PR.
See above comments.

johnbrvc:
What about the  idx == 0 and the setDataFileName() and setAnswerFileName() comment I made above.  This is a bug created by the addition of the sample files being read.  The wrong value will be set if loading both sample and secret data.  The secret data first test case names will overwrite the sample data first test case names.

lane55:
> What about the idx == 0 and the setDataFileName() and setAnswerFileName() comment I made above. This is a bug created by the addition of the sample files being read. The wrong value will be set if loading both sample and secret data. The secret data first test case names will overwrite the sample data first test case names.

> What about the idx == 0 and the setDataFileName() and setAnswerFileName() comment I made above. This is a bug created by the addition of the sample files being read. The wrong value will be set if loading both sample and secret data. The secret data first test case names will overwrite the sample data first test case names.

This is very old code that works fine, there is no bug.

This code are not related to filenames that Executeable uses unless
the input type is File I/O.  You can look in Executable to see where
thses fields are used.

Since we do not use File I/O input type, these fields are not used.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
clevengr: Not something that should hold up approval; just sayin'....

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
clevengr: Again, not something that should hold up approval... I just don't care for adding code that's confusing (even if it executes correctly) when it could be avoided...

johnbrvc:
> > What about the idx == 0 and the setDataFileName() and setAnswerFileName() comment I made above. This is a bug created by the addition of the sample files being read. The wrong value will be set if loading both sample and secret data. The secret data first test case names will overwrite the sample data first test case names.
> 
> > What about the idx == 0 and the setDataFileName() and setAnswerFileName() comment I made above. This is a bug created by the addition of the sample files being read. The wrong value will be set if loading both sample and secret data. The secret data first test case names will overwrite the sample data first test case names.
> 
> This is very old code that works fine, there is no bug.
> 
> This code are not related to filenames that Executeable uses unless the input type is File I/O. You can look in Executable to see where thses fields are used.
> 
> Since we do not use File I/O input type, these fields are not used.

Well, I still believe it is a bug.  The **_Edit Problem_** dialog will not show the FIRST file in the judging data, rather, it will show the first file in the "secret" judging data, instead of the first file in the sample judging data.

load

In the picture, the files in the **_Input Data File_** and **_Answer File_** fields specify the first case of secret data,
not sample data, which would have been "1.in" and "1.ans"  (see WF Dhaka "mosaic" problem).

This can lead to confusion when looking at the **_Edit Problem_** dialog to see the name of the "first" file.  Even though it is not used per se', it is a visual aid to the user as to what judge data is being used.  Basically, it now works (looks) differently than it used to, since the files are being read from two different places.

While this may not be cause for stalling the PR, another issue should probably be created as a "someday" to make those fields reflect reality.  Perhaps in the "refactor" of sharing the loadDataFiles() routine?

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
clevengr: @johnbrvc  The original issue which I pointed out -- the fact that `inputFileNames` was being sorted while `answerFileNames` was not -- has been fixed (by the addition of a call to `Arrays.sort(answerFileNames)`.  However, I'm not sure whether to mark this conversation "resolved" because it appears maybe you introduced an addition issue in your comment.  What's your take -- does the addition of the call to `Arrays.sort()` fix the problem, or is there still another problem (besides "dunsel code") ?

clevengr:
@lane55  said:
> This is very old code that works fine, there is no bug.

It may be "very old code", but it (no longer) "works fine":  as @johnbrvc points out, the addition of the new code which adds the sample data files causes the Edit Problem GUI to display incorrect information.  This in my mind is clearly a bug that's been introduced with the addition of including the sample data files and should be fixed (in other words, the code that adds sample data files is not doing everything correctly that it should).

clevengr:
@lane55 said:

> This code are not related to filenames that Executeable uses unless the input type is File I/O. You can look in Executable to see where thses fields are used.
> 
> Since we do not use File I/O input type, these fields are not used.

It's completely wrong to say "we do not use File I/O".  First of all, most if not all the Sample Contests we supply are configured to use File I/O by default (something that bugs me every time I load SumitHello, for example).  Secondly, even if we didn't provide any Sample Contests (or if we changed them to default to StdIn I/O),  we still provide the option to users to select File I/O.  As long as we provide that option, we should not be including code updates that will display incorrect data if a user selects that mode.  

IMHO, this issue needs to be fixed before this PR can be approved.  (I know you don't want to hear that, but c'mon...  including code that displays wrong data if a user selects a PC2-provided mode?)

johnbrvc:
I think we have to revisit this PR since its functionality is VERY important and automates the loading of a new contest.

lane55:
That Edit Problem existing "bug" should not stop this from moving forward.   If it si critical then open an issue.

The Finals mosaic problem would have been judged correctly, because the fields in the example in Edit Problem
are **not used**.  Those names are not used in a Stdin data source.

johnbrvc:
> 

IMHO since this is such a simple fix, is it really worth going through the whole issue/pr/merge process to address it?  Why not just add it as a CI now with its own commit.  With this PR, it would work "differently" than it used to - and that's my issue with it.  Previously it used to show the first "secret" test-case file name.  Now it will show the (N+1)th test case, where N is the # of sample test cases.  Again, IMHO.

lane55:
I see a lot of testing work to fix this bug, that is why I suggested a new Issue - to define the fix and provide testing for this fix.

johnbrvc (approved):
Since the functionality in this PR is very much needed, I don't think the points raised in the comments should block the approval.

clevengr:
Hmm... the PR has two approvals, but it can't be merged because of conflicts in the ContestSnakeYAMLLoaderTest file.  @lane55 : can you please look at what those conflicts are and resolve them?

clevengr (approved):
I could swear I indicated "Approve" on my last review...

clevengr (approved):

PR #639, opened 2022-12-13
I564 support per problem output limit

### LINKED ISSUE #564: Does not handle per-problem output limit, always uses contest global getMaxFileSize()
**Describe the issue**: 
The newer problem package format, as described here: [https://icpc.io/problem-package-format/spec/problem_package_format#problem-metadata](url) allows a per-problem output file size limit (**output** key) to be specified.  Currently, PC2 does not handle this, and always uses the system wide default value in the contest object.

**To Reproduce**: 
Read a problem package "**problem.yaml**" file that contains an **output** key that has a different value than the default PC2 output limit.  The problem will still use the PC2 default value, not the per-problem value.

**Expected behavior**: 
The maximum output generated for the problem should be set to the per-problem limit, if it is present, otherwise, use the default PC2 value.

**Actual behavior**: 
The default PC2 value is always used.

**Environment**: 
All.

**Log Info**: 
N/A 

**Screenshots**:
N/A

**Additional context**: 
Based on a discussion with @clevengr on 2022-10-14:
Update the **core.model.Problem.java** class to have its own **maxFileSize** value, which can have a default value of 0.
Add accessor (and mutator) methods: **getMaxFIleSize** and **setMaxFileSize** to **Problem.java** (much like **timeOutInSeconds**)
Change **core.execute.Executable.java getMaxFileSize()** method to first check the problem model (**problem.getMaxFileSze()**), and if non-zero, use that as the output size limit, otherwise, return what it does now, **contest.getContestInformation().getMaxFIleSize()**.
Update problem configuration GUI to allow display and configuration of **MaxFileSize** on a per-problem basis with a checkbox to enable using the per-problem value.
Update the reading of **problem.yaml** on the loading of a contest configuration to pay attention to and handle the **output** key.

clevengr:

Adds support for setting "max output limit" on a per-problem basis (whereas currently, all problems acquire the same "global" output limit, either the system default or a value specified using the `max-output-size-K` keyword in the `contest.yaml` file.)

Specifically, adds a "max output in KB" field to the `Problem` class, and provides both interactive  (via the `EditProblemPane` GUI) and YAML (via the ContestSnakeYAMLLoader class) support for setting the max output value for a problem, overriding any value specified in `contest.yaml`.

### Issue which the PR fixes

Fixes #564 

Windows 10.1, Java 1.8.0_201, Eclipse Version: 2019-12 (4.14.0)

#### Test 1:  JUnits
- Run updated JUnit `/pc2v9/test/edu/csus/ecs/pc2/core/model/ProblemTest`, which has updates to test that problems get the correct default max output size if none is specified, and that setting an explicit per-problem output limit works.
- Run updated JUnit `/pc2v9/test/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoaderTest`, which has updates to test various combinations of output limit specifications (default, limit in contest.yaml but not in problem.yaml, limit in problem.yaml, etc.)
- Run the complete JUnit test suite; verify that all JUnits pass.

NOTE:  all the following tests assume that the contest which is loaded into the server has (at least) one problem configured.

#### Test 2: Default max output
- Start a server which loads a contest where there are no specifiers for max output size in either contest.yaml or any problem.yaml.
- Start an Admin, select the ConfigureContest>Problems tab, and verify that every problem has the default MaxOutput(KB) value (512)

#### Test 3: max output defined by `contest.yaml`
- Start a server which loads a contest where there is a `max-output-size-K` setting in `contest.yaml`, but NO settings of max output value in any problem.yaml.
- Verify that each problem  has the Max Output(KB) value specified in the `contest.yaml` file.

#### Test 4: Max output in `problem.yaml` overrides default
- Start a server which loads a contest where there is NO `max-output-size-K` setting in `contest.yaml`, but there IS such a setting in a `problem.yaml` file.
- Verify that the problem containing the max-output-size-K setting has the Max Output(KB) value specified in the `problem.yaml` file.

#### Test 5: Max output in `problem.yaml` overrides `contest.yaml` setting
- Start a server which loads a contest where there is a `max-output-size-K` setting in `contest.yaml`, and ALSO a (different) setting of max output value in a problem.yaml.
- Verify that the problem containing the max-output-size-K setting has the Max Output(KB) value specified in the `problem.yaml` file.

#### Test 6: Max output in "limits" section of `problem.yaml` is recognized
- Start a server which loads a contest where there is a `max-output-size-K` setting in `contest.yaml`, and where a `problem.yaml` file has a `limits:` section containing an `output:` keyword specifying a unique max output value (different from the 512K default and also different from that specifified in `contest.yaml`).
- Verify that the problem containing the `limits: output:` setting has the Max Output(KB) value specified in the limits: section of the `problem.yaml` file.

#### Test 7: Interactive setting of problem output limit
- Start a server which loads a contest with at least one problem.
- On the Admin "Problems" pane, select a problem and click "Edit".
- Enter a non-zero value into the "Max output size (KB) textbox.
- Click "Update".
- Exit the Admin and the Server.
- Restart the server and Admin.
- Verify that the edited problem retained the new max output value.

#### Test 8: Interactive switch to global/default max output limit
- Start a server which loads a contest where there is a `max-output-size-K` setting in `contest.yaml`,
- On the Admin "Problems" pane, select a problem and click "Edit".
- Enter the value zero into the "Max output size (KB)" textbox.  (This is how you specify "no per-problem output limit; use the global setting").
- Click "Update".
- Exit the Admin and the Server.
- Restart the server and Admin.
- Verify that the edited problem retained the max output value defined in the `contest.yaml` file.


src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
```diff
@@ -972,7 +972,7 @@ protected ContestInformation getFromFields() {
         //fill in additional field values
         String maxFileSizeString = "0" + getMaxOutputSizeInKTextField().getText();
         long maximumFileSize = Long.parseLong(maxFileSizeString);
-        newContestInformation.setMaxFileSize(maximumFileSize * 1000);
+        newContestInformation.setMaxOutputSizeInBytes(maximumFileSize * 1000);
```
johnbrvc: Is this supposed to be 1000 or 1024?  I see a mish-mash of 1000's and 1024's throughout.  I know the spec is in MiB (1024).  Is there a reason this is 1000?

clevengr:
I do not understand what's going on.  When I look at the line of code in
question, it DOES contain "1024" instead of "1000".   See screenshot:

//fill in additional field values
String maxFileSizeString = "0" + getMaxOutputSizeInKTextField().getText();
long maximumFileSize = Long.parseLong(maxFileSizeString);
newContestInformation.setMaxFileSize(maximumFileSize * 1000);
newContestInformation.setMaxOutputSizeInBytes(maximumFileSize * 1024);
newContestInformation.setAllowMultipleLoginsPerTeam(
getAllowMultipleTeamLoginsCheckbox().isSelected());
newContestInformation.setTeamScoreboardDisplayFormat(
getTeamScoreboardDisplayFormatTextfield().getText());

src/edu/csus/ecs/pc2/ui/ProblemsPane.java
```diff
@@ -281,6 +283,14 @@ protected Object[] buildProblemRow(Problem problem) {
         // c[i++] = new MCLBCenteredStringCellRenderer(Integer.toString(problem.getTimeOutInSeconds()));
         c[i++] = Integer.toString(problem.getTimeOutInSeconds());
 
+        // problem output limit
```
lane55, line 286: This presents the user with invalid information.   If the max output limit is zero then the user sees another
value entirely.   If they then use the Problems report or export the yaml then zero will be output leaving
the user wondering what happened.   

Note: here is another discussion about the value of zero for max output size in K.

lane55 (changes requested):
Invalid display of max output field in GUI, when field is zero will show value another/default value,  when should show zero
Need to discuss max output field zero as a "default"
In Executable looks like change will unexpectedly set max output file size to 50 bytes.

johnbrvc:
I do.  I DELETED that comment after I realized that it was addressed one 
of the 35 remaining commits on that branch (whew! that's a TON of commits
to go through).  I'm not sure why you even saw the comment, unless you 
got notified when I added it, but then I deleted it a few mins later?
Maybe I should work backwards to avoid missing changes that were done by 
a later commit

src/edu/csus/ecs/pc2/ui/ProblemsPane.java
johnbrvc: I agree.  Instead of using "0" to mean "use the system global", perhaps have a check box that says: "Override maximum output for this problem", then an edit box next to that checkbox that becomes enabled when the checkbox is checked.  Basically, any place where you can enter 0 to "use the default" should maybe have a checkbox as described?  This is usually what we do in our GUI apps when the user has a choice to override something.  They have to make a conscious decision to override (checkbox), and then (and only then) can they type text in the override box. Just a suggestion.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2336,12 +2336,45 @@ private String searchForPackage(String file) {
...
+                
+        } else {
+            
+            //problem is null; log error and return zero
+            if (log != null) {
+                log.log(Log.WARNING, "Problem is null, cannot determine output size limit; returning zero");
+            } else {
+                System.err.println("WARNING: log is null; cannot log message 'Problem is null, cannot determine output size limit; returning zero'");
+            }
+            return 0L;
```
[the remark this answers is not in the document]
lane55: Resolved, change 0L to  Constants.DEFAULT_MAX_OUTPUT_SIZE_K*1024

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -34,7 +34,9 @@ public class Problem implements IElementObject {
     private static final long serialVersionUID = 1708763261096488240L;
 
     public static final int DEFAULT_TIMEOUT_SECONDS = 30;
-    
+
+    public static final int DEFAULT_MAX_OUTPUT_FILE_SIZE_KB = 0; // zero indicates no per-problem output limit has been set
```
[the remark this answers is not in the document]
lane55: comment updated.

src/edu/csus/ecs/pc2/ui/ProblemsPane.java
lane55: After meeting, an explanation was given that 0 for max in Problems column would be accurate but misleading.   New code added in Executable to change 0 to the model default max (from ContestInformation).

Also noted that in Edit Problem the max is displayed as is, so zero shows max output limit to be zero.   There also is a
description in the "What is this" icon next to the field (getLblWhatsThisMaxOutputSize())

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -572,7 +572,7 @@ public IInternalContest fromYaml(IInternalContest contest, String[] yamlLines, S
             if (maxOutputSize instanceof Integer) {
                 int maxSizeInK = ((Integer) maxOutputSize).intValue();
                 if (maxSizeInK > 0) {
-                    setMaxOutputSize(contest, maxSizeInK * 1000);
+                    setMaxOutputSize(contest, maxSizeInK * 1024);
```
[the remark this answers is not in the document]
lane55: Constant added and used: Constants class,  public static final int BYTES_PER_KIBIBYTE = 1024;

lane55 (approved):
Approved

lane55 (approved):
Approved

johnbrvc (approved):
The comments I had were echoed for the most part by @lane55's.
Not a show-stopped, but just a suggestion regarding using a check-box to indicate that a problem is overriding the output limit on the edit problem pane.  Again, this is just a personal preference, and what is there now is sufficient.

PR #648, opened 2022-12-21
fix reference to github.ref

troy2914:
i was testing this in troy2914/pc2v9 with a echo on $TARGET after the if,  and i push to master showed it as going to builds.
previously it was not going there.


clevengr (approved):
I've looked at this change and at why the code it replaces didn't work; it makes sense.

PR #657, opened 2023-01-07
i656: expand Admin clockLabel and clockPane to avoid truncation.

### LINKED ISSUE #656: Start Contest countdown display is truncated
**Describe the issue**: 
When a contest is configured with a future "Start Time" and then an Admin is started, the upper left corner of the Admin displays a "countdown to start of contest"  (in the same field where the "contest remaining time" is displayed as a countdown once the contest actually starts).   The field in which this "countdown to start of contest" is displayed is not wide enough to handle certain very reasonable countdown values.  

For example, if you configure a contest to start in, say, two weeks (which is not unreasonable in today's online contest environment), the display will look like `14 Days 1.` -- when what is intended is (something like) `14 days 10:57:03`. 

The issue is in the default width attribute of either the `JPanel clockPane` or `JLable clockLabel` fields within the `JPanel centerPane`  contained in the `JPanel topPanel` of the top-level `JPanel` in the `ui.admin.AdministratorView` JFrame.  
The `clockLabel` and `clockPane` classes were most likely written back when it was inconceivable that the countdown values in those fields would ever be more than just a few hours...

**To Reproduce**: 
- Configure a contest with a Scheduled Start Time which is at least 10 days, but less than 30 days, in the future.
- Start an Admin.
- Look at the "countdown to contest start" value in the upper left corner.

**Expected behavior**: 
The countdown should say (something like) `10 days 18:06:33` (if indeed the specified start time is 10 days, 18 hours, 6 minutes and 33 seconds from "now").

**Actual behavior**: 
The countdown in the upper left corner of the Admin will say (something like) `10 days 1.`.  In other words, the full countdown value is truncated.

**Environment**: 
 - OS name and version under which the issue was detected: Windows 10.1 (but it probably happens in most environments).
 - PC² version 9.9build-6468.develop.zip   

**Screenshots**:  
A screenshot showing the error:
image

Note that with regard to the above screen shot,
- the actual time/date at which this screenshot was taken was `2023-01-06 22:15`.
- the "scheduled start time" is shown as `2023-02-05 21:00`, in other words, about 29 days, 22 hours, 42 minutes (or so) from "now".
- the "countdown time" is truncated.  

The Admin screen _should_ look something like:
image

**Additional context**: 
- The second screenshot above (i.e. with the countdown fully visible) was generated after changing `clockLabel.preferredSize` to 200 pixels wide and changing `clockPane.preferredSize` to 220 pixels wide.
- The above "fix" does have the side effect of _reducing_ the available pixels in the adjacent "messagePane".  This seems like a "reasonable" tradeoff...

Note:
The reason the above "Steps to reproduce" specify that the start time should be _less than 30 days_ in the future is that there is logic in the code which displays the "countdown" such that if the start time is more than a month away, it just displays `xx Months`, with no HH:MM:SS fields.  (Similarly, if the start is more than a year away it just displays `xx Years` without displaying Months, Days, Hours, Minutes, or Seconds.)

clevengr:
Expands the width of the AdministratorView `clockLabel` and `clockPane` fields to avoid truncation of large "future start time" values.

### Issue which the PR fixes
Fixes #656 

- Windows 10.1
- Eclipse Version: 2019-12 (4.14.0)

- Start a server configured with a contest having a Scheduled Start time that is at least 10 days (but not more than 30 days) in the future
- Start an Admin.
- Examine the "Countdown to start of contest" field in the upper left corner of the Admin view.  Note that it displays `xx Days hh:mm:ss`, where `xx` is a two-digit number of days.  Note also that the entire countdown (that is, the days, hours, minutes, and seconds) is displayed (i.e. that it is not truncated.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/admin/AdministratorView.java
+++ b/src/edu/csus/ecs/pc2/ui/admin/AdministratorView.java
@@ -577,11 +577,11 @@ public class AdministratorView extends JFrame implements UIPlugin, ChangeListene
         if (clockPane == null) {
             clockLabel = new JLabel();
             clockLabel.setText("STOPPED");
-            clockLabel.setPreferredSize(new Dimension(100, 24));
+            clockLabel.setPreferredSize(new Dimension(200, 24));
             clockLabel.setFont(new java.awt.Font("Dialog", java.awt.Font.BOLD, 18));
             clockPane = new JPanel();
             clockPane.setLayout(new BorderLayout());
-            clockPane.setPreferredSize(new java.awt.Dimension(85,34));
+            clockPane.setPreferredSize(new Dimension(220, 34));
             clockPane.add(clockLabel, BorderLayout.WEST);
         }
         return clockPane;


troy2914:
I would like a tooltip with the full string displayed always when you hover on that string.   Was this change tested on Ubuntu?

lane55 (commented):
approved

lane55 (approved):
Approved

johnbrvc (approved):
Haha. I saw this same issue because our practice starts in 24 days.

clevengr:
> I would like a tooltip with the full string displayed always when you hover on that string. Was this change tested on Ubuntu?

I don't see any benefit to "a tooltip with the full string displayed", because with this fix the full string will always be displayed in the GUI.  What's the point of having a tooltip that displays the same thing as what's already in the displayed field?

But no, I didn't test it on Ubuntu and yes, that should be done.  I'll work on that.

johnbrvc:
> > I would like a tooltip with the full string displayed always when you hover on that string. Was this change tested on Ubuntu?
> 
> I don't see any benefit to "a tooltip with the full string displayed", because with this fix the full string will always be displayed in the GUI. What's the point of having a tooltip that displays the same thing as what's already in the displayed field?
> 
> But no, I didn't test it on Ubuntu and yes, that should be done. I'll work on that.

Tested on Ubuntu - looks fine.

image

lane55:
Tested tool tip on Windows it was added years ago because of this very bug.
In a number of places on the GUI where JLabels are cutoff (esp under unix) added a tooltip with the full field/label text.

lane55:
The benefit using the tool tip is if the label is cut off and since it is not auto resized it will be cut off if
the hours/date are large enough, then there is a way to view the full value.

Issue #626, opened 2022-12-02
load-data-files json key is written to exported problem.yaml file(s)


johnbrvc:
**Describe the issue**: 
When exporting the contest yaml from the reports tab, the individual problem.yaml files for each problem contains the load-data-files json key.  This key is never read in when the problem.yaml file is loaded.  The key is actually read from the problemset.yaml file under each problem's section.

**To Reproduce**: 
Simply export a contest YAML from a reports tab and look int he exported problem.yaml files and you'll find the load-data-files key in them.

**Expected behavior**: 
The key should not be in the problem.yaml.  Note that it IS exported to the problemset.yaml correctly for each problem.  

**Actual behavior**:
The load-data-files key is exported to both the problemset.yaml file in each problem's section, AND to the problem's problem.yaml.

**Environment**:
All

**Log Info**:  
**Screenshots**:  if applicable, please add attachments showing screenshots to help explain the issue.

**Additional context**: 
This caused a create deal of confusion at the ACPC 2022 contest when Coach believed that changing the problem.yaml's load-data-files key would actually have an effect (it did not).  He assumed since it was exported from PC2 it would be used.

We can discuss if the load-data-files key should be in the problemset.yaml OR problem.yaml, but it should NOT be exported to both.  Pick one and stick with it.

lane55:
confirmed: load-data-files in problem.yaml is ignored, whereas the value in the *yaml files (contest, problemset,etc) is used.

lane55:
I think load-data-files   should be removed from problem.yaml (and any code/support/refrence removed form
when problem.yaml is loaded.)

lane55:
With those changes  load-data-files  should only be exported to contest.yaml

lane55:
PR deleted, Small LOE removed, this "fix" requires additional analysis including
implications about flag used in Edit Problem.

lane55:
marked low priority because a seemingly trivial change has been made non-trivial

PR #675, opened 2023-01-31
i626: remove load-data-files json key from problem yaml files

lane55:

remove load-data-files key/line from export to problem.yaml
remove load-data-files key/line from export to problemset.yaml

keep load-data-files key/line in export to contest.yaml

### Issue which the PR fixes

Fixes 626 

java version "1.8.0_121"

start server, --load problemflagtest
start admin
export using Run Contest -> Export -> Export Yaml
Select Save

### Expected

There should be no load-data-files line 
in sumit/problem.yaml nor in problemset.yaml

### Actual

in sumit/problem.yaml
```
   load-data-files: true
```

in sumit/problemset.yaml
```
   load-data-files: true
```

### DIFF
--- a/src/edu/csus/ecs/pc2/core/export/ExportYAML.java
+++ b/src/edu/csus/ecs/pc2/core/export/ExportYAML.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2022 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.export;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


johnbrvc (approved):
Seems to work for me.
By the way, on the PR main page, it does not automatically detect that this closes issue 626 (on the right pane).  I thought it should detect that, maybe you need "- fixes #626" or some such on the PR title?

clevengr (changes requested):
I question whether the change implemented by this PR is reasonable.  Specifically, as I understand the PR it removes support for the inclusion of `load_data_files` from `problem.yaml` and also from `problemset.yaml`, and causes it to only be recognized/imported/exported to/from `contest.yaml`.

This is the same thing as saying that the `load_data_files` attribute is (solely) a _contest-wide_ attribute.  However, this is NOT accurate:  the "load-data-files" property is specified in the Admin>Configure>Problem>Edit screen _on a per-problem basis_.  Having it specified per-problem (which is appropriate) in the Admin GUI and specified (solely) contest-wide in YAML (as I believe is what is proposed by this PR) produces contradictory configuration information.

Note:  if the PR was to ADD support for load-data-files to `contest.yaml`, causing it to be a "global _default_ value" which could then be overridden by problem-specific entries in `problem.yaml`, much like `timelimit` and `output_limit` work currently, I'd be OK with that.  But I do NOT think it's appropriate to _remove_ support for it in `problem.yaml` -- it's a _problem-specific setting_.  

(I could go either way on keeping it, or removing it, from `problemset.yaml`).

clevengr:
And yes, as @johnbrvc pointed out, simply putting the Issue Number in the PR does not cause it to close that Issue; the Issue number must be preceded by a word like "Closes" or "Fixes".  PR #685 changes the PR template to (hopefully) clarify this.

lane55:
closed, changes required require additional analysis and testing.

Issue #680, opened 2023-02-08
WTI scoreboard is blank or shows all teams from all divisions


lane55:
**Describe the issue**: 

A team WTI scoreboared with 3 divisions should show just
the team's divisions rankings/runs.  The scoreboard shows
all teams in all divisions.

Under certain cirsumstances the WTI scoreboard is blank
and throws an exception, found in the log and below
Log Info section.

**To Reproduce**: 

Use tc1 CDP at PacNWSpring2023/testcontest1/config

start server with tc1
start wti 
login as team101
click Scoreboard

**Expected behavior**: 

Should show only teams from D1

**Actual behavior**: 

Shows all 47 teams from all 3 divisions

**Environment**: 

Linux ip-172-16-1-46 5.15.0-1028-aws #32-Ubuntu SMP Mon Jan 9 12:28:07 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment (build 17.0.5+8-Ubuntu-2ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.5+8-Ubuntu-2ubuntu122.04, mixed mode, sharing)

**Log Info**:  

The scorboard2's scoring settings for Minute penalty is null

Caused by: java.lang.NumberFormatException: Cannot parse null string
    at java.base/java.lang.Integer.parseInt(Integer.java:630)
    at java.base/java.lang.Integer.parseInt(Integer.java:786)
    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.getPropIntValue(DefaultScoringAlgorithm.java:248)
    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.getPenaltyPointsPerYesMinute(DefaultScoringAlgorithm.java:278)
    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.calcProblemScoreData(DefaultScoringAlgorithm.java:208)
    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.generateStandingsValues(DefaultScoringAlgorithm.java:947)
    at edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm.getStandings(DefaultScoringAlgorithm.java:447)
    at controllers.ContestController.getStandings(Unknown Source)

**Screenshots**:  

**Additional context**:

lane55:
[...]

Issue #32, opened 2020-03-05
Halt non-GUI shadow if shadow cannot start, issue message


lane55:
**Feature Description**: 
When starting a shadow from the command line and the shadow cannot start, ex. cannot contact primary CCS, halt the program and issue an error message.

Currently the program silently fails and continues.   This give a false perception that the shadow has started.

**Have you considered other ways to accomplish the same thing?** 

Yes, but that requires a user to know know that the shadow has failed... somehow.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, please describe.  

If there is any Exception on shadow startup, 
1 - print the message and full exception to the log
2 - print the exception message to console
3 - print message "Shadow could not start, halting program"
4 - System.exit(10);  // non zero exit code

**Additional context**: 

none

clevengr:
This issue was handled in PR #60 ; once that PR has been _verified_ then this issue can be closed.

PR #681, opened 2023-02-08
i680: fix NPE by assigning scoring properties to properties

lane55:
NPE caused by sometimes missing scoring properties before calling dsa.getStandings(internalContest
assigned default DSA properties, and NPE avoided.

For tc1, on WTI team, Instead of showing blank scoreboard now show 47 teams.

### Issue which the PR fixes

partial #680

Developed under these, but bug not reproduced.

java version "1.8.0_121"

Test using:

Linux ip-172-16-1-46 5.15.0-1028-aws #32-Ubuntu SMP Mon Jan 9 12:28:07 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment (build 17.0.5+8-Ubuntu-2ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.5+8-Ubuntu-2ubuntu122.04, mixed mode, sharing)

load tc1, --load PacNWSpring2023/testcontest1/config
start WTI
login as team 101
click Scoreboard
expect 47 teams listed

### Actual

click Scoreboard
shows blank screen


clevengr:
I just tried following the "Steps to Reproduce" listed in the Issue (#680 ).  I downloaded the PacNWSpring23/testcontest1, then ran it using the December 2022 Public Distribution (for WTI) and the pc2v9 GitHub "master" branch for server/admin/judges (which I believe should effectively be the same as running the Dec 22 public Distribution).  The WTI scoreboard was NOT blank and the Scoreboard log did NOT show any exceptions thrown.  So, unable to reproduce the issue that way.

I then reran the contest from scratch (complete reset) using the Develop branch for the server/admin/judges but still using the December 2022 Public Distribution  for WTI.  Once again, the WTI scoreboard was NOT blank and the Scoreboard log did NOT show any exceptions thrown.  

I then rebuilt the entire pc2v9 system (locally) and reran the same test using the local build; same results -- could not reproduce the issue.  

Then I went back and tried using the Dec '22 Public Distribution WTI -- and lo & behold, I got a blank WTI scoreboard, and a NPE in the WTI log.

Then I went back and tried the local build again (exactly what had just previously worked), and _again_ I got a blank scoreboard (and NPE).  

Bottom line:  I have no idea exactly how to reproduce the problem.

BTW, the WTI "Problem List" dropdown IS properly filtering out problems that aren't in the logged-in team's division, so the only problem is that the WTI Scoreboard isn't doing the same filtering...

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -566,6 +573,30 @@ public Response getStandings(
...
+
+						        Arrays.sort(keys);
+								xlog(logger, "debug 22   Scoring Properties, there are " + keys.length + " keys");
+
+						        for (String string : keys) {
+									xlog(logger, "debug 22      " + string + "='" + props.get(string) + "'");
+						        }
+						}
+						
+						props = DefaultScoringAlgorithm.getDefaultProperties();
```
clevengr, line 591: I do not understand why `props = DefaultScoringAlgorithm.getDefaultProperties` is _always_ being executed.  I believe it should only be executed IF the above call to `Properties props = internalContest.getContestInformation().getScoringProperties();` returns null.  Otherwise, the code is forcing the use of "default properties _even when valid properties have been obtained_.

Said another way, I think this line should be an `else` for the previous `if` statement, not an unconditionally executed statement.

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -566,6 +573,30 @@ public Response getStandings(
...
+						xlog(logger, "debug 22  set Default Scoring Props for "+key);
```
clevengr, line 592: What is the value of the variable `key` in the above statement?  (As far as I can see, it's undefined...

projects/WTI-API/src/main/controllers/ContestController.java
clevengr: However, if what you are addressing is that the `ScoringProperties` object is coming back non-null, but one or more of the Properties _inside that object_ is null when it shouldn't be, then I guess this makes sense.

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -566,6 +573,30 @@ public Response getStandings(
...
+
+						
+						// TODO use key to find clientId then use clientId to find division runs for this client
+//						ClientId clientId = null;
+//						Run[] runs = ScoreboardUtilites.getRunsForUser(clientId, internalContest);
+						
+                        Run[] runs = internalContest.getRuns();
```
clevengr, line 599: What is the purpose of the above statement?  It defines a variable which (as far as i can see) never used.  Please remove, or explain why it's there.

clevengr (approved):
There are minor problems in the code (which I've noted) but nothing that rises to the level of refusing to approve.  But we need a new issue regarding the root cause of some of the properties returned from `Properties props = internalContest.getContestInformation().getScoringProperties();` being null.

projects/WTI-API/src/main/controllers/ContestController.java
lane55: Why unconditionally assignment?   Crunch time.  Needed to fix it to go onto the the next task.

lane55: Also with a conditional there would be additional testing and one of the other fixes was to 
add a .ini entry (PR #682 ), and further there are three different integer scoring properties...  so
strictly it would be 3 conditionals and 3 tests.

projects/WTI-API/src/main/controllers/ContestController.java
lane55: It is debugging it will be removed.

lane55: Answer is ... key is the team login name, ex 'team151'

projects/WTI-API/src/main/controllers/ContestController.java
lane55: Hmm, the commented out section is how we'll filter the team's runs

lane55: The getRuns is existing code, it seems to be a github diff thing, the getRuns is not new.

lane55:
> There are minor problems in the code (which I've noted) but nothing that rises to the level of refusing to approve. But we need a new issue regarding the root cause of some of the properties returned from `Properties props = internalContest.getContestInformation().getScoringProperties();` being null.

done.  Added issue #686 - Find/fix root cause for NPE in WTI DSA, details in i680

lane55:
> BTW, the WTI "Problem List" dropdown IS properly filtering out problems that aren't in the logged-in team's division, so the only problem is that the WTI Scoreboard isn't doing the same filtering...

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -543,13 +547,16 @@ public Response getStandings(
 		try {
 			logger.info("Standings requested by team " + userInformation.getMyClient().getLoginName());
 			
+			xlog(logger, "debug 22 Start" );
```
johnbrvc, line 550: Is the intent to leave this debugging log stuff in production code? If not, it should be removed.

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -566,6 +573,30 @@ public Response getStandings(
 					// we got the internal contest; pass it to the DefaultScoringAlgorithm and get back updated standings
 					try {
 						Properties props = internalContest.getContestInformation().getScoringProperties();
+
+						xlog(logger, "debug 22 props "+props);
```
johnbrvc, line 577: Is the intent to have all the xlog debug 22 messages in production?  If not, they should be removed.  To me, they clutter up the code.

projects/WTI-API/src/main/controllers/ContestController.java
johnbrvc: Is this going to be removed?  Comment indicates it would be removed, but it appears to still be there?

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -609,7 +640,17 @@ public Response getStandings(
 		}
 	}
 
-	
+	/**
```
johnbrvc, line 643: Is this debug code being left in production?

johnbrvc (changes requested):

lane55:
> I just tried following the "Steps to Reproduce" listed in the Issue (https://github.com/pc2ccs/pc2v9/issues/680 ). ...
Me neither, the only time I was able to reproduce the problem was stated in the PR description

projects/WTI-API/src/main/controllers/ContestController.java
lane55: No. in fact changes for #680 have already been pushed without debug 22, so this PR/branch may
be moot.

lane55:
Sent out in an email with subject: i680 status, PR 681 is moot

re: summary

   I think that PR #681 (https://github.com/pc2ccs/pc2v9/pull/681)
   can be closed because all fixes are present/merged into develop already.

   The comments which were made about debug 22, etc.
   struck me wrong, I remembered fixing those items already, turns
   out that I did and they were merged into develop.

   PR 682 was used to merge the 680 changed into develop

   Issue 686 was opened to find root cause for why there was a WTI
   blank scoreboard.
   https://github.com/pc2ccs/pc2v9/issues/686

re: gory details 

   There is an outstanding PR #681
   i680: fix NPE by assigning scoring properties to properties #681

   but it looks like that is "old" code (the develop code
   already has fixes for that PR).

   There was a previous PR for 680 that was pushed into develop
   i680: implement filter WTI scoreboard by team's division #682
   (this removed all debug 22 logging messages, etc.)

   There were 2 fixes in PR 682 ( https://github.com/pc2ccs/pc2v9/pull/682 )

   PR682 first was WTI Scoreboard was blank 
      root cause was the scoring properties integer values
      were null.
      fixed by force assigning DSA.getDefaultProperties
      line 618 in ContestController on develop branch
      and added 
// TODO i686 - Find/fix root cause for NPE in WTI DSA, details in i680 #686

   PR682 - Next was adding support for divisions
      Added ini variable to enable, from PR...
Add to pc2v9.ini in [server] section
wtiBoardUseDivisions=Yes

   There was also a need to regenerate the standings each
   time a team requested their standings, they needed to 
   see their standings with their divisions teams.

   So we conditionally ( wtiBoardUseDivisions=Yes ) added
   a forced recalc at line 578 class ContestController

re: but WTI blank scoreboard cannot be reproduced

   I was not able to reproduce the problem except during the PacNW
   with the following env, as stated in the PR.   On my laptop there
   was no bug.

   Linux ip-172-16-1-46 5.15.0-1028-aws #32-Ubuntu SMP Mon Jan 9 12:28:07 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment (build 17.0.5+8-Ubuntu-2ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.5+8-Ubuntu-2ubuntu122.04, mixed mode, sharing)

johnbrvc (approved):

PR #696, opened 2023-02-22
i695: adjust submissions event "files" href (add /contest)

### LINKED ISSUE #695: Event feed serves wrong end point for submitted files
**Describe the issue**

The PC2 Event Feed client  produces events of type "submissions".  Such events contain a field "files" which contains an href giving the URL which is supposed to be able to be used to obtain a zip of the files associated with the submission.  Currently the format of the field in a `submissions` event is:
```
"files":[{"href":"/submissions/1/files"}]
```
This is an _absolute_ path, which per the CLICS specs any clients should use as-given (whereas if the path were a _relative_ path then spec-compliant clients should handle using "multi-contest-awareness" paths -- in other words, clients should add a "contest-id" at the front of the path).

The problem is that the PC2 webserver (in class`services.web.SubmissionService.java`) does not recognize path "/submissions"; it defines the SubmissionService endpoint as `/contest/submissions`.

The "files" href should look like:
```
[{"href":"/contest/submissions/1/files"}]
```

Ultimately of course, PC2 should be made CLICS multi-contest compliant, but for now at a minimum the href in the `files` field needs to be changed to match what the webserver expects.

clevengr:
  Changes the hardcoded `href` entry in the `files` field of a `submissions` event from `/submissions` to `/contest/submissions` (one-line change).   This is necessary to provide an href that matches what the PC2 SubmissionsService is providing as an endpoint.

### Issue which the PR fixes
Fixes #695 

Windows 10.1, Java 1.8_271

- Start a contest with multiple submissions.
- Start an Event Feed client.
- Start a CDS hitting the EF.
- Verify that the CDS can pull submissions (previously it was getting 404s on every submission files access request)

### DIFF
--- a/src/edu/csus/ecs/pc2/core/util/JSONTool.java
+++ b/src/edu/csus/ecs/pc2/core/util/JSONTool.java
@@ -89,7 +89,7 @@ public class JSONTool {
         // TODO shadow add time and mime elements to submission
 //        element.put("mime","application/zip");
         
-        String pathValue = "/submissions/" + submission.getNumber() + "/files";
+        String pathValue = "/contest/submissions/" + submission.getNumber() + "/files";
         
         ObjectMapper mymapper = new ObjectMapper();
         ArrayNode arrayNode = mymapper.createArrayNode();


ashoos72 (approved):
verified correct path added to href and tested against CDS

PR #721, opened 2023-03-20
i343: Autojudge/manual judge makes desktop unusable while judging - fixes #343

### LINKED ISSUE #343: Auto-judge GUI makes desktop unusable while it's judging

**Describe the issue**: 
On a machine that is auto-judging (pc2judge in AJ mode using **GUI**), the machine is virtually useless for anything while autojudging is enabled due to focus being lost.

**To Reproduce**:
Enable AJ (pc2judge gui) on a set of busy problems.  Then, try to do anything else on that computer once runs start coming in.
The pop-ups from AJ for each test case make it impossible to do anything.  

**Expected behavior**: 
The system should be usable while AJ is going on.   Maybe have an option on the AJ to not do popups for judging status
while judging (compiles, runs, etc).  (I am aware of the --nogui option, but it still would be nice if the GUI version didn't hog
the whole desktop by stealing focus).

**Actual behavior**:
The system loses focus constantly while pop-ups are displayed for judge cases.  Sometimes, typing while test cases are popping causes the judging of that problem to fail.  I think it's a function of what was being typed at the time the pop-up comes in.

 - Ubuntu 18.04
 - PC2 Version (e.g. 9.7.0):   9.7.0

**Log Info**: 
N/A  

**Screenshots**: 
N/A

johnbrvc:
Instead of creating a new frame for each step  (compile, executing each test case, validate), a single frame is used and updated in place.  This prevents focus from being stolen from any other active application.  The act of creating a new frame steals focus.  Updating a frame does not. 

In addition, a more descriptive message is shown while executing the test cases (the test case # is shown).

#343 

Windows 11
Eclipse Version: 2021-12 (4.22.0)
JDK1.8.0_321

1) Bring up any contest that has submissions (preferably one that has a submission for a problem that has many test cases.
2) Select a submission with lots of test cases (preferably)
3) Rejudge (or judge) the run.
4) Note the execute window is created once and updated in place instead of being created for each step.  While the submission is being compiled/executed, and another application has focus, the focus will not be stolen from that application.
5) You can repeat the the process for a GUI auto-judge by enabling the auto-judge, editing a submission, and changing it to "QUEUED FOR COMPUTER JUDGEMENT", at which point the auto-judge will pick it up and execute it, again, not stealing focus, but reusing the same frame over.

Notes: 
Essentially, a lot of the execute timer frame code from ExecuteTimer.java was moved into a new class, ExecuteTimerFrame.java.  This new frame is created once in Executable.java and passed into ExecuteTimer.java instead of having ExecuteTImer create a new one each time.


johnbrvc:
I just noticed there are some problems with the unit tests not building due to the changes.   Since I can't build the unit tests on my laptop in eclipse, I'll build by hand and see if I can fix those tests.

johnbrvc:
> I just noticed there are some problems with the unit tests not building due to the changes. Since I can't build the unit tests on my laptop in eclipse, I'll build by hand and see if I can fix those tests.

I have addressed the unit test issue and the PR should now be good to go.  It should also be noted that this code was used for the 2022 Greater NY Regional on Feb 26, 2023 without any issues.

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.awt.Color;
+
+/**
+ * @author John Buck
+ *
+ */
+public interface IExecuteTimerFrame {
+
+    // sets the frame to it "known" state
```
lane55, line 17: change // to Java doc comment

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.core.execute;
+
+/**
+ * @author John Buck
+ *
+ */
+public interface IExecuteFrameNotify {
+
+    // Terminate button pressed
```
lane55, line 10: change // to java doc comment, perhaps comment like:
Notify Observer that auto judging has been terminated.

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
```diff
@@ -1171,7 +1174,7 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
     }
 
     private String getExecuteDirectoryName() {
-        Executable tempEexecutable = new Executable(getContest(), getController(), run, runFiles);
+        Executable tempEexecutable = new Executable(getContest(), getController(), run, runFiles, null);
```
lane55, line 1177: No change required, note for future development
OOP note, instead of requiring a null param (esp at end) create a new constructor

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
lane55, line 17: picky picky.  to it's ...
Also "known" is too vague, it this the initial state?

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
```diff
@@ -222,4 +231,59 @@ public String getPluginTitle() {
...
+                }
+            });
+           
+        }
+    }
+
+    @Override
+    public void setTimerFrameVisible(boolean bHow)
+    {
+        // The AJ Status frame is not controlled by the execution timer, but the AJ itself
```
lane55, line 249: Seems like an incomplete comment, complete it.

lane55 (changes requested):
No code changs, but a number of comments should be updated

lane55:
Test results, execute no longer steals focus - looks good.
Noted that there is a flickering JButton now that toggled between Terminate and Stop AutoJudging

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
johnbrvc: Noted.  I just want to be sure anyone that uses this class knows how it works and what a null frame implies; requiring that extra frame param enforces that contract (to some extent).  That is all.  I think we spoke about this offline - in C++ I might make that last argument a default argument, eg( ...  , frame = NULL).

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
johnbrvc: Cleared up the confusion, hopefully.

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
johnbrvc: Changed.

johnbrvc: It's fixed.

lane55: confirmed (fixed by removing/replacing comment_

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
lane55: code still contains // Terminate button pressed
should be java doc.

lane55: The github diff used, shows the // comment was still there.

lane55: confirmed fix, reviewed code directly on branch and it is fixed.

lane55 (approved):
Approved.
changes made

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * show or hide the timer frame
+     * 
+     * @param bVis
+     */
+    public void setTimerFrameVisible(boolean bVis);
+    
+    /**
+     * change the color of the timer text (for the "Test" button if execute time gets too big)
+     * @param fg
```
clevengr, line 34: I do not understand this comment.  It talks about a "Test" button... but as far as I can see there's no such button on the implementing frame.

johnbrvc: Recall that the same ExecuteTimerFrame is used in the Team model when a team selects the "Test" button.  In this case, the run is allowed to go past the assigned time limit and the CPU time turns red indicating it has exceeded the time limit.

clevengr: Ok.  But doesn't this mean that some implementers of this Interface will have the need to change the Timer label color, while others will not (because they weren't invoked from a frame containing a "Test" button)?   Which in turn means that implementers of the Interface that do NOT need to have a changeable-color capability are _required_ to implement a method which is irrelevant to them.  (It's true that the "implementation" can be trivial -- simply `return` -- but (in IMHO) when an interface requires implementing a method that's not needed, it indicates a potential flaw in the Interface design.)

In particular, it seems like there should be a "super-interface" that provides a default implementation of the method (e.g. one that simply executes `return`), and a "sub-interface" that doesn't mention the method.  That way, classes which don't need the ability to change the label color implement the sub-interface (and inherit the super-interface default implementation) without ever having to pay attention to the notion of changing the color, while classes which do need to change the color can override the method.  This seems like it better follows OO design principles.

It could also be turned around the other way:  have the super-interface _not mention_ the method (and classes which have no need for the color-change capability implement that interface), then have a derived Interface which simply adds the color-change method (and classes which need that method implement the sub-interface).  I'm not sure which of these two approaches is "better"...

And note:  I'm not asking for such a change; I wouldn't consider this item by itself a reason for not approving the PR.  I'm just pointing out what I think is a (slight) flaw in the design.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
```diff
@@ -138,10 +143,14 @@ private JPanel getMessagePanel() {
     private JButton getStopAutoJudgingButton() {
         if (stopAutoJudgingButton == null) {
             stopAutoJudgingButton = new JButton();
-            stopAutoJudgingButton.setText("Stop AutoJudging");
+            stopAutoJudgingButton.setText("Stop Auto Judging");
```
clevengr, line 146: I'm not sure about this change (although I grant it's trivial).  We refer to the components doing automatic judging as "AJ"s, not "A J"s, and the relevant classes are named like `AutoJudgeStatusFrame`, not `Auto_Judge_Status_Frame`.

I guess it's a personal style thing...  but I'll bet there are _dozens_ of places where "AutoJudging" is used, not "Auto Judging"...

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
```diff
@@ -138,10 +143,14 @@ private JPanel getMessagePanel() {
...
             stopAutoJudgingButton.addActionListener(new java.awt.event.ActionListener() {
                 public void actionPerformed(java.awt.event.ActionEvent e) {
-                    startStopAutoJudging();
+                    if(notifyFrame != null) {
+                        notifyFrame.executeFrameTerminated();
+                    } else {
+                        startStopAutoJudging();
+                    }
```
clevengr, line 153: I'm not at all clear about the changes in the above `actionPerformed()` method.  Previously, pressing the "start/stop AutoJudging" button (i.e., invoking this method) would unconditionally invoke method `startStopAutoJudging()`, which serves two functions: (1) toggling the label on the "Start/Stop AutoJudging" button between Stop/Start, and (2) enabling or disabling the `AutoJudgingMonitor` (that is, making the AJ start/stop auto-judging runs as appropriate).

Now, if `notifyFrame != null`, this method is never called.  This doesn't seem logical to me.  It appears to me that the intent of the PR is to have the **Terminate** button notify the `notifyFrame` that the operator (user) has terminated the current execution.  That seems to me to be something totally unrelated to what happens when the `startStopAutoJudging` button is pressed; the former has to do with terminating a Submission execution while the latter has to do with enabling/disabling the AJ's ability to fetch/judge runs, which are two totally separate things.

Or am I totally misunderstanding the structure of the PR?  (Which is possible; it's quite complex and I'm struggling to get a clear understanding of what's going on...)

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
johnbrvc: Actually, I take that back - I misread the code (`!isTeam()` instead of `isTeam()`).  In fact, the timer label is only set to red if the client type is NOT team and the time has been exceeded.  I confirmed this operation.  This is not new functionality and `ExecuteTimer `has always set the color.  However, the implementation of how it is done IS new (using the interface).  The implementation of this method is required to provide a visual cue indicating time has been exceeded.  I will update the comment in the interface to reflect its actual functionality in more detail.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
johnbrvc: And I'm not sure why I even made that change - it was June of last year :-) I probably thought since it was a GUI thing it was more readable to a user to be "Auto Judging" instead of "AutoJudging" - "AutoJudgxxxx" is internal.  That's all I can think of.  I can put it back.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
johnbrvc: The handling of the `Terminate `/ `Start-Stop AJ `button is done in one place now.  If `notifyFrame `is non-null, it means that a run is taking place (the  execute timer is running), in which case, the button should only terminate the judging currently going on and not stop _Autojudging_ altogether (IE the `Terminate `button).  If `notifyFrame `is null, it means it is either between test cases or it has finished judging the last submission and there are no more submissions waiting (IE Stop Autojuding).  This all has to do with the fact that there is a single frame now, rather than a new (separate one) for each test case as well as one for the quiescent case (judge queue empty).   Does that make sense?

johnbrvc:
I merged latest develop into this branch.
No conflicts, strangely enough.  I tested it, it seems to (still) work as designed.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimer.java
```diff
@@ -116,61 +109,28 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
...
-     *
-     * @param arg1
-     *            java.awt.event.ActionEvent
-     */
-    private void connEtoC1(java.awt.event.ActionEvent arg1) {
+    public void executeFrameTerminated() {
         try {
-            this.btnTerminateActionPerformed(arg1);
+            if (doAutoStop) {
+                log.config("ExecuteTimer - User hit terminate while AJ'ing.");
```
clevengr, line 124: I don't understand why the log message above refers to "...while AJ'ing."  It seems to me that the action of invoking this method applies to both AutoJudging AND Manual Judging.  (The method is indeed called when a user presses the "Stop AutoJudging" button on a Judge which is enabled for AutoJuding, but it is _also_ called if a (human) judge presses the "Terminate" button during _manual_  judging.)  I think the message is misleading because it implies this is something tied specifically to use of an "AutoJudge".

Also, it seems to me that `log.config()` is the wrong logging method -- it implies that this message has something to do with "configuration", when the act of the ExecuteTimerFrame being terminated is a result of either the user hitting the "Terminate" button, the user hitting the `Stop AutoJudging` button,  or the system completing the submission execution (unless I've badly misunderstood what the PR changes are intending to do).  Assuming I'm correct, it seems like the logging call should be (something like) `log.info(...)`.  (Perhaps the use of `config()` was based on the concept that the method is being invoked when a user presses the "Stop AutoJudging" button -- but as noted above, from what I can tell that's not the _only_ thing which invokes this method/causes this log message.)

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
```diff
@@ -0,0 +1,237 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
clevengr, line 1: Copyright date should probably be 2023...

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+    
+    /**
+     * Return the ExecuteTimerLabel1 property value.
+     *
+     * @return javax.swing.JLabel
+     */
+    private JLabel getExecuteTimerLabel() {
+        if (ivjExecuteTimerLabel == null) {
+            try {
```
clevengr, line 144: The contents of this `try` block appear to be identical to the settings applied in the first `if` statement in method `resetFrame()`, above.  It seems like having two separate places setting the same values is just inviting a down-the-road issue where one of the places gets changed but the other does not, resulting in inconsistent GUI appearance/behavior.  It seems like the two code blocks should be encapsulated in a single method...

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+    
+    /**
+     * Return the TimerCount property value.
+     *
+     * @return javax.swing.JLabel
+     */
+    private JLabel getTimerCountLabel() {
+        if (ivjTimerCount == null) {
+            try {
```
clevengr, line 172: Same comment as above; the body of this `try` seems to be virtually identical to most of the (second) `if` statement in method `resetFrame()` and should probably be encapsulated into its own method.

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
clevengr, line 1: Second date in copyright should be 2023.

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
clevengr, line 1: Update second date in copyright notice.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
clevengr: You said:
> The handling of the `Terminate `/ `Start-Stop AJ `button is done in one place now. If `notifyFrame `is non-null, it means that a run is taking place (the execute timer is running), in which case, the button should only terminate the judging currently going on and not stop _Autojudging_ altogether (IE the `Terminate `button). If `notifyFrame `is null, it means it is either between test cases or it has finished judging the last submission and there are no more submissions waiting (IE Stop Autojuding). This all has to do with the fact that there is a single frame now, rather than a new (separate one) for each test case as well as one for the quiescent case (judge queue empty). Does that make sense?

Not yet, although I grant that it's possible it's my lack of understanding.  But here's what I see:
  
You said: `The handling of the Terminate / Start-Stop AJ button is done in one place now.` That may be, but those are _not_ the same thing.  The `Terminate` button is used by a (human) judge to terminate the execution of a test case which is taking too long (but the human judge doesn't want to just let it run until the runtime-limit is reached).  The `Start-Stop AJ` button is used to enable/disable the ability of a Judge client to automatically fetch/execute/validate runs.  Those are two different operations, and it's not clear to me how (or more correctly, why) the new code is trying to make them work the same (which is what it appears to me that it's trying to do).

Also, you said `and there are no more submissions waiting (IE Stop Autojuding)`.  Those are two separate (and relatively unrelated) conditions (that is, "`there are no more submissions waiting`" is a different condition from" `Stop AutoJudging`").  When "there are no more submissions waiting", the AJ simply waits, polling on occasion to find out if a new run has been submitted.  That is _not_ the same condition as "Stop AutoJudging"; the latter means "turn off the judge client's ability to poll for and to pull runs to be judged (i.e., switch to "manual, non-auto" judging mode).  So I'm still confused about what the code is trying to do.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
```diff
@@ -19,7 +22,7 @@
  */
 
 // $HeadURL$
-public class AutoJudgeStatusFrame extends javax.swing.JFrame implements AutoJudgeNotifyMessages {
+public class AutoJudgeStatusFrame extends javax.swing.JFrame implements AutoJudgeNotifyMessages, IExecuteTimerFrame {
```
clevengr, line 25: I am struggling to understand the overall design approach for handling the "execute timer frame".  Specifially, I don't understand the intent behind specifying that an `AutoJudgeStatusFrame implements IExecuteTimerFrame`, as stated in the above line. 

 In OOP parlance (including UML and Java, for instance),  "implements" is a form of _inheritance_; it specifies the "Is-A" relationship.  It doesn't make sense to me to say that `an AutoJudgeStatusFrame is a(n) IExecuteTimerFrame`.  

It appears to me that the relationship ought to be that of the OOP notion of _composition_ ("HAS-A"),  like that which is specified in the code for class `Executable`:  an `Executable`  HAS-A(n) `IExecuteTimerFrame`.  It is declared within the `Executable` class (as a private variable), and it is set by the parameter passed into the `Executable` constructor.  In the same way that it makes sense to say "An `Executable` has an `IExecuteTimerFrame`", it would make sense to say "An `AutoJudgeStatusFrame` HAS-AN `IExecuteTimerFrame`".   

Note that there is a class `ExecuteTimerFrame` which `implements IExecuteTimerFrame`.  This (sort of) makes sense; class `ExecuteTimerFrame` is a concrete implementation of the `IExecuteTimerFrame` interface.  I'm not sure it really makes sense to say the equivalent about class AutoJudgeStatusFrame.  Why does class `Executable` "have-an" `IExecuteTimerFrame` while class `AutoJudgeStatusFrame` "is-an" `IExecuteTimerFrame`?  It seems like there should be _one_ class (`ExecuteTimerFrame`) implementing `IExecuteTimerFrame`, and that one object should be passed to `AutoJudgeStatusFrame` just like it gets passed to `Executable`.  

Note that making such a design change also has the benefit that class `AutoJudgeStatusFrame` would no longer be (effectively) _duplicating_ the methods implemented in `ExecuteTimerFrame`; it would instead call the relevant methods (as needed) in its `IExecuteTimerFrame` member.

I grant that the code as given in the PR may "work" - meaning, it does what is intended from the user point of view.  However, I'm concerned that implementing what (IMHO) appears to be a weird design may in the long run cause us much anguish (when we get around to trying to fix other things, like the AJ execution cycle).  I would much rather see either a redesign at this point, or at least an explanation which counters my arguments above.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/ui/AutoJudgeStatusFrame.java
johnbrvc: Maybe I didn't explain it clearly.  Think of the `notifyFrame `member as an indication that a run is being executed at the moment if it is non-null.   In that case, the semantics of the `Terminate `button correspond to those of terminating a running submission.  If `notifyFrame `is null, it means that the AJ is not currently executing a submission, rather, it is sitting there waiting for more runs;  the user can now press the `Stop Autojudging` button.  Basically, the button is being "doubled-up" on depending on what is going on.  If executing a submission, it terminates the run.  If waiting for a submission, it stops autojudging.

`notifyFrame `is set by a call to `setTerminateButtonNotify(...)` from `ExecuteTimer `when a submission starts executing.  
It is set to **null** when the `ExecuteTimer `finishes (`ExecuteTimer.stopTimer(...)`) (at which point the `AutoJudgeStatusFrame `is again, waiting for new runs).

We can always have a google meet and trace through it if that helps.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimer.java
johnbrvc: Correct.  I guess I was in an "AJ" state of mind when writing this.  
log.config() changed to log.info().

Fixed in next push.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/execute/IExecuteTimerFrame.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: Agreed, but they are not exactly the same.  One sets the color back to the default, but, yes, the other stuff is the same.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: Agreed.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
```diff
@@ -0,0 +1,237 @@
...
+        super();
+        initialize();
+    }
+    
+    /**
+     * Return the ExecuteTimerFrame property value.
+     *
+     * @return javax.swing.JFrame
+     */
+    private void initialize() {
```
clevengr, line 43: The javadoc comment above this method doesn't make sense.  It says that the method returns a property value, but the method return type is `void`.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
clevengr: I sort of interpreted "Agreed" as meaning "I agree, and I'll fix it".  But I don't see any changes.  Are you agreeing that it should be changed but declining to change it?

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
clevengr: My question, as above, is:  do you intend to change this?  (I was hoping that since you "Agreed", it meant you'd pull the common functions out into a single method, call that method, and then have each block perform whatever additional steps are needed to account for the differences..)

src/edu/csus/ecs/pc2/core/execute/IExecutableMonitor.java
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+/**
+ * IExecutableMonitor interface
+ * This interface should be implemented by classes that use the Executable class.
```
clevengr, line 4: I don't know that the above is really an accurate statement; I don't think there's any imperative that a class which "uses the Executable class" necessarily NEEDS to use this interface; the purpose of the interface is to provide classes which use Executable with a mechanism to "monitor" the ongoing status of Executable operations -- but they don't HAVE to do that; they could simply invoke Executable methods and wait for them to complete without caring about monitoring ongoing status changes.  I think the comment is misleading, and/or incomplete in its description of the intention of the interface.

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+package edu.csus.ecs.pc2.core.execute;
+
+/**
+ * @author John Buck
+ *
+ */
+public interface IExecuteFrameNotify {
```
clevengr, line 8: The javadoc forthis interface should at least contain a brief summary of the purpose/intended usage of the interface.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: Agreed.  Will fix.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: This has been fixed and pushed.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: This has been fixed and pushed.

src/edu/csus/ecs/pc2/core/execute/IExecuteFrameNotify.java
johnbrvc: Fixed and pushed.

src/edu/csus/ecs/pc2/core/execute/IExecutableMonitor.java
johnbrvc: Fixed and pushed.

src/edu/csus/ecs/pc2/core/execute/ExecuteTimerFrame.java
johnbrvc: Fixed and pushed.

johnbrvc:
One enhancement still to implement for this PR is to stop the status message length from changing when it changes from Executing to Validating.  The appearance of the JFrame label seems to be irritating when it changes length since it is centered.

clevengr (approved):
I built a distribution from the latest push and tested submission of multiple runs.  I verified that both AutoJudging a run works as expected (the "AutoJudgeStatusFrame" displays the current Executable status, showing a "Terminate" button while execution is under way and then going back to showing a "Stop AutoJudging" button when execution finishes) and that Manual Judging works as expected (the "ExecuteTimerFrame" displays the current Executable status and shows a "Terminate" button).

I've noted the following things with regard to the current PR state:
- The "Executing/Validating" message jiggles back and forth as it changes values, which is annoying.  We've already had discussions about this and @johnbrvc has indicated he either plans to submit an update to fix this, or to file it as a separate issue.  It's trivial (but it would be nice for users if it was fixed).
- When the "Terminate" button is pushed (in either the AJ or Manual Judging situation), sometimes the "Compare" link on the Test Results screen shows "Output halted by Operator" for the team output (which is what is expected), but sometimes it instead shows (for a correct submission) output which is identical to the expected (judge's) output.  This happens even though the test case is (correctly) marked as "Failed".  This is probably due to (1) hitting the Terminate button after validation has completed but before the Executable instance finishes handling that test case, combined with (2) the fact that the PR _allows_ hitting the Terminate button (whereas before it was virtually impossible to do so).  This is likely NOT an issue with this PR -- but it should be filed as a separate issue (it'll be confusing to users to see output which looks correct but which the system marks as "Failed').
- When runs are judged (either AJ or Manual), the _Admin_ does not pick up the fact that the run state has changed (e.g. to "Correct" or "WA").  Rather the run remains as "BEING_JUDGED" on the Admin.  This again is probably NOT a fault of this PR; it is a separate issue.

Since none of the above are directly concerned with this PR, I'm approving it.

johnbrvc:
> 

Pushed fix and tested it on Windows and Linux.  Maintains correct width now - no jumping back and forth.

clevengr (approved):
I pulled the updated code, reviewed it, and built a new distribution which I used to run tests of both AutoJudging and Manual Judging.  

The fix for the "wiggling message" isn't exactly what I envisioned:  the left part of the message still jiggles (when the text changes from "Executing" to "Validating" -- the left edge of the text expands and contracts).  This is still a bit disturbing to the eye (although not as much as it was previously, where the entire message jumped around).  

I personally would prefer a solution where the left edge of the text remains fixed while at the same time the right end (showing the test case number) also remains fixed (for a given number of digits in the test case number).  It seems like it would be pretty simple to do this:  use monospaced font (as this change does), and then force the left side of the text string to always be 10 characters (padding both "Working" and "Executing" with blanks while adding no extra characters to "Validating"). It also seems like this would actually be _simpler_ than the existing solution of using `String.format()`.  

However, this perhaps boils down to (just) a personal preference, and perhaps a solution like I'm suggesting would actually end up being more complicated than I imagine.  I'm still ok with approving and merging the solution (we've got bigger things to focus on...)

johnbrvc:
> I pulled the updated code, reviewed it, and built a new distribution which I used to run tests of both AutoJudging and Manual Judging.
> 
> The fix for the "wiggling message" isn't exactly what I envisioned: the left part of the message still jiggles (when the text changes from "Executing" to "Validating" -- the left edge of the text expands and contracts). This is still a bit disturbing to the eye (although not as much as it was previously, where the entire message jumped around).
> 
> I personally would prefer a solution where the left edge of the text remains fixed while at the same time the right end (showing the test case number) also remains fixed (for a given number of digits in the test case number). It seems like it would be pretty simple to do this: use monospaced font (as this change does), and then force the left side of the text string to always be 10 characters (padding both "Working" and "Executing" with blanks while adding no extra characters to "Validating"). It also seems like this would actually be _simpler_ than the existing solution of using `String.format()`.
> 
> However, this perhaps boils down to (just) a personal preference, and perhaps a solution like I'm suggesting would actually end up being more complicated than I imagine. I'm still ok with approving and merging the solution (we've got bigger things to focus on...)

I went through a great deal of trouble (and testing) to make it fixed.  I tried on both Linux (X windows) and Windows and it is rock solid on my systems.  There is no wiggling at all.  The text for Executing/Validating changes in-place.  I'm not sure why you are seeing that.   Maybe it's something with the font(s) on your system? The font is monospaced.
The string is always a fixed length.  During debugging I had each message sent to a file and checked it; all the way to over 100 test cases, and it was correct. 

Maybe I don't understand what you mean by "wiggling".  It will alternate as follows:
```
 Executing test case 1...
Validating test case 1...
 Executing test case 2...
Validating test case 2...
...
 Executing test case 119...
Validating test case 119...

```
That is, the "`...ting test case #...`" stays in the same place all the time in the label (and on the pane).  
The V and E of `Validating `and `Executing `will not line up (that is by design).  If that is desired, that's just a **printf()** format:` %-10s` instead of `%10s`, but I didn't like the extra space coming and going between "`ting`" and "`test`".

clevengr:
> 
> Maybe I don't understand what you mean by "wiggling". It will alternate as follows:
> 
> ```
>  Executing test case 1...
> Validating test case 1...
>  Executing test case 2...
> Validating test case 2...
> ...
>  Executing test case 119...
> Validating test case 119...
> ```
> 
> That is, the "`...ting test case #...`" stays in the same place all the time in the label (and on the pane). The V and E of `Validating `and `Executing `will not line up (that is by design). If that is desired, that's just a **printf()** format:` %-10s` instead of `%10s`, but I didn't like the extra space coming and going between "`ting`" and "`test`".

Yes, that's what I meant by "wiggling" -- the "E" and "V" jumping back and forth on the left side of the message.  I thought it would look better if the "E" and "V" were always in the same place, and there was just an extra blank after the word "Executing " which wasn't there when the message was "Validating".  Something like this:
```java
 Executing  test case 1...
 Validating test case 1...
 Executing  test case 2...
 Validating test case 2...
...
 Executing  test case 119...
 Validating test case 119...
```
I even thought about maybe something like this:
```java
 Executing.... test case 1...
 Validating... test case 1...
 Executing.... test case 2...
 Validating... test case 2...
...
 Executing.... test case 119...
 Validating... test case 119...
```
(Notice there's an extra "." after Executing...)

It just seemed awkward to me to have the beginning character(s) of the string jumping around.  But I didn't actually go to the trouble of altering the code to see whether such options_really_ look better, and anyway as I said, it's possible it's just "personal preference".   I'm ok with any of the above (including the one that's checked in :) )

lane55 (approved):
approve changes

PR #727, opened 2023-04-04
I46 add entry point support, fixes #46

### LINKED ISSUE #46: Add entry_point processing when read from event feed (CLICS and Shadow implementation)
**Feature Description**: 

For shadow and in the contest model for a Run, add an entry point.

The JSON CLICS EF from a primary supplies the entry point 
for the team's submission.   PC2 solved this in
a different way by the user selecting the entry point
file.  Using the EF the primary CCS specifies.

This entry point would be used in Executable class to
run/execute the correct program/class if there is
more than one team submission file (in a Run)

Yes. N/A

Add entry point char [] field in Run.
When executing if entry point defined use that as the pc2 main file.

**Additional context**: 

none.

johnbrvc:
Obeys the **CLICS** spec for the event feed submissions object when an `entry_point `property is present.  The `entry_point`, if specified on the event feed, replaces the main source file basename when substituting the `{:basename}` substitution variable.  This allows submissions that are in a java package to work properly, in addition to Kotlin submissions.  

Also added a new _conditional_ substitution variable, `{:ensuresuffix=_xxx_}`, that provides a mechanism to conditionally add text to an execute command (Executable.java).   This is necessary, for example, if PC2 is shadowing and the event feed provides an `entry_point ` for a submission, but PC2 _also_ accepts submissions from **WTI** or the _fat-team_ client.  If this is a Kotlin submission (for example), and the event feed provides a valid `entry_point` of, say, **MyClassKt**, and a submission is made to PC2 via WTI or the fat-team client, we want to be able to add the trailing **Kt** in the execute command in the latter case but not the former case.  A string such as: `{:basename}{:ensuresuffix=Kt}` could be used in this case, and it will work for both event feed submissions AND WTI/fat-client submissions.

#46 

Windows 11/Linux Ubuntu 22.04
Java 1.8.0_321

This is very complicated to test as it requires an event feed (shadowing) that contains submissions that have `entry_point`'s specified.
Assuming that you have a contest set up and you have access to a shadow event feed that has submissions with `entry_point`'s that are something OTHER than the main source file _basename_ (such as ANY Kotlin submission), the process is straight forward.
1) Define the language execute string use `{:basename}{:ensuresuffix=Kt}` 
2) Connect to the shadow server (primary)
3) Wait for a Kotlin run to be submitted on the event feed that correctly specifies an `entry_point`.
4) Verify that the run executes properly.  That is, that it correctly used the supplied entry_point and got a valid (expected) judgement.
5) Attempt to re-judge the submission from an administrator account (for example).
6) Verify that the run executes properly (this tests the retention of the supplied `entry_point`)
7) Make a Kotlin submission using the _fat-client_ (see Note 1 below).
8) Verify that it executes properly and gives a valid (expected) judgement .

Note 1: Kotlin has a rather unique way of naming it's classes.  If the source file is named **test-hello.kt**, the main class produced by the Kotlin compiler is: **Test_helloKt**   If the source is named **Test-hello.kt**, the same main class is produced.  If the source is named **Test_hello.kt**, again, **Test_helloKt** is the main class name produced.  So, all three of those source files produce the same main class name (and consequently, the same **Test_helloKt.class** file).  Of course, this means if the source file is case and hyphen different from the main class file produced, then a **WTI**/_fat-client_ submission will not work.  It will only work if the main source file is case/underscore identical to the main class name produced.  Keep this in mind when testing.

Note 2: For submissions in java where the source files are all part of a user defined package, an entry point is required on the event feed, and this works fine for submissions received on the event feed using this PR.  There is (currently) no way, however, for a user to specify the entry point from **WTI** or the _fat-client_, as such, there is no way to test this case.

Note 3: Thanks to @lane55 for providing a starting code base for this PR.  Any mistakes are mine, not his!
 
A **JUnit** test was added to test the new conditional string substitution code.

The image below shows the language configuration for Kotlin using the new conditional substitution variable.

langpage

If you need assistance setting up a test environment, @johnbrvc can provide guidance on how testing was performed, and help you set up the same contest and environment.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/Constants.java
+++ b/src/edu/csus/ecs/pc2/core/Constants.java
@@ -109,11 +109,11 @@ public final class Constants {
      * This can be used for languages (such as Kotlin), that has a different main class name than the basename of
      * the source.  eg.  MyClass.kt  uses MyClassKt as the main class name.  This is conditional in that if the
      * trailing string (suffix) isn't already immediately to the left of the operand of this substitution variable, then it gets added.
-     * eg. {:basename}{:ifsuffix=Kt}  if {:basename} is MyClassKt then, the {:ifsuffix=Kt} is a no-op (empty string).  If
-     * {:basename} is MyClass then the {:ifsuffix=Kt} evaluates to "Kt" making the result: MyClassKt.
+     * eg. {:basename}{:ensuresuffix=Kt}  if {:basename} is MyClassKt then, the {:ensuresuffix=Kt} is a no-op (empty string).  If
+     * {:basename} is MyClass then the {:ensuresuffix=Kt} evaluates to "Kt" making the result: MyClassKt.
      * Use of this is not just restricted to executables or basename; it can be used anywhere in the string.
      */
-    public static final String CMDSUB_COND_SUFFIX = "{:ifsuffix=}";
+    public static final String CMDSUB_COND_SUFFIX = "{:ensuresuffix=}";
     
     /**
      * Default file name for judgement ini file.
--- a/src/edu/csus/ecs/pc2/core/Constants.java
+++ b/src/edu/csus/ecs/pc2/core/Constants.java
@@ -100,6 +100,22 @@ public final class Constants {
     
     
     /**
+     * Command substitution variable for name of thing to execute (basename of source file)
+     */
+    public static final String CMDSUB_BASENAME_VARNAME = "{:basename}";
+    
+    /**
+     * Command substitution variable for conditional suffix
+     * This can be used for languages (such as Kotlin), that has a different main class name than the basename of
+     * the source.  eg.  MyClass.kt  uses MyClassKt as the main class name.  This is conditional in that if the
+     * trailing string (suffix) isn't already immediately to the left of the operand of this substitution variable, then it gets added.
+     * eg. {:basename}{:ifsuffix=Kt}  if {:basename} is MyClassKt then, the {:ifsuffix=Kt} is a no-op (empty string).  If
+     * {:basename} is MyClass then the {:ifsuffix=Kt} evaluates to "Kt" making the result: MyClassKt.
+     * Use of this is not just restricted to executables or basename; it can be used anywhere in the string.
+     */
+    public static final String CMDSUB_COND_SUFFIX = "{:ifsuffix=}";
+    
+    /**
      * Default file name for judgement ini file.
      */
     public static final String JUDGEMENT_INIT_FILENAME = "reject.ini";
--- a/src/edu/csus/ecs/pc2/core/NullController.java
+++ b/src/edu/csus/ecs/pc2/core/NullController.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core;
 
 import java.io.IOException;
@@ -540,6 +540,12 @@ public class NullController implements IInternalController {
 
     @Override
     public void submitRun(ClientId submitter, Problem problem, Language language, SerializedFile mainSubmissionFile, SerializedFile[] additionalFiles, long overrideTimeMS, long overrideRunId) {
+        // TODO Auto-generated method stub
+        
+    }
+
+    @Override
+    public void submitRun(ClientId submitter, Problem problem, Language language, String entry_point, SerializedFile mainSubmissionFile, SerializedFile[] additionalFiles, long overrideTimeMS, long overrideRunId) {
         // TODO Auto-generated method stub
         
     }
--- a/src/edu/csus/ecs/pc2/core/NullController.java
+++ b/src/edu/csus/ecs/pc2/core/NullController.java
@@ -534,25 +534,21 @@ public class NullController implements IInternalController {
 
     @Override
     public IInternalContest getContest() {
-        // TODO Auto-generated method stub
         return null;
     }
 
     @Override
     public void submitRun(ClientId submitter, Problem problem, Language language, SerializedFile mainSubmissionFile, SerializedFile[] additionalFiles, long overrideTimeMS, long overrideRunId) {
-        // TODO Auto-generated method stub
         
     }
 
     @Override
     public void submitRun(ClientId submitter, Problem problem, Language language, String entry_point, SerializedFile mainSubmissionFile, SerializedFile[] additionalFiles, long overrideTimeMS, long overrideRunId) {
-        // TODO Auto-generated method stub
         
     }
 
     @Override
     public void submitJudgeRun(Problem problem, Language language, SerializedFile mainFile, SerializedFile[] otherFiles, long overrideSubmissionTimeMS, long overrideRunId) throws Exception {
-        // TODO Auto-generated method stub
         
     }
 
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -1818,7 +1818,7 @@ public class Executable extends Plugin implements IExecutable {
              * We only do this substition for execution of the run, not compiles or validators, so it can't
              * be done by substitutionAllStrings without refactoring.
              */
-            if (run.getEntryPoint() != null) {
+            if (!StringUtilities.isEmpty(run.getEntryPoint())) {
                 // change Constants.CMDSUB_BASENAME_VARNAME to entry_point rather than basename from {:mainfile} before
                 // other substitutions (overrides :mainfile)
                 cmdline = replaceString(cmdline, Constants.CMDSUB_BASENAME_VARNAME, run.getEntryPoint());
@@ -2193,7 +2193,7 @@ public class Executable extends Plugin implements IExecutable {
             String programName = language.getExecutableIdentifierMask();
             
             // the "executable" program name is the entry point, if one exists, so try to substitute that first
-            if (run.getEntryPoint() != null) {
+            if (!StringUtilities.isEmpty(run.getEntryPoint())) {
                 // change Constants.CMDSUB_BASENAME_VARNAME to entry_point rather than basename from {:mainfile} before
                 // other substitutions (overrides :mainfile)
                 programName = replaceString(programName, Constants.CMDSUB_BASENAME_VARNAME, run.getEntryPoint());
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -1812,6 +1812,15 @@ public class Executable extends Plugin implements IExecutable {
             }
 
             log.log(Log.DEBUG, "before substitution: " + cmdline);
+
+            /**
+             * Special substitution for entry_point
+             */
+            if (run.getEntryPoint() != null) {
+                // change {:basename} to entry_point rather than basename from {:mainfile} before
+                // other substitutions (overrides :mainfile)
+                cmdline = replaceString(cmdline, "{:basename}", run.getEntryPoint());
+            }
             cmdline = substituteAllStrings(run, cmdline, dataSetNumber+1);
             log.log(Log.DEBUG, "after  substitution: " + cmdline);
 
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -2438,72 +2438,6 @@ public class Executable extends Plugin implements IExecutable {
 
         return buf.toString();
     }
-
-    /**
-     * Perform conditional subtititions
-     * 
-     * @param origString
-     * @param condVar
-     * @return - possibly modified string
-     */
-    public String replaceStringConditional(String origString, String condVar) {
-        
-        int subStartIdx, endIdx, suffLen;
-        String suffString;
-        
-        if (origString == null || condVar == null) {
-            return origString;
-        }
-        // if substitute variable has an = sign, we try to match the substring up to and including the
-        // equals sign, eg: {:ifsuffix=Kt}, we want to match: {:ifsuffix=
-        int subIdx = condVar.indexOf('=');
-        
-        // Need at least two chars to left of =
-        if(subIdx < 2) {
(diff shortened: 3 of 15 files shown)


johnbrvc:
While doing additional testing with the event feed and some arcane cases, a problem was found using Kotlin if the source file base name case/dash compare does not match the produced "**Executable Filename**".  Ex.  `mainKt.class` != `MainKt.class`.  PC2 will indicate a "Compile error" even if the compile works, unless the "Script or Interpretted Language" checkbox is marked (in which case it works fine).  This is because the `{:basename} `substitution variable only uses "**entry_point**" for the "**Program Execution Command Line**"; it should also use it for the "**Executable Filename**" but it does not.  This was an oversight in the coding, and I will fix it.

johnbrvc:
> While doing additional testing with the event feed and some arcane cases, a problem was found using Kotlin if the source file base name case/dash compare does not match the produced "**Executable Filename**". Ex. `mainKt.class` != `MainKt.class`. PC2 will indicate a "Compile error" even if the compile works, unless the "Script or Interpretted Language" checkbox is marked (in which case it works fine). This is because the `{:basename} `substitution variable only uses "**entry_point**" for the "**Program Execution Command Line**"; it should also use it for the "**Executable Filename**" but it does not. This was an oversight in the coding, and I will fix it.

This has been addressed with a new commit to this PR.

test/edu/csus/ecs/pc2/core/execute/ExecutableTest.java
```diff
@@ -1198,6 +1198,42 @@ public void testSubstituteFiles() throws Exception {
...
+        
+        RunFiles runFiles = new RunFiles(run, sumitFilename);
+        
+        Executable executable = new Executable(contest, controller, run, runFiles);
+        executable.setExecuteDirectoryName(executeDirectoryName);
+        executable.setUsingGUI(false);
+        
+        String actual = executable.substituteAllStrings(run, "foo {:basename}{:ifsuffix=it}");
+        
+        assertEquals("Expected substitution", actual, "foo ISumit");
```
lane55, line 1227: these assertEquals parameters are backwards... they are ordered "Expected" actualStr, expectedStr
All the other assertEquals are ordered "Expected ", expectedStr, actualStr

test/edu/csus/ecs/pc2/core/log/NullController.java
```diff
@@ -654,6 +654,13 @@ public void submitRun(ClientId submitter, Problem problem, Language language, Se
         
     }
 
+    @Override
+    public void submitRun(ClientId submitter, Problem problem, Language language, String entry_point, SerializedFile mainFile, SerializedFile[] auxFiles, long overrideTimeMS,
+            long overrideSubmissionID) {
+        // TODO Auto-generated method stub
```
lane55, line 660: If there is nothing to do/follow up on, remove TODO statement

src/edu/csus/ecs/pc2/core/NullController.java
```diff
@@ -544,6 +544,12 @@ public void submitRun(ClientId submitter, Problem problem, Language language, Se
         
     }
 
+    @Override
+    public void submitRun(ClientId submitter, Problem problem, Language language, String entry_point, SerializedFile mainSubmissionFile, SerializedFile[] additionalFiles, long overrideTimeMS, long overrideRunId) {
+        // TODO Auto-generated method stub
```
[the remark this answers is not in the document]
johnbrvc: Fixed.

test/edu/csus/ecs/pc2/core/log/NullController.java
johnbrvc: Fixed.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2414,6 +2425,72 @@ public String replaceString(String origString, String beforeString, String after
...
     }
 
+    /**
+     * Perform conditional subtititions
+     * 
+     * @param origString
+     * @param condVar
+     * @return - possibly modified string
+     */
+    public String replaceStringConditional(String origString, String condVar) {
```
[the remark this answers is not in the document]
johnbrvc: Fixed, however.... it seems that ExecuteUtilities has many routines in it that are duplicated in Executable, and none of them are being used by anyone except the JUnit tests.  (There are a couple of unique routines dealing with directories and files that ARE used in ExecuteUtilities, but string manipulation routines (replaceString, etc.) are never used anywhere and are duplicated.  This should be fixed someday. *sigh*  I think part of the issue is that in order to use some of the replace/substitute routines in ExecuteUtilities, it requires instantiating an ExecuteUtilities object, and all the baggage that goes along with that.  If the methods are static, it's easy, but some are not.

lane55 (approved):
Approved - all fixes done.

clevengr:
The PR description contains two references like `the {:basefile} substitution variable`.  I don't think there's any such substitution variable as `basefile`; I think these are supposed to be references to `basename`.

Obviously this doesn't affect the correctness of the _code_, but it seems like it would be good to have the PR contain an accurate description just in case we come back to it in the future.

src/edu/csus/ecs/pc2/core/Constants.java
```diff
@@ -99,6 +99,22 @@ private Constants() {
...
+    /**
+     * Command substitution variable for conditional suffix
+     * This can be used for languages (such as Kotlin), that has a different main class name than the basename of
+     * the source.  eg.  MyClass.kt  uses MyClassKt as the main class name.  This is conditional in that if the
+     * trailing string (suffix) isn't already immediately to the left of the operand of this substitution variable, then it gets added.
+     * eg. {:basename}{:ifsuffix=Kt}  if {:basename} is MyClassKt then, the {:ifsuffix=Kt} is a no-op (empty string).  If
+     * {:basename} is MyClass then the {:ifsuffix=Kt} evaluates to "Kt" making the result: MyClassKt.
+     * Use of this is not just restricted to executables or basename; it can be used anywhere in the string.
+     */
+    public static final String CMDSUB_COND_SUFFIX = "{:ifsuffix=}";
```
clevengr, line 116: I find the chosen substitution variable name (`ifsuffix`) to be confusing relative to what is actually being done.  The use of "if" in the name makes it sound like it's something like "if (the suffix is equal to xxx) then add xxx" when in fact it's just the opposite:  what it's doing is saying "if (the suffix is NOT equal to xxx) then add xxx".   This was one of the reasons I was so confused about what the new conditional variable was doing -- i.e., what its purpose was.

A more accurate description of the action being performed would be something like `{:ensuresuffix=}`, which is a lot closer to what is actually being done.   This is a bit longer -- but it's no longer than other substitution variables we already have (e.g. `problemshort`, `languagename`), and it's (IMHO) a much better description of what's being done.   I suggest replacing `{:ifsuffix=}` with `{:ensuresuffix=}`.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1812,6 +1812,17 @@ protected boolean executeProgram(int dataSetNumber) {
             }
 
             log.log(Log.DEBUG, "before substitution: " + cmdline);
+
+            /**
+             * Special substitution for entry_point.
+             * We only do this substition for execution of the run, not compiles or validators, so it can't
+             * be done by substitutionAllStrings without refactoring.
+             */
+            if (run.getEntryPoint() != null) {
```
clevengr, line 1821: The 2022-07 CLICS spec defines the type of the "entry point" property as `string ?` (meaning a string which might be null).  The spec further gives explicit examples of what must be provided as the entry point for Python (the name of the file which must be run), and for Java and Kotlin (the fully-qualified class name, including any package prefixes).  The spec further explicitly states that "For C and C++ it must be `null`".  However, the spec is silent about what 
must/may/should/should not be provided for the entry point for any _other_ language.  I believe that it's possible that an implementer could legitimately, for example, pass in the _empty string_ as an indication that "no entry point is specified" for some (other) language.  

If that were to happen, I believe the above test would cause PC2 to fail -- because it would NOT see a "null" value returned from `run.getEntryPoint()` (because "the empty string" is not the same as "null") -- so it would (improperly) replace the main program command name with an empty string.   An entry point which was not null and also not empty but contained only whitespace would cause a similar problem.

For defensive programming (i.e., to avoid troubles down the road) I think the above test should be something like 
```java
String ep = run.getEntryPoint() ;
if ( ep != null  &&  !ep.isEmpty()  &&  !ep.isBlank() ) {
...  //do the substitution
```

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2178,13 +2189,27 @@ protected boolean compileProgram() {
 
             packageName = "";
             packagePath = "";
-            String programName = replaceString(language.getExecutableIdentifierMask(), "{:basename}", removeExtension(runFiles.getMainFile().getName()));
+            
+            String programName = language.getExecutableIdentifierMask();
+            
+            // the "executable" program name is the entry point, if one exists, so try to substitute that first
+            if (run.getEntryPoint() != null) {
```
clevengr, line 2196: Same comment as above; I think this should also test for `!isBlank()` and `!isEmpty()`.

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
```diff
@@ -52,6 +52,22 @@ public RemoteRunSubmitter(IInternalController controller) {
...
+        submitRun(clientIdString, problemID, languageID, null, mainFile, auxFiles, overrideTimeMS, overrideRunId);
+    }
+    
+    /**
+     * Submit a run to the server.
+     * 
+     * @param clientIdString - "teamN" client name
+     * @param problemID - problem id
+     * @param languageID - CLICS language id
+     * @param entry_point - java/kotlin entry point, null if default or none
```
clevengr, line 64: I don't particularly care for the fact that the above line explicitly refers to "java/kotlin", when according to the CLICS spec it could be non-null for ANY language other than Python, C, or C++.

Not something I'd deny Approval for; just a comment.

test/edu/csus/ecs/pc2/core/execute/ExecutableTest.java
```diff
@@ -1196,7 +1196,43 @@ public void testSubstituteFiles() throws Exception {
         
         String actual = executable.substituteAllStrings(run, "foo {files}");
         
-        assertEquals("Expected substitution", actual, "foo ISumit.java");
+        assertEquals("Expected substitution", "foo ISumit.java", actual);
+    }
+    
+    /**
+     * test {:ifsuff=xxx} substitution.
```
clevengr, line 1203: The above line should be updated to reflect the new name of the substitution variable (`{:ensuresuffix=xxx}`)

test/edu/csus/ecs/pc2/core/execute/ExecuteUtilitiesTest.java
```diff
@@ -538,4 +538,29 @@ public void testmatchIncludeReNegative() throws Exception {
         }
     }
 
+    /**
+     * test {:ifsuff=xxx} substitution.
```
clevengr, line 542: Update substitution variable name (above).

test/edu/csus/ecs/pc2/core/log/NullController.java
clevengr: I'd just like to point out that there are currently **_107_** methods in just this class that contain the `//TOD Auto-generated method stub` comment.  Seems a bit picky to ask for _one_ of them to be removed...

src/edu/csus/ecs/pc2/core/execute/Executable.java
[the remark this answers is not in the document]
lane55: re: dupicate methods.  Yes, the intention was to replace the dups with static methods from ExecuteUtilities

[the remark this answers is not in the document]
lane55: ExecuteUtilties should never need to be instanciated.   All new and existing methods should be static to
avoid any baggage.

test/edu/csus/ecs/pc2/core/log/NullController.java
lane55: I disagee that it is picky to remove them, it is an opportunity to remove unsed TODO items.   The alternative as cleanup is to review every single //TODO Auto-generated, if we can reduce that or provide an actual TODO then there will be less work for the team to do later.
Further removing auto generated TODO statements should be considered a habit, in order to remove them when one knows their context.   Again less work later to delete/update one line.

src/edu/csus/ecs/pc2/core/execute/Executable.java
lane55: Note: there is no isBlank method in Java 8

lane55: I recommend using StringUtilities.isBlank(s), that way when isBlank is available that change can be reflected

test/edu/csus/ecs/pc2/core/execute/ExecutableTest.java
johnbrvc: Fixed.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: Fixed.  Used @lane55's recommendation of using StringUtilities.isEmpty().

src/edu/csus/ecs/pc2/shadow/RemoteRunSubmitter.java
johnbrvc: Agreed.  Changed comment.

clevengr:
> The PR description contains two references like `the {:basefile} substitution variable`. I don't think there's any such substitution variable as `basefile`; I think these are supposed to be references to `basename`.
> 
> Obviously this doesn't affect the correctness of the _code_, but it seems like it would be good to have the PR contain an accurate description just in case we come back to it in the future.

I went ahead and fixed these things myself (they only affect the PR description text, not anything in the PR code).

clevengr (approved):
After completing a review of all the code files which changed, I then pulled the code from JohnB's fork and tested it using the procedures described in [this document](https://docs.google.com/document/d/1iJZ-_y0q4-8m4-y7akp1oMyHVQll_3JFAeXLB8J-6CA/edit).  After many days of pulling out what little hair I have left, I was able to verify that both Java and Kotlin programs, given the correct language definitions, correctly use the "entry_point" attribute in order to run the program if such an attribute exists in the Event Feed submission.  I believe the PR correctly does what is intended, and I'm approving it.

PR #759, opened 2023-05-27
i758: in shadow, add message when run cannot be submitted to list box

### LINKED ISSUE #758: Shadow, when run cannot be submitted silently fails - should show error in listbox
**Describe the issue**: 

When shaodow cannot submit a run - silently fails.

In NAC23, when the domjudge account was not configured, all runs from domjudge (more than 20 runs)
silently (no GUI indication of a failure) failed to submit   Each failure was logged but we had to debug to find out
why the runs did not submit.

**To Reproduce**: 

Submit runs that will fail (no such account, no such language, etc.) on pc2 server.

**Expected behavior**: 

For each submission a failure a message should appear on the shadow connection list box.

**Actual behavior**: 

Silently logs error in submitting runs.

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**:

lane55:

When shadow cannot submit run, show error message in shadow UI (connection) list box

Fixes #758

Primary sends run to shadow and the submitting user account does not exist
on pc2 server, ex domjudge.

### DIFF
--- a/src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
+++ b/src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
@@ -26,6 +26,8 @@ import edu.csus.ecs.pc2.core.model.ClientType.Type;
 import edu.csus.ecs.pc2.core.model.IFile;
 import edu.csus.ecs.pc2.core.model.RunUtilities;
 import edu.csus.ecs.pc2.ui.ShadowCompareRunsPane;
+import edu.csus.ecs.pc2.ui.ShadowControlPane;
+import edu.csus.ecs.pc2.ui.ShadowControlPane.ShadowStatus;
 
 /**
  * This class listens for submission events in the event feed input stream from a remote CCS CLICS Contest API Event Feed.
@@ -491,6 +493,7 @@ public class RemoteEventFeedMonitor implements Runnable {
                                                     submitter.submitRun(runSubmission.getTeam_id(), runSubmission.getProblem_id(), runSubmission.getLanguage_id(),
                                                             runSubmission.getEntry_point(), mainFile, auxFiles, overrideTimeMS, overrideSubmissionID);
                                                 } catch (Exception e) {
+                                                    ShadowControlPane.addToConnectTableEntry(ShadowStatus.FAILURE, "Unable to submit run "+overrideSubmissionID+" "+ e.getMessage());
                                                     // TODO design error handling reporting
                                                     logAndDebugPrint(log, Level.WARNING, "Exception submitting run for event: " + event, e);
                                                 }
--- a/src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
@@ -117,9 +117,9 @@ public class ShadowControlPane extends JPanePlugin implements IShadowMonitorStat
     
     private JTextField lastEventTimeTextField;
     
-    private JTableCustomized connectStatusTable;
+    private static JTableCustomized connectStatusTable;
     
-    private DefaultTableModel connectStatusTableModel;
+    private static DefaultTableModel connectStatusTableModel;
     
     private int statusScrollBarMax = 0;
     
@@ -127,7 +127,7 @@ public class ShadowControlPane extends JPanePlugin implements IShadowMonitorStat
     
     private String lastToken = null;
     
-    private SimpleDateFormat lastDateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+    private static SimpleDateFormat lastDateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
     
     // Lightish green for success
     private Color statusColorSuccess = new Color(128, 255, 128);
@@ -135,7 +135,7 @@ public class ShadowControlPane extends JPanePlugin implements IShadowMonitorStat
     private Color statusColorFailure = new Color(255, 128, 128);
 
     // Status column for JTable notifications
-    enum ShadowStatus {
+    public enum ShadowStatus {
         SUCCESS,
         FAILURE,
         INFO
@@ -678,7 +678,7 @@ public class ShadowControlPane extends JPanePlugin implements IShadowMonitorStat
         resizeColumnWidth(connectStatusTable);
     }
     
-    private void resizeColumnWidth(JTableCustomized table) {
+    private static void resizeColumnWidth(JTableCustomized table) {
         SwingUtilities.invokeLater(new Runnable() {
             public void run() {
                 TableColumnAdjuster tca = new TableColumnAdjuster(table, HORZ_PAD);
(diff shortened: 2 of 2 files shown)


src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -987,6 +987,34 @@ public void run() {
         
     }
     
+    public static void addToConnectTableEntry(ShadowStatus stat, String msg)
```
johnbrvc, line 990: Since the code body of `addToConnectTableEntry() `is the same as `addConnectTableEntry()`, it probably makes sense to make `addConnectTable ``public static`, remove `addToConnectTableEntry`, and change the reference to `addToConnectTableEntry `in `RemoteEventFeedMonitor `to `addConnectTableEntry`.

lane55:
There are a couple try/catch that silently fail, will add a method call to add those exceptions/info to 
the Shadow connnection status list

lane55:
CI described above to improve error reporting to the user will add MessageManager.fireMessageListener to
show messages to users, for the following catch blocks
```
  } catch (IllegalContestState | JAXBException | IOException e) {
            log.log(Log.WARNING,"Exception creating PC2 scoreboard JSON: " + e.getMessage(), e);

                    } catch (MalformedURLException e) {
            log.log(Level.WARNING, "Malformed Remote CCS URL: \"" + remoteCCSURLString + "\" ", e);
```

lane55:
An intitial prototype was created and tested which used a poor design,
lots of changing fields/methods to static so they could be invoked with
as little change to existing code.

This prototype cdoe was removed and replaced with new code using
an Observable pattern.

lane55:
re: the current implementation/PR.

A goal for sending messages to the Shadow test connection status list is
to provide a way to add a single statement to send a message to the status list.

An observable pattern with a listener list was used.  The class MessageManager
provides method to add and delete listeners and to "fire" send messges (class MessageRecord)
to that list of listeners.   

A listener was implemented in ShadowControlPane using MessageManager.addMessageListener(...)
that will add the row to the Shadow test connection status list.

With this infrastructure a single statment can be used to send/add a message to the Shadow status list.
MessageManager.fireMessageListener(...)
Ex from code. 
```MessageManager.fireMessageListener(new MessageRecord("Unable to submit run " + overrideSubmissionID + " " + e.getMessage(), MessageScope.SHADOW_UI, e));```

**Comment by lane55** [comment]
The MessageManager.fireMessageListener(...) was added to a number of locations where shadow processing throws an Exception

File: src/edu/csus/ecs/pc2/ui/ShadowControlPane.java (hunk 1 above)
Comment (lane55) [reply src/edu/csus/ecs/pc2/ui/ShadowControlPane.java]: pushed fix, removed 'static' (and some other unneeded methods)

File: src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java [hunk 2]
```diff
@@ -25,7 +25,12 @@
 import edu.csus.ecs.pc2.core.model.ClientType.Type;
 import edu.csus.ecs.pc2.core.model.IFile;
 import edu.csus.ecs.pc2.core.model.RunUtilities;
+import edu.csus.ecs.pc2.ui.MessageManager;
+import edu.csus.ecs.pc2.ui.MessageRecord;
+import edu.csus.ecs.pc2.ui.MessageScope;
 import edu.csus.ecs.pc2.ui.ShadowCompareRunsPane;
+//import edu.csus.ecs.pc2.ui.ShadowControlPane;
```
johnbrvc, line 32: These should be removed.  The history is always available using GIT.  Leaving old code, just makes things confusing.

johnbrvc (changes requested):
I have tested this PR by forcing bad submissions, and it seems to work.
The new message observable seems to work fine.
I'd like to see my comments on some of the code addressed (very minor changes).

src/edu/csus/ecs/pc2/shadow/RemoteEventFeedMonitor.java
lane55: removed, pushed fix

src/edu/csus/ecs/pc2/ui/ShadowControlPane.java
```diff
@@ -960,7 +974,7 @@ protected static Map<String, Object> getMap(String jsonString) {
         }
     }
 
-    private void addConnectTableEntry(ShadowStatus stat, String msg)
+    public void addConnectTableEntry(ShadowStatus stat, String msg)
```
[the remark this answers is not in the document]
lane55: no. pushed fix/change

johnbrvc (approved):
Requested changes made.

src/edu/csus/ecs/pc2/ui/MessageManager.java
```diff
@@ -0,0 +1,48 @@
...
+
+
+    /**
+     * Send message to all listeners.
+     * 
+     * @param record - message
+     */
+    public static void fireMessageListener(MessageRecord record) throws Exception {
+        for (int i = 0; i < messageListenerList.size(); i++) {
+            messageListenerList.elementAt(i).messageAdded(record);
```
clevengr, line 41: Since a `MessageRecord` already has a defined `MessageScope`, it seems like this method should only add messages to listeners interested in messages of the appropriate scope.

I don't consider this a reason to not approve the PR, but it's one of those things that seems pretty simple to fix now, while the new code is fresh in the mind... whereas the alternative will be the "Someday" that never comes...

clevengr (commented):
I have completed a review of the PR code and while I made a couple of comments/suggestions, I don't have any requested changes that would block approval.   
I'm now proceeding to a runtime test.

src/edu/csus/ecs/pc2/ui/MessageManager.java
lane55: If there were a someday task here, I would agree.   There are a number of potential future "later" plans to improve the users'   ability to address both the minor and significant conditions, this change serves as a basis for a message system that will both cleanup  existing less than adequate/helpful error handling and provide for better insight/tasks for users to do to handle mis-configuration, security and other items.     

Adding MessageScope provides a way to add new messaging without accidentially sending/showing existing messages to the users at the "wrong" place in the GUI.

clevengr: `If there were a someday task here, I would agree. `  I have no idea what this means.  My point was that the current code already defines `MessageScope`, and it already has code which inserts messages in Message Listeners.  It seems to me it should only insert those messages if they are in the scope which the listener is interested in... and that this would be a simply thing to add "now" instead of waiting for "someday".

But again, I don't think this rises to the level of prohibiting Approval.  I just hate to see things coded "almost correctly" when it appears to be a very minor amount of effort to do it "correctly".

clevengr (approved):
I've tested the code using a remote (Primary) CCS provided by John B (thanks), and I've verified that submissions from remote teams which do not exist in the PC2 Shadow system do indeed result in errors being displayed in the Shadow GUI.  Since that was the essence of the original Issue, I approve the PR.

lane55:
Two approvals merging PR

PR #770, opened 2023-06-01
I 763 add clar tab to shadow feeder frame

### LINKED ISSUE #763: Add Clarifications tab to Shadow Frame

It is helpful to be able to easily view clars while
monitoring shadowing.

**Feature Description**: 

Add the Clarifications tab to the Shadow Frame.

n/a

Do per description.

**Additional context**:

lane55:

Adds Clarificaiton tab to Feeder UI.

Also ensures that the shadow/feeder can NOT answer clars.

fixes #763 

java version "1.8.0_121"

start server using mini contest 
```
pc2server --login s --contestpassword contest  --load mini
```

start feeder
```
pc2ef --login ef1
```

Passes if:
1. Clarifications tab present
2. There is no Answer button on the Clarifications pane

### DIFF
--- a/src/edu/csus/ecs/pc2/core/PermissionGroup.java
+++ b/src/edu/csus/ecs/pc2/core/PermissionGroup.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core;
 
 import edu.csus.ecs.pc2.core.model.ClientType;
@@ -142,7 +142,6 @@ public class PermissionGroup {
         feederPermissionList.addPermission(Type.TEST_RUN);
         feederPermissionList.addPermission(Type.VIEW_CLARIFICATIONS);
         feederPermissionList.addPermission(Type.VIEW_RUNS);
-        feederPermissionList.addPermission(Type.ANSWER_CLARIFICATION);
         feederPermissionList.addPermission(Type.SUBMIT_CLARIFICATION);
         feederPermissionList.addPermission(Type.VIEW_ALL_JUDGEMENTS);
         feederPermissionList.addPermission(Type.VIEW_STANDINGS);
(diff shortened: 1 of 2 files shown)


src/edu/csus/ecs/pc2/core/PermissionGroup.java
```diff
@@ -142,7 +142,6 @@ private void initialize() {
         feederPermissionList.addPermission(Type.TEST_RUN);
         feederPermissionList.addPermission(Type.VIEW_CLARIFICATIONS);
         feederPermissionList.addPermission(Type.VIEW_RUNS);
```
johnbrvc, line 144: I am a little concerned about the ANSWER_CLARIFICAITON permission.  Will this preclude the processing of answered clarifications received on the event feed?  This is one of the things we are adding for WF - the ability to read and process submitted clarifications AND the judge's responses to those clarifications received on the event feed.  Removing this permission may cause that not to work.  Based on my review of the code (PacketHandler), I think it may NOT work.

This also brings up another question in InternalController.java where this same permission is used (but in a different context).  There is a routine called InternalController.canCheckoutRunsAndClars().  The routine checks two permissions: JUDGE_RUN || ANSWER_CLARIFICATION.  Note the || (or).  The routine checks one OR the other, not AND.  It turns out the only place this is used is in the same module and it does appear that OR is the desired operation, but the routine appears to be misnamed.  Should an issue be created for this, or just a simple CI along with this PR, since it was noticed while reviewing this PR?

johnbrvc (commented):
Will await response to comment.

clevengr:
I agree with @johnbrvc about possible concerns regarding hidden ramifications of this change.  It's true that it's trivial to just add a tab to an existing pane -- but there may be hidden subtleties and these need to be extensively reviewed.  I think this disqualifies the PR from "One Approval".

src/edu/csus/ecs/pc2/core/PermissionGroup.java
lane55: Yes I also suspect that the answer clar from the EF for the FEEDER would fail if the ANSWER_CLARIFICAITON
is not present.   For that reason alone this is no longer a One Approval PR, because at this time it requires discussion
to firm up the requirements.   The real concern is that a person using shadow could answer a clar... and an answer
might be delayed on the Primary CCS.

re: renaming the canCheckoutRunsAndClars, note that it is an English vs computer speak naming and also
is combining two different permissions.   As such another issue could be created to actually create two
different methods one for runs the other for Clars.

lane55:
Not a priority, removing to reduce clutter

Issue #34, opened 2020-03-05
Provide a way to add sample data files to judges test data


lane55:
**Feature Description**: 

Add option for judge to execute sample problem data files (data/sample/*)
then execute judges data files (data/secret/*).   At NAC 2020 this was required
by the judges.

In CDP the samples are in directory
<short-name>/data/sample
the judge's data fils are in
<short-name>/data/secret

**Have you considered other ways to accomplish the same thing?** 

Yes, but they required renaming and moving files rather than
using the test data in place.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** 

Add a boolean field in ContestInformation to use to load samples (first)

Add a yaml setting to set the boolean

Add a checkbox on the Admin Settings pane

Print the field on a report

**Additional context**: 

**Additional info**: 

New setting name is TBD.

troy2914:
based on #125  it sounds like merging the sample and secret into 1 data structure does not make sense.   It should be two data structures where on the judge it can execute the sample first, then the secret.

clevengr:
It is not appropriate to execute the samples first, then the secret data --
this is what I was trying to explain in my copious earlier notes.  The
requirement is to take ALL data files, put them in alphabetical order, and
execute them in that order.  If the name of a sample file puts it between
two secret files, that's the order they should be executed in.

This actually came up at NAC, when PC2 was doing it "samples, then secret"
and Kattis was doing it "alphabetically".  The spec says to do it
alphabetically on all data files.  Kattis was right; pc2 is wrong.

troy2914:
whatever, sort it from both data sources.

PR #774, opened 2023-06-03
I 701 fix load samples find samples in executable

### LINKED ISSUE #701: when loading judge data sets (see #244) sample files not found when executing run
**Describe the issue**: 

Per change described in #244, when Executable attempts to find sample test data set,
those files are not found test case will fail.

But only happens IF one or more sample data sets.
Only happens if --load with external data files are used.

Effectively this does the same as this workaround:
in contest.yaml use this entry:
load-sample-judges-data: false

**To Reproduce**:

--load mini
view problem is Problem tab

submit a run
judge a run

**Expected behavior**:

There should be 7 data sets.

There should not be these three sets
sumit-samp-0.in
sumit-samp-1.in
sumit-samp-2.in

- when judging all data sets should be executed

**Actual behavior**:

There should be 7 data files.
sumit.in
sumit1.in
sumit2.in
sumit3.in
sumit4.in
sumit5.in
sumit6.in

- not all all data sets are executed

**Environment**:

All.

**Log Info**:

**Screenshots**:

**Additional context**:

All file names (basename) are loaded into Problem
pc2 coded to only find files in data/secret directory
Since samples are not found in that directory, the execute/test case fails.

lane55:

Added support for loading sample data files when loading config via yaml files ( --load option)
By default the load does NOT load sample files, to support existing/previous requirements.

To loading sample files add this line to the yaml
```
load-sample-judges-data: true
```

Fixes #701 

java version "1.8.0_121"

Load mini contest CDP which has the following line to load the samples.
```
load-sample-judges-data: true
```

1 - confirm that samples are loaded before judge data files
Edit Problem, view file names on Test Data Files
2 - submit and judge a run, confirm that all loaded data files are used

Test 2 - default behavior

load qanat sample contest
1 - confirm no samples are loaded before judge data files
Edit Problem, view file names on Test Data Files
2 - submit and judge a run, confirm that all loaded data files are used


johnbrvc:
Changes appear to work.  That is, the sample and secret files are properly presented for each test case to the submission.  However, there is no way to tell which files are the samples.  Displaying the problem report file just indicates the **basename** of the data file(s).  The "_true path_" shown in the report is to the problem folder, so that's not helpful.

```
   Using external files : true path = C:\MyNetProjects\ICPC\NAC2023\contests\nac23dress\config\cornhole
       Data File name 1 : 1.in
     Answer File name 1 : 1.ans
       Data File name 2 : 2.in
     Answer File name 2 : 2.ans
       Data File name 3 : 3.in
     Answer File name 3 : 3.ans
       Data File name 4 : 1.in
     Answer File name 4 : 1.ans
       Data File name 5 : 2.in
     Answer File name 5 : 2.ans
       Data File name 6 : 3.in
     Answer File name 6 : 3.ans
       Data File name 7 : p1004.in
     Answer File name 7 : p1004.ans

```

Note that the first 3 sets of files are from the sample folder and the next 3 sets are from the secret folder (same basenames, but no way to tell which is which).

After judging, there is still no way to tell what the actual path of the input file is (was it sample/1.in or secret/1.in)?

image

Both test case 1 and test case 4 have an input file of "1.in", but they are in fact different (sample/1.in and secret/1.in), but no way to tell which is which.  Keep in mind that the correct files are used for each test case, it's just the judge/user has no way to know which file they are looking at.

Suggest adding the `data/ `or `secret/ `to the report AND the judge's input/answer file dialogs.  Also, the `Edit Problem->Test Data Files` page should probably show `data/` or `secret/` , or maybe have another column that indicates if it's **sample** or **secret**.

image

lane55:
For identical data file base name, ex 1.in, added parent dir to Problems report for each file.

Before:
```
  judge data file '1.in' 0 bytes, External SHA1 = -179-26-110-901122-99-64-52-55-6346-987397-62-6532
```
After recent push
```
  judge data file '1.in' 0 bytes, External C:\repos\lane55fork\samps\contests\mini\config\sumit\data\sample SHA1 = -179-26-110-901122-99-64-52-55-6346-987397-62-6532
```

johnbrvc (approved):
Requested change made so reports show full path path of imported files.

johnbrvc:
BTW, when this PR is closed, shouldnt this also close issue #34 ??

clevengr:
> BTW, when this PR is closed, shouldnt this also close issue #34 ??

Yes, #34 is about the same issue as this PR (#774) -- which just goes to show how far back this issue goes.  Therefore, when this issue (i701) is closed it should also result in closing #34.    However, having said that, it's important to note that there was a _lengthy_ discussion in that PR (#34) regarding the "correct solution" (in particular, about the issue that the CLICS requirement is that all test data cases -- whether they come from "sample" or "secret" -- are to be executed in alphanumeric order by file name).  There were solutions proposed in PR #34 that I do not believe complied with this requirement; we need to make sure that when we finally merge a solution it _does_ meet this requirement.

lane55:
> about the issue that the CLICS requirement is that all test data cases -- whether they come from "sample" or "secret" -- are to be executed in alphanumeric order by file name)

At this point the pc2 implementation 
- loads samples (optionally) before loading secret data sets
- the code sorts the sample data sets in lexical order by filename then loads them
- then the code sorts the secret data sets in lexical order by filename then loads them

It is my understanding that the spec says that the samples are to be tested before the secret data sets.

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -159,11 +161,25 @@ public static String getSecretDataPath(String baseCDPPath, String problemShortNa
     }
 
     /**
-     * Return CCS path for input data and answer file names.
+     * Return CLICS path for sample data and answer file names.
```
clevengr, line 164: I believe the above comment about what the following method is doing is wrong.  It (appears to) claim that the method is returning the CLICS path to the **_sample_** directory -- but the name (and function) of the method is `get**Secret**DataPath()`.  Please correct the comment (or indicate why the comment is correct, and fix the method name).

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -78,12 +79,17 @@ public final class Utilities {
...
 
     /**
-     * CCS directory where data files are stored (under problem short name).
+     * CLICS directory where judge's (secret) data file are stored
      * 
      * @see #getSecretDataPath(String, Problem)
      * @see #getSecretDataPath(String, String)
      */
-    public static final String SECRET_DATA_DIR = "data" + File.separator + "secret";
+    public static final String SECRET_DATA_DIR = "data" + File.separator + ExportYAML.SECRET_DIRECTORY_NAME;
```
clevengr, line 87: It is very confusing that there are two different definitions in the code for the "secret data directory name".  Here, it defines `SECRET_DATA_DIR`, which sounds for all the world like _the name of the directory which holds the secret data_.  However, class `ExportYAML` _also_ contains what sounds like an equivalent definition:  `SECRET_DIRECTORY_NAME`.  It seems like there should be something to more clearly distinguish between these two.  Perhaps the constant defined here (in `Utilities`) should really be something like `SECRET_DATA_DIR_PATH`, since it's defining a _path_ rather than a specific directory name?

It also bothers me that the definition of `SECRET_DIRECTORY_NAME` is contained in class `ExportYAML`.  Use of the "secret directory name" is clearly not just something relating to YAML exporting.  I think that constant belongs elsewhere (maybe in `Constants.java`?)

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -78,12 +79,17 @@ public final class Utilities {
...
+    
+    /**
+     * CLICS directory where judge's sample data file are stored
+     */
+    public static final String SAMPLE_DATA_DIR = "data" + File.separator + ExportYAML.SAMPLE_DIRECTORY_NAME;
```
clevengr, line 92: The same comment which I made above (in regard to having two different definitions of the secret directory name) applies equally here:  having different constants `SAMPLE_DATA_DIR` and `SAMPLE_DIRECTORY_NAME` is confusing; I think one of the names should be changed (and I think `SAMPLE_DIRECTORY_NAME` doesn't belong in `ExportYAML`).

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -159,11 +161,25 @@ public static String getSecretDataPath(String baseCDPPath, String problemShortNa
...
+    
+    /**
+     * Return CLICS path for sample data and answer file names.
+     */
+    public static String getSampleDataPath(String baseCDPPath, String problemShortName) {
+        return baseCDPPath + File.separator + problemShortName + File.separator + SAMPLE_DATA_DIR;
+    }
+
+    /**
+     * Return CLICS path for input data and answer file names.
```
clevengr, line 178: This comment should say "... for **_sample_** data and answer file names, not "input" names...

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -960,58 +976,75 @@ public long stringToLongSecs(String s) {
...
+        return null;
+    }
+    
+    /**
+     * find judges data file under secret/ or samples/.
+     * 
+     * @param problem
+     * @param baseFileName
+     * @param alternateCDPPath
+     * @param judgeDataFile
```
clevengr, line 1008: This attribute describes an input parameter that doesn't exist in the method declaration.

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -960,58 +976,75 @@ public long stringToLongSecs(String s) {
...
-            secretPathPattern = secretPathPattern.replace('\\', '.');
+        return null;
+    }
+    
+    /**
+     * find judges data file under secret/ or samples/.
+     * 
+     * @param problem
+     * @param baseFileName
+     * @param alternateCDPPath
```
clevengr, line 1007: It sure would be nice if the documentation actually described the _meaning_ or _purpose_ of parameters like this.  It's not clear at all what an "alternate CDP path" is for or why it's needed.  What is it the "alternate" to?  Why would there exist _two_ CDP paths (a "main" and and "alternate")?

troy2914: I believe the alternateCDPPath is referring to the judgeCDPPath.

src/edu/csus/ecs/pc2/core/Utilities.java
lane55: Is this a bug that you have been able to reproduce?

clevengr: > Is this a bug that you have been able to reproduce?

I don't understand the above question.  The original comment is not about a "bug", it's about incorrect/illogical code.  The comment on this method says that it returns the "CLICS path for **_sample_** data and and answer files names".  The name of the method is `getSecretDataPath()`.   Either the comment is wrong, or the method is incorrectly named.  What is there to "reproduce"?

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -2926,40 +2927,16 @@ public Problem loadCCSProblemFiles(IInternalContest contest, String dataFileBase
...
         if (inputFileNames.length == answerFileNames.length) {
 
-            Arrays.sort(inputFileNames);
-            Arrays.sort(answerFileNames);
-
             ArrayList<SerializedFile> dataFiles = new ArrayList<SerializedFile>();
             ArrayList<SerializedFile> answerFiles = new ArrayList<SerializedFile>();
             
             if (loadSamples) {
+                // Load all sample files first, before loading files from sample dir.
```
clevengr, line 2934: This comment doesn't seem to make sense.  What does it mean to "load all sample files first",  but "before loading files from sample dir"?   Loading the sample files is going to load files from the sample dir, isn't it?

clevengr:
I loaded the "mini" contest.  The Test Data Files tab shows:

image

Since the `contest.yaml` file includes `load-sample-judges-data: true`, and the mini contest has 10 data sets (3 samps, 7 secret) I expected to see 10 data files in the display.  Either I'm doing something wrong, or the PR doesn't work as it's supposed to.

johnbrvc (approved):

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -440,7 +440,8 @@ public IInternalContest fromYaml(IInternalContest contest, String[] yamlLines, S
             setCcsTestMode(contest, ccsTestMode);
         }
         
-        // TODO 701 change LOAD_SAMPLE_JUDGES_DATA, false); to , true);
+        // TODO samples - At this time loading samples is not the default, there is a plan to load by default
+        // at a later date.   This would be done to avoid a configuration mistake of not loading sample data
         boolean loadSamples = fetchBooleanValue(content, LOAD_SAMPLE_JUDGES_DATA, false);
```
troy2914, line 445: did this part change with the change to the default?

clevengr: > did this part change with the change to the default?

As I understand it, making a change to this part is what "changing the default from false to true" is about.  I'm pretty sure that the change in PR #800 is precisely to change the `false` in the above call to `true` (along with deleting the preceding comments).

clevengr (changes requested):
I decided to change my review procedures:  rather than through the CODE first, I chose to execute the "Test Procedures" given in the PR (I'll review the code itself later).  When I tried to follow the PR Test Procedures, I ran into difficulties.

First, when I loaded the “mini” contest, ensuring that `load-sample-judges-data: true` was  present in the `contest.yaml`, it does seem to load both sample and secret data, but executing a submitted run produces `Error executing command: /usr/bin/wrapper_java 30 ISumit`.  The language definition needs changing, which isn’t covered in the PR test procedures.

Changing language definition by unchecking the “Use Judge specific execution command line” box produces:  

image

  But THEN it produces:

image

The production of the error message, when the files apparently are there, needs to be fixed.

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -957,61 +967,61 @@ public long stringToLongSecs(String s) {
...
+            testFileName = getSampleDataPath(alternateCDPPath, problem) + File.separator + baseFileName;
             if (fileExists(testFileName)) {
                 return testFileName;
             }
+        }
 
-            testFileName = serializedFile.getAbsolutePath();
+        testFileName = alternateCDPPath + File.separator + problem.getShortName() + File.separator + baseFileName;
+        if (fileExists(testFileName)) {
+            // try to find filename under shortname/ path
```
clevengr, line 1013: This comment seems superfluous (left over from previous code versions?) It describes neither the operation associated with the (preceding) `if` statement, nor the (single) line of code which follows it.   Suggest either rewriting the comment to better explain whatever it is that it's trying to explain, or else deleting it.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public String toJSON() throws JsonProcessingException {
+        ObjectMapper om = JSONObjectMapper.getObjectMapper();
+        om.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
+        // Older Jackson versions:
+        // om.configure(SerializationConfig.Feature.SORT_PROPERTIES_ALPHABETICALLY, true);
+        return om.writeValueAsString(this);
+    }
+
+}
```
clevengr, line 94: It strikes me as very unusual that this class provides no method for adding a team to the list of teams currently getting this award.  This means that if client code creates a `CLICSAward` and later wants to assign an additional team to receive the same award (something that seems like a _very likely occurrence_), the client must jump through unnecessary and inefficient hoops:  read the `team_ids` from the existing `CLICSAward`, construct a new array which contains the existing team ids plus the new team id, create a new `CLICSAward` using the new array, then (hopefully) dispose the old `CLICSAward`.  This becomes especially complicated if other client code has maintained a reference to the original `CLICSAward`; all such references would have to be tracked and accounted for when the new awardee was added.

It would be MUCH cleaner to add to the class a method like

```java
/**
  Adds an additional team to the current list of teams getting this CLICSAward.
  @param newTeamId  the id of the additional team to be added to the award list.
*/
public void addTeamToAwardList(String newTeamId) {

        String [] newTeamIds = new String [team_ids.length+1];
        System.arraycopy(team_ids, 0, newTeamIds, 0, team_ids.length);
        newTeamIds[team_ids.length] = newTeamId;
        team_ids = newTeamIds;
}
```
        
I consider it important enough to develop robust code that I request that you FIX this shortcoming; I even bothered to spend the time to provide necessary FIX code.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -0,0 +1,506 @@
...
+import edu.csus.ecs.pc2.core.standings.json.StandingScore;
+import edu.csus.ecs.pc2.core.standings.json.TeamScoreRow;
+
+/**
+ * A set of methods for CLICS API JSON
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class CLICSJsonUtilities {
```
clevengr, line 50: I do not understand why the name of this class includes the word "Json".  The class contains 13 non-private methods, only _two_ of which (`writeAwardsJSONFIle(PrintWriter, List<CLICSAward>)` and `writeAwardsJSONFIle(String, List<CLICSAward>)`) have anything to do with JSON; the rest of them mostly involve processing `List<CLICSAward>` elements.  I think a better name for the class would be `CLICSAwardUtilities`.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -0,0 +1,506 @@
...
+     * 
+     * @param contest
+     * @return
+     * @throws IOException
+     * @throws IllegalContestState
+     * @throws JAXBException
+     * @throws JsonMappingException
+     * @throws JsonParseException
+     */
+    public static List<CLICSAward> createAwardsList(IInternalContest contest) throws JsonParseException, JsonMappingException, JAXBException, IllegalContestState, IOException {
```
clevengr, line 89: Why does this method declare that it `throws JsonParseException` and `JsonMappingException`?   There's nothing in the method (or things it indirectly calls) that can throw those exceptions.  Methods shouldn't declare that they `throw` something that the don't/can't.  Recommend deleting these two exceptions from the `throws` clause.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -0,0 +1,506 @@
...
+                            /**
+                             * Team has to solve at least one problem to be the group winner 
+                             */
+                            groupWinners.put(teamGroup, clientId);
+                        }
+                    }
+                }
+            }
+        } catch (JAXBException | IllegalContestState | IOException e) {
+            StaticLog.getLog().log(Level.WARNING, "Error while calculating group winners DSA", e);
```
clevengr, line 153: I don't understand (and don't see any reason for) the term "DSA" in the above log message.  First, it's not really clear what this code has to do with "DSA" (if that's a reference to the "Default Scoring Algorithm").  Second, the appearance of such an acronym (or even of the full class name) in a log message isn't really helpful to our own developers and, more importantly, is just likely to confuse average users who are looking through the logs. Recommend removing "DSA" from the log message.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -0,0 +1,506 @@
...
+        }
+
+        Set<Group> problemElementIds = groupWinners.keySet();
```
clevengr, line 156: The variable name `problemElementIds` here is pretty misleading; the `keyset` of `groupWinners` is a set of `Group`s; it has nothing to do with "problems".  Recommend refactoring the name of the variable to something more accurate, like `groupElementIds`.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSJsonUtilities.java
```diff
@@ -0,0 +1,506 @@
...
+        }
+    }
+
+    public static ClientId createClientId(TeamStanding teamStand) {
+        ClientId clientId = null;
+
+        try {
+            return new ClientId(Integer.parseInt(teamStand.getTeamSiteId()), ClientType.Type.TEAM, Integer.parseInt(teamStand.getTeamId()));
+        } catch (Exception e) {
+            ; // ignore excepion return null
```
clevengr, line 181: Seems like this `catch` clause should log an error before falling through to return null.  Wouldn't we want to have a record of the fact that either the `teamSiteId` or the `teamId` couldn't be parsed properly?  That would seem to be an (internal) error worth logging...

samps/contests/mini/config/contest.yaml
```diff
@@ -63,7 +68,6 @@ languages:
     runner: 'java'
     runner-args: '{:basename}'
     use-judge-cmd: true
-    judge-exec-cmd: '/usr/bin/wrapper_java {:timelimit} {:basename}'
```
clevengr, line 66: This removes the (machine-specific) definition of the judge's execute command line (which seems appropriate).  However, the previous line still says `use-judge-cmd: true`.  Won't that result in an attempt to use an undefined judge's execute command line?

Note that this same comment applies to all the remaining changes (deletions) in this file.

src/edu/csus/ecs/pc2/core/imports/clics/CLICSAward.java
clevengr: Upon further review I've realized that, while I still think the above comment about the code needing the indicated method for robustness is true, that the fact is that there is no place anywhere in this PR that falls into the category of  actually _needing_ the suggested update.  Since that's the case, this should have properly been treated as a request for an _enhancement_ (rather than a PR fix), and accordingly I'm withdrawing this comment and will submit it as a separate "Enhancement" issue.

test/edu/csus/ecs/pc2/core/imports/clics/CLICSAwardUtilitiesTest.java
```diff
@@ -548,14 +548,16 @@ public void testreadAwardsList() throws Exception {
         String dataDir = getDataDirectory(getName());
 
         //        ensureDirectory(dataDir);
-        //        startExplorer(dataDir);
+//                startExplorer(dataDir);
+        
+        System.out.println("debug 22 AA "+dataDir);
```
clevengr, line 553: Debugging output should be omitted.

src/edu/csus/ecs/pc2/services/web/ContestService.java
```diff
@@ -153,13 +153,6 @@ public Response setStartime(@Context HttpServletRequest servletRequest, @Context
...
         }
 
-        // verify the JSON didn't contain any OTHER key/value information
-        if (requestMap.size() != 2) {
-            controller.getLog().log(Log.WARNING, "Contest Service PATCH: JSON input contains illegal extra data: '" + jsonInputString + "'");
-            // return HTTP 400 response code per CLICS spec
-            return Response.status(Status.BAD_REQUEST).entity("Extra data in contest patch request").build();
-        }
-
         // if we get here the JSON is valid and contains exactly one element: starttime
```
clevengr, line 156: This comment is no longer accurate; with the removal of the test that rejects the request if the JSON contained other fields, it's no longer necessarily true that the JSON `contains exactly one element`.

src/edu/csus/ecs/pc2/ui/EditRunPane.java
```diff
@@ -270,7 +270,35 @@ protected void updateRun() {
...
             RunStates prevState = run.getStatus();
             RunStates newRunState = (Run.RunStates) runStatusComboBox.getSelectedItem();
-
+            String errMsg = null;
+            
+            // Make sure it's safe to change the runstate
+            switch(newRunState) {
+            case JUDGED:
+            case MANUAL_REVIEW:
+            case REJUDGE:           // TODO: remove REJUDGE state completely since nobody appears to use it -- JB
```
clevengr, line 279: @johnbrvc I don't understand the above comment; I've seen multiple contests where judges "rejudged" a run (or many runs...)

src/edu/csus/ecs/pc2/ui/MessageManager.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Vector;
+
+/**
+ * Manage a set of message listeners.
+ * 
+ * @author Douglas A. Lane <pc2@ecs.csus.edu>
+ *
+ */
+public class MessageManager {
```
clevengr, line 12: Frankly, I don't think this class, or the related classes such as `IMessageRecordListener`, belong in this PR; they have little or nothing to do with the stated purpose of the PR ("fix load samples find samples in executable").  These classes should have been submitted as a separate Issue/PR.

However, what's done is done and I certainly don't, at this point, want to suggest we go back and redo everything.  I just want to point out that I don't think it was properly done to begin with; it's very hard to justify something so big (multiple classes; around 200 lines of code) as being just "continuous improvement" or that it rightly belongs in this PR...

test/edu/csus/ecs/pc2/core/imports/clics/CLICSAwardUtilitiesTest.java
```diff
@@ -0,0 +1,764 @@
...
+        
+        System.out.println("dumpStandings: "+message);
+        ContestStandings contestStandings = ScoreboardUtilites.createContestStandings(contest);
+
+        List<TeamStanding> teamStands = contestStandings.getTeamStandings();
+        for (TeamStanding teamStanding : teamStands) {
+            ClientId clientId = CLICSAwardUtilities.createClientId(teamStanding);
+            Group teamGroup = CLICSAwardUtilities.getGroupForTeam(contest, clientId);
+            if (teamGroup == null) {
+                System.out.println(teamStanding.getRank()+" "+teamStanding.getSolved()+" "+teamStanding.getPoints() + " " + //
```
clevengr, line 88: I thought we agreed that test files should not be writing to stdout...

test/edu/csus/ecs/pc2/core/imports/clics/CLICSAwardUtilitiesTest.java
```diff
@@ -0,0 +1,764 @@
...
+                        teamStanding.getTeamName() + " school:" + teamStanding.getShortSchoolName() + " TEAM HAS NO GROUP");
+            } else {
+                System.out.println(teamStanding.getRank()+" "+teamStanding.getSolved()+" "+teamStanding.getPoints() + " " + //
```
clevengr, line 91: I thought we agreed that test files should not be writing to stdout...

test/edu/csus/ecs/pc2/core/report/ContestCompareModelTest.java
```diff
@@ -182,46 +180,6 @@ public void testJSONComparisonMatch() throws Exception {
 
     }
 
-    /**
-     * Test compare of pretest CDP and existing event feed file.
-     * 
-     * @throws Exception
-     */
-    public void testReport() throws Exception {
```
clevengr, line 190: Why is this file being changed in this PR?  The file doesn't appear to have anything to do with the purpose of the PR...

test/edu/csus/ecs/pc2/core/report/ReportsTest.java
```diff
@@ -278,6 +278,7 @@ public void testReportTitles(){
                 "Problem Group Assignment", //
                 "Scoreboard JSON Report", //
                 "Compare Primary with model Report", //
+                "Awards Report", //
```
clevengr, line 281: Why is this file being changed in this PR? The change in the file doesn't appear to have anything to do with the purpose of the PR...

test/edu/csus/ecs/pc2/core/util/AbstractTestCase.java
```diff
@@ -792,6 +792,22 @@ public void editFile(String filename) {
         }
     }
     
+    public void editFile(String filename, String comment) {
```
clevengr, line 795: Why is this file being changed in this PR? The change in the file doesn't appear to have anything to do with the purpose of the PR...

clevengr (approved):
Following Doug's latest update to the PR, I re-ran my tests.  I found it now passes Test 1 (as given in the PR) :  it no longer generates an incorrect error message (stating that the data files could not be found) when a submitted run is executed.

However, Test 2 in the PR, which says
```
Test 2 - default behavior

load qanat sample contest
1 - confirm no samples are loaded before judge data files
Edit Problem, view file names on Test Data Files
2 - submit and judge a run, confirm that all loaded data files are used
```
fails. That is, according to what this test says,  no sample data files should be loaded.  In fact, the sample data files ARE loaded.

However, I suspect that the test in the PR was never updated to account for the subsequent PR which changed the default operation from "do not load sample data files" to "by default, DO load sample data files.  Assuming my suspicion is correct and the test should say "confirm sample data files ARE loaded", then this test would have passed.

I also tested by adding `load-sample-judges-data: false` to the `qanat` `contest.yaml` file and verified that in this case it did NOT load the sample data files.

Although I still have a number of comments and suggested code changes submitted as part of my review, none of them rise to the "showstopper" level so I approve this PR.  That does not however mean I consider the comments "resolved"; I believe they should be left until (if) we ever get to that "Someday" when we get time to actually do a thorough "code cleanup".

lane55:
> Following Doug's latest update to the PR, I re-ran my tests. I found it now passes Test 1 (as given in the PR) : it no longer generates an incorrect error message (stating that the data files could not be found) when a submitted run is executed.

When requirements change on the fly document/instruction lags while focusing on bugs.   
We agreed to this requirement change, so not a good use of time to have to read/respond to essentailly
what is not a bug per current requirements.

src/edu/csus/ecs/pc2/ui/EditRunPane.java
johnbrvc: Of course judges rejudge (I do it all the time), but that particular STATE is never used.  "BEING_RE_JUDGED" is the state that is used.  I think that REJUDGE must have been a left-over or something.  Hence, my comment.

johnbrvc (approved):
I am re-approving this (after reviewing the couple of files it said had changed since the last time I approved it).

PR #798, opened 2023-08-17
I538 support interactive problems

### LINKED ISSUE #538: Add support for interactive problems
**Describe the issue**: 

Add support for interactive problems.

TODO - find specification reference for interactive problems

In the pretest contest ( https://github.com/icpcsysops/ccsconfig-dhaka/tree/master/contests/pretest) there is a 'interactive' identifier identifying that the problem is interactive (?)

config/accessdenied/problem.yaml: validation: custom interactive
config/boredombuster/problem.yaml: validation: custom interactive

**To Reproduce**: 

**Expected behavior**: 

**Actual behavior**: 

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**: 

wf 2022 pretest had a interactive problem

johnbrvc:
Add built-in support for interactive problems.
Previously, interactive problems were handled by munging the execute command line from the edit problem GUI or in a problem's YAML file.
This PR makes PC2 aware of interactive problems.  Special custom shell scripts are used to setup the pipe work needed to connect a run's I/O to the O/I of a CLICS compliant custom interactive validator.  There are separate scripts for sandbox and non-sandbox problems.
An additional special script is used during the PC2 `validateProgram()` phase that copies the results of an interactive run performed by one of the aforementioned shell scripts to the location desired by PC2.

Summary of changes to the code:

- Add support in `ContestSnakeYAMLLoader.java` to recognize an interactive problem and perform appropriate validation of the validator supplied.
- Add support to the `ProblemPane` GUI to indicate if a problem is interactive.
- Add support to the `EditProblemPane `GUI to allow configuration of interactive problems.
- Add support for interactive problems to the `Problem` object.
- Expandthe `CustomValidatorSettings `object to be aware of interactive problems.
- `Executable.java`

  - Change to deal with interactive problems.
  - Refactor some code to put commonly used blocks in methods.
  - Add more substitution variables to support interactive problems and also allow for better logging in validators (`{:testcase}`, `{:executeinfofilename}`) .
  - Add routines to set up the interactive environment in the execute folder.
  - Modify build.xml and package.xml to correctly Zip and (gzip) the new **scripts** folder.  Remove  the **sandbox** folder.

- Create new scripts folder to hold all helper scripts used by sandboxes and interactive validators.  This involved getting rid of the `sandbox `folder (really, renaming it) and creating the `scripts `folder. 
- CI: Fix missing "`toString`" properties.
- CI: Make` teams2.tsv` be the preferred teams file during loading of a CDP.  This provides organizations.
- CI: better reporting of errors during reading of the CDP when loading a contest.  Combats the antagonizing "_check logs_" message previously reported.
- CI: Update copyright notices

There is still some future work to do in reading the results of the interactive validator (and sandbox) and reporting them on the GUI in PC2.  The current interactive and sandbox scripts create logs and "`reports`" in the "`execute*`" folder.  These could be read by PC2 (in the future) and reported to the user.  In addition, the current report files should be changed to create a single **NDJSON** type file (the framework is already in place for this).  This is NOT required for correct operation of interactive problems.  It would, however, make it nicer for someone reviewing results or manually judging.

Fixes #538

Windows 11/Ubuntu 22.04.2
Eclipse
java version "1.8.0_321"

**I might suggest TESTING the PR first before doing a code review so we can shake out any bugs before the code review.**

Please allocate enough time and patience to test this PR.  It's not a 10 minute task.  **This only works on Linux** due to sandboxes and interactive problems being used.  Appropriate warnings and denials will occur on _Windows_.  In addition, it requires the use of cgoupv2 - so be sure it's configured on the Linux system you are testing on.  The scripts including with PC2 will setup and configure the cgroup required by PC2, assuming the kernel supports it.

1. Load the `samps/contest/interactive` contest.  This is a new sample contest and it has 2 interactive problems (**Amazing** and **guess**), and a non-interactive problem (**pollution solution**)
2. By default, the CDP arranges for all problems to be judged by **AJ Judge1**.  If you want to manually judge, then you'll have to modify the problem and set it to manual judging.
3. Start an admin client.
4. Set the **Judges Data Path** so it points to the correct spot, namely:  basedir/samps/contests/interactive/config (or wherever)
5. Start **judge1** client (the autojudge) - if you get errors about missing data, then your **Judge Data Path** is wrong.
6. Start the contest (if you didn't use the GUI server, you'll need to start an admin client to start the contest)
7. Start **team1** client.
8. Submit one of the accepted solutions to **Amazing** (`samps/interactive/config/amazing/submissions/accepted`)
9. Observe AJ behavior and result received by team client.  Should be **correct**.
10. Submit more runs from **Amazing** and **Guess** to confirm interactive operation works.
11. Submit one or more runs from the **Pollution** problem.  This is a standard, non-interactive, built-in _CLICS_ validator problem (consider it a "_control_" problem).
12. Now submit some of the Wrong answer or time limit exceeded submissions also found under each problem.
13. Confirm correct operation.

Congratulations, you've confirmed that the PR works (assuming, it did!)

Now, do the code review.  First, calm yourself... even though there appear to be lots of commits, many of them are merge's from **develop** to keep the code up-to-date.  Other commits are for comments.  Others are because things were added, like the `samps/contests/interactive contest`.  This is not to say there aren't a lot of changes: there are.  It's just not as overwhelming as it may seem based on the number of commits and _Files Changed_ (currently, a whopping 212 files - mostly new ones for the new sample contest - so again, relax).


src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -3382,4 +3519,87 @@ private String formatTestCasePhase(String runPhase, int testCase)
...
+    /**
+     * Get the filename of a judge's file.  This could be a full path if the data files are external.  If the
+     * files are internal, then we return the first one's name since each dataset will get copied to the name
+     * of the first item.
+     * 
+     * @return filename of a judge's file
+     * @param wantInput - boolean indicating if we want the input file (true).  If false, we want the answer file
+     * @param setIndex - Which dataset element number are we interested in [0, #datasets-1]
+     */
+    private String getJudgeFileName(Utilities.DataFileType type, int setIndex)
```
lane55, line 3532: FIX: move this logic into JudgeUtilties so only a single line add/change in Executable.

lane55 (changes requested):
See comments with FIX in them

src/edu/csus/ecs/pc2/validator/customValidator/CustomValidatorSettings.java
```diff
@@ -102,7 +109,7 @@ public void setValidatorProgramName(String progName) {
      * @throws {@link RuntimeException} if the settings do not contain a valid Validator Interface mode setting
      */
     public String getCustomValidatorCommandLine() {
-        if (this.isUseClicsValidatorInterface()) {
```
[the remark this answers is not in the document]
johnbrvc: Agreed.  I was just following what the existing code that was there did.  It would look funny, seem odd and probably solicit more comments in the review if:

         if (this.isUseClicsValidatorInterface() || isUserInteractiveValidatorInterface()){

(one had `this `and one does not).
This is done in other places in this module as well (which doesn't make it right, but...).  All should be refactored at some point if we do it one place.  (Whoever wrote the module decided that "`this`" should be used for whatever reason.)
At this point, going through and removing all the "`this.`" is out of scope, no?

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -3764,7 +3809,33 @@ public void mousePressed(MouseEvent e) {
             + "\n be used by PC^2 to display additional information to the Judges and/or Teams."
 
             + "\n\nFor more information, see the PC^2 Contest Administrator's Guide; in particular, the Appendix on Validators."
-            + "\nSee also the CLICS specification at https://clics.ecs.baylor.edu/index.php/Problem_format#Validators.";
+            + "\nSee also the CLICS specification at https://icpc.io/problem-package-format/spec/problem_package_format#output-validators.";
```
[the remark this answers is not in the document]
johnbrvc: I agree.  It is horrendous that all these huge string constants for HELP messages are in actual source code.  These should be moved to a separate resource or file and looked up as needed.  This way they could easily be modified without a code change (when things like URL's change!).   But, out of scope here.  This can be part of the refactor plan.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -1427,6 +1444,7 @@ public void loadProblemInformationAndDataFiles(IInternalContest contest, String
             pc2FormatProblemYamlFile = true;
         }
 
+        // TODO: I am not sure why this test is contingent on pc2FormatProblemYamlFile. (JB)
```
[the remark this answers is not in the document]
johnbrvc: I do understand what the 'if' _does_, but I think it's silly.  If the title is `null`, then just set it to the `name `in the **yaml** file, right?  Why does it have to be a **pc2FormatProblemYamlFile** IMHO it has nothing to do with being a **pc2FormatProblemYamlFile**?  If '`name`' is in the file, use it.  The very next statement would cause it to abort out anyway if it's **null**, why not avoid that?

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: The problem is, `getJudgeFileName()` requires several context variables from `Executable`: `problem`, `problemDataFiles` and `log`  - these would have to be passed into the utility class method,  making it a very specific utility class method useful only to a caller that has "all those things", eg. `Executable`.  Do we want to do that?  That is, do we want:

    public static String getJudgeFileName(Utilities.DataFileType type, int setIndex, Problem problem, ProblemDataFiles problemDataFiles, Log log)

as the utility method?

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -1409,12 +1409,29 @@ public void loadProblemInformationAndDataFiles(IInternalContest contest, String
...
+        // also if the problem is interactive.
+        String validationType = fetchValue(content, VALIDATION_TYPE);
+        boolean isInteractive = false;
+        if (validationType != null) {
+            // validationType is a list of validation options
+            String[] valOpts = validationType.split("\\s");
+            // Must be one of: "default", "custom", "custom interactive"
+            if (valOpts.length >= 1 && valOpts[0].equals("custom")) {
+                usingCustomValidator = true;
+                if(valOpts.length >= 2 && valOpts[1].equals("interactive")) {
```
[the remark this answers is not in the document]
johnbrvc: Agree. Will fix this.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -1409,12 +1409,29 @@ public void loadProblemInformationAndDataFiles(IInternalContest contest, String
...
+                    // Note that interactive problems require a custom validator
+                    isInteractive = true;
+                }
+            }
```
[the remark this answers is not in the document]
johnbrvc: Agree. Will fix this.

src/edu/csus/ecs/pc2/core/execute/Executable.java
lane55: I will add this to my refactor list, no change needed at this time.

lane55 (approved):
Execellent coverages - approved

package.xml
```diff
@@ -122,21 +126,23 @@
...
+        <tar destfile="${gzip.path}" basedir="${build.unpack.dir}" compression="gzip" excludes="${release}/bin/**,${release}/samps/**,${release}/projects/**,${release}/scripts/**">
             <tarfileset mode="755" dir="${build.unpack.dir}">
                 <include name="${release}/bin/**"/>
                 <include name="${release}/samps/scripts/**"/>
                 <include name="${release}/samps/sleep.exe"/>
                 <include name="${release}/samps/**/distribute_score**"/>
                 <include name="${release}/samps/**/rotate**"/>
-                <include name="${release}/sandbox/**"/>
+                <include name="${release}/scripts/*.sh"/>
+                <include name="${release}/scripts/*.cmd"/>
```
clevengr, line 137: Previously, "all" files in the (formerly-named "sandbox") folder were included (`sandbox/**`).  Now, only `.cmd` and `.sh` files are being included.  This excludes, for example, the file `README.txt` in that folder -- which has some useful information about the organization and purpose of the files in the folder.  Shouldn't `README.txt` be included?  In fact, is there any reason not to just use `scripts/**`, analogously to what was there before?

samps/contests/interactive/config/system.pc2.yaml
```diff
@@ -0,0 +1,42 @@
...
+#
+
+
+# 8M
+max-output-size-K: 8192
+
+stop-on-first-failed-test-case: true
+team-scoreboard-display-format-string: "{:displayname}"
+
+load-data-files: false
```
clevengr, line 13: I wonder why data files are not being loaded into PC2; doing it this way requires the user/tester to deal with setting a correct Judge's Data Path, whereas if the files were loaded the user wouldn't have to do that.  The files do not seem to be very big -- a few K at most...

samps/contests/interactive/config/system.yaml
```diff
@@ -0,0 +1,65 @@
...
+- judges
+- sysops
+- operations
+- ccs
+default-clars:
+- Read the problem statement carefully
+- No comment
+languages:
+  - name: 'Java'
+    compilerCmd: '/usr/bin/javac -encoding UTF-8 -sourcepath . -d . {files}'
```
clevengr, line 11: The use of paths like `/usr/bin/javac` makes this "sample contest problem" configuration _platform-specific_.  I would strongly prefer seeing all of our sample contests use platform-independent (e.g. relative) paths -- that is, assume that things like `javac` are in the PATH, so that the files will work on any platform.

Note that this comment also applies to numerous lines below; I won't bother to add separate comments on each of those...

scripts/pc2validate_interactive.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+# Usage is: copy_xml_result.sh {:resfile} {:feedbackdir} {:testcase}
```
clevengr, line 2: This comment doesn't make sense -- it describes a usage for a script named `copy_xml_result.sh`, which is not this script.

src/edu/csus/ecs/pc2/core/Constants.java
```diff
@@ -191,18 +191,70 @@ private Constants() {
...
+    
+    /**
+     *  Filename of the script to run during the validate phase AFTER the interactive validator has completed
+     */
+    public static final String PC2_INTERACTIVE_VALIDATOR_NAME = "pc2validate_interactive.sh";
+    
+    /**
+     *  Command to execute to validate results of an interactive run AFTER the interactive validator has completed
+     */
+    public static final String PC2_INTERACIVE_VALIDATE_COMMAND = "./pc2validate_interactive.sh {:resfile} {:feedbackdir} {:testcase}";
```
clevengr, line 237: Since `PC2_INTERACTIVE_VALIDATOR_NAME` is a predefined constant (see above), the constant definition should be used in this line rather than the hard-coded string _pc2validate_interactive.sh_.

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -962,7 +962,7 @@ public long stringToLongSecs(String s) {
      * 
      * @param problem
      * @param serializedFile
-     * @param judgeDataFile
+     * @param judgeDataFile (not used)
```
clevengr, line 965: If the method parameter isn't used, it should be refactored/omitted.

Also, the javadoc parameter list omits one of the parameters which IS used:  `alternateCDPPath`.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -471,6 +481,11 @@ public IFileViewer execute(boolean clearDirFirst) {
                 int dataSetNumber = 0;
                 boolean passed = true;
 
+                // create somewhat secure ndjson file name for per test case execution results.
+                // format of the name is: RSSexecuteinfo.ndjson where R = run number, SS = current clock seconds
+                // eg: for run 453 at 12:00:09:   4539executeinfo.ndjson
```
clevengr, line 486: I think this format will be confusing to anyone who looks at the folder containing the files and tries to figure out the logic behind the file naming;   `4539` looks (something like) a `RunId`... but I would NEVER guess that I should split off the last digit and treat `453` and `9` as separate fields.  Wouldn't it be more helpful to do something like insert periods between the fields -- e.g. `453.9.executeinfo.ndjson`?  Or even better, `run493.time9.executeinfo.ndjson`?

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -3382,4 +3519,87 @@ private String formatTestCasePhase(String runPhase, int testCase)
...
+        return(result);
+    }
+    
+    /**
+     * Unpacks and creates a custom validator program.  If any error happens, an exception is thrown
+     * 
+     * @throws SecurityException - if the copy of the output validator can not be generated in the execute folder
+     * @throws IllegalStateException - if not output validator is available
+     */
+    void createValidatorProgram()
```
clevengr, line 3575: This method declaration should have `private` on it.

scripts/pc2validate_interactive.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
clevengr, line 1: I think the name of this script (`pc2validate_interactive.sh`) is a bit misleading.  The name makes it sound like this is a pc2 script which does "interactive validation".  In fact, I think what this script is doing (correct me if I'm wrong) is running a pc2 "post-validation" step.  That is, the thing known (in CLICS terminology) as the "interactive validator" -- the thing which interacts with the contestant submission and ultimately returns either 42 (for a correct submission) or 43 (for an incorrect submission) --  has _already been run_ when it comes time to run this script, and what this script is doing is handling the things that PC2 needs to do _after the "interactive validator" has been run_.  I think it would be better if the name of the script gave a clearer indication of this -- for example, if it was called `pc2_postinteractive_validation` (or maybe there's a better variant) -- but I think as it's current named, it's confusing about what it's really doing.

src/edu/csus/ecs/pc2/core/Constants.java
```diff
@@ -191,18 +191,70 @@ private Constants() {
...
+    
+    /**
+     *  Command to execute to run an interactive problem when not using a sandbox
+     */
+    public static final String PC2_INTERACTIVE_COMMAND_LINE = "./pc2_interactive.sh {:validator} {:infilename} {:ansfilename} {:testcase}";
+    
+    /**
+     *  Filename of the script to run during the validate phase AFTER the interactive validator has completed
+     */
+    public static final String PC2_INTERACTIVE_VALIDATOR_NAME = "pc2validate_interactive.sh";
```
clevengr, line 232: See my comment at the top of the `pc2validate_interactive.sh` script regarding updating the name of this script.

src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2429,7 +2459,93 @@ private void copyPC2Sandbox() throws Exception {
...
+        //use the VersionInfo class to get the PC2 installation directory
+        VersionInfo versionInfo = new VersionInfo();
+        String home = versionInfo.locateHome();      
+        
+        // Now, copy the files one at-a-time from the list
+        for(String scriptName : scriptNames) {
+            //point to the file that we want to create
+            String targetFileName = prefixExecuteDirname(scriptName);
+    
+            //point to the PC2 Internal Sandbox file (under "/sandbox" in the home, i.e. installation, directory)
```
clevengr, line 2533: I think this comment needs to be updated; I don't think there is an longer a "/sandbox" folder...

samps/contests/interactive/config/system.pc2.yaml
johnbrvc: I think we determined this was user error.

samps/contests/interactive/config/system.yaml
johnbrvc: Since interactive problems only work with Linux (at the moment), having Linux paths is not a major issue; at least that was my thought process.  I can just remove the FQP's and it SHOULD still work on Linux.  WIll try it.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: I am reluctant to remove it at this point because in our refactoring later, we may possibly use it.  Will address the extra Javadoc param.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: I modeled the format after the security results file that is currently created by PC2 (`#sssXRSAM.#`). (see `Executable.validateProgram()`.  It is mean to be obscure to avoid someone guessing it.   This file is not used yet.  It will hold the test case summary results.  It is current a replacement variable (`{:executeinfofilename}`) that we can easily add later to the validator command line.

scripts/pc2validate_interactive.sh
johnbrvc: 1) Users should never have to come in contact with this script.  It is internally called by PC2 and never by a user.  For nosy users, there is a nice, verbose comment in the script that describes what it does.  I originally was going to call it: PCVI.sh but I figured that would raise more than 2 eyebrows... (not to mention the confusion with "vi")
2) I thought the name told what it did:  It performs a PC2 style validation of the interactive results: PC2validate_interactive. 
3) Aside from the name being wrong in the usage comment (which I will fix), unless there are major objections, IMHO it can stay as is.  Again, this is a private to PC2 script - no user serviceable parts inside.

src/edu/csus/ecs/pc2/core/Constants.java
johnbrvc: Noted and will address that.

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -350,8 +350,13 @@ public enum SandboxType {
     
     private int memoryLimitMB = DEFAULT_MEMORY_LIMIT_MB;
     private SandboxType sandboxType = DEFAULT_SANDBOX_TYPE;
-    private String sandboxCmdLine = Constants.PC2_INTERNAL_SANDBOX_COMMAND_LINE;
-    private String sandboxProgramName = Constants.PC2_INTERNAL_SANDBOX_PROGRAM_NAME;
+    private String sandboxCmdLine = null;
+    private String sandboxProgramName = null;
```
clevengr, line 354: I'm not sure I understand the logic of deleting the default assignments for a Problem's `sandboxCmdLine` and `sandboxProgramName`.  Every Problem has a default memory limit (for example); shouldn't it also have a default sandbox command line/program name?  (Is it because the default sandbox TYPE is `Sandbox_Type.NONE`?)

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -248,7 +248,30 @@ public class EditProblemPane extends JPanePlugin {
     
     private EditProblemSandboxPane editProblemSandboxPane;
 
-
+    /**
+     * This enum defines the specification followed by the custom validator
+     */
+    private enum CUSTOM_VALIDATOR_TYPE {
```
clevengr, line 254: I think this is a bad name for this enum.  The enum name implies that it is an enumeration of _types of custom validators_; in fact, it is an enumeration of the _types of **interfaces** which a custom validator can use to communicate with PC2_.  I think it should be changed to (something like) `CUSTOM_VALIDATOR_INTERFACE_TYPE`.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -3703,6 +3744,10 @@ public void mousePressed(MouseEvent e) {
         return lblWhatsThisCLICSValidator;
     }
 
+    /*
+     * TODO: Future development: All these strings have to be moved out of the source code and into files or resources or something.  This stuff does
+     * not belong in source code. (JB)
```
clevengr, line 3749: I agree that (perhaps) the strings should be moved into their own class (it would certainly make reading THIS class, and ones like it, easier).  However, I'm not sure I agree that they should be moved (for example) into files.  External files could be edited by someone, resulting in "customization" of the messages in ways we would not be happy with (adding derogatory comments, for example).   Maintaining the strings in source code (although perhaps in separate files would require someone to edit the source and then rebuild the system.  (True, they could fork the project and do that, since it's open source.  But they couldn't alter the text being displayed by OUR DISTRIBUTION...)

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -5340,8 +5434,10 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
      */
     private CustomValidatorSettings getCustomValidatorSettingsFromFields() {
 
+        CUSTOM_VALIDATOR_TYPE cvt = getCustomValidatorTypeFromUI();
+        
         // make sure exactly one of the Interface mode buttons is selected
-        if (!(this.getUsePC2ValStdRadioButton().isSelected() ^ this.getUseClicsValStdRadioButton().isSelected())) {
+        if (cvt == CUSTOM_VALIDATOR_TYPE.CV_NONE) {
```
clevengr, line 5440: This change does not (I don't think) produce the correct effect.  The effect of the original code (the XOR in the `if` statement) was to show a message (and throw an exception) if it was the case that (somehow) either NONE of the "which interface to use" buttons were selected, or instead MORE THAN ONE of the buttons was selected.   This PR adds a third button -- but it removes the test which verifies that exactly one (of the now three) buttons is selected, replacing it with a test to see whether the "custom validator type" (which is really the "custom validator _interface_ type, as noted in another comment elsewhere) is NONE.  That may be an important test -- but it has removed the other test in the process.

I suppose one could argue that since the buttons are put in a `ButtonGroup`, they are automatically mutually exclusive (meaning, selecting one automatically deselects the others) -- which is true.  But that doesn't guarantee that one gets selected in the first place.  Currently, the code _does_ cause one to get selected.  But if some future programmer messes around with the code and (accidently or unknowingly) deletes the code which initially selects a button, then removal of this test will cause that condition to go undetected.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -1494,15 +1517,32 @@ private VALIDATOR_TYPE getValidatorTypeFromUI() {
         return validatorType;
     }
 
+    private CUSTOM_VALIDATOR_TYPE getCustomValidatorTypeFromUI() {
```
clevengr, line 1520: As noted in the comment about `enum CUSTOM_VALIDATOR_TYPE` above, this method isn't really returning a "custom validator TYPE"; it's returning an indication of the type of _interface_ used by the custom validator.    (I imagine for example that in the future we might have multiple different types of custom validators, of which "interactive" is just one type -- perhaps we would have CUSTOM_VALIDATOR_TYPE.HIGH_EFFICIENCY and CUSTOM_VALIDATOR_TYPE.MANUAL, for example -- the latter being a validator which operates by communicating with a human through the Judge's keyboard.  In that case, we wouldn't want the enum named `CUSTOM_VALIDATOR_TYPE`  to be returning the _type of interface used_; we'd want it to return the _type of the validator itself_.)

I think the method should be named (something like) `getCustomValidatorInterfaceTypeFromUI()` (or maybe a shorter form -- but something that makes it clear it's returning the type of the _interface used by the validator_, not the actual _type of the validator_.

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -1293,6 +1302,10 @@ public void setPrelimaryNotification(boolean prelimaryNotification) {
         this.prelimaryNotification = prelimaryNotification;
     }
 
+    public boolean isInteractive() {
+        return isUsingCustomValidator() && customValidatorSettings.isUseInteractiveValidatorInterface();
```
clevengr, line 1306: It seems to me that this is perhaps an inappropriate overloading of functionality.  The concept `problem is interactive` is (subtly, but IMHO clearly) different from `custom validator is using interactive validator interface`.  It seems like there should be a `Checkbox` somewhere in the problem definition that says "This problem is interactive", and the state of that checkbox should be what is used to determine the return value for method `isInteractive()`.

See also my comment in `EditProblemPane.getCustomValidatorOptionsSubPanel()` regarding the creation of the "Custom Validator 'Checkbox'".

src/edu/csus/ecs/pc2/core/model/Problem.java
```diff
@@ -2085,6 +2098,28 @@ public void setMemoryLimitMB(int memLimitInMB) {
...
+        // the sandbox name or command line even if the sandbox strings have not been defined yet
+        if(this.getSandboxType() == SandboxType.PC2_INTERNAL_SANDBOX || sandboxProgramName == null) {
+            if(isInteractive()) {
+                sandboxProgramName = Constants.PC2_INTERNAL_SANDBOX_INTERACTIVE_NAME;
+                sandboxCmdLine = Constants.PC2_INTERNAL_SANDBOX_INTERACTIVE_COMMAND_LINE;
+            } else {
+                sandboxProgramName = Constants.PC2_INTERNAL_SANDBOX_PROGRAM_NAME;                
+                sandboxCmdLine = Constants.PC2_INTERNAL_SANDBOX_COMMAND_LINE;
+            }
+        }
```
clevengr, line 2120: The above bracket marks the end of `if (this.getSandboxType() == SandboxType.PC2_INTERNAL_SANDBOX || sandboxProgramName == null)`.  It seems like there should be an `else` clause to deal with what happens if the `if` clause is false.  (It may be that the only logical thing to do for the `else` at this point in time is something like throw an exception -- maybe an `ILLEGAL_STATE` exception -- but it _does_ seem like the code _should_ do some such thing if such a condition is arrived at.)

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -5140,7 +5218,23 @@ private JPanel getCustomValidatorOptionsSubPanel() {
             gbc_labelWhatsThisCLICSValStd.gridx = 2;
             gbc_labelWhatsThisCLICSValStd.gridy = 3;
             customValidatorOptionsSubPanel.add(getLabelWhatsThisCLICSValStd(), gbc_labelWhatsThisCLICSValStd);
-        }
+            
+            // add the "This is an interactive problem" checkbox to the subpanel
```
clevengr, line 5222: The above comment relates to what I was saying in my comments on class `Problem`.  The comment clearly indicates that the concept is that there should be a _checkbox_ indicating "this is an interactive problem" -- but in fact what the code is adding is a _radio button_.   While it's true that a radio button _can_ be used in a way that it operates the same as a checkbox, THIS radio button is doing way more than that (it is, as I said in the comment on the `Problem` class, overloaded).  

Suppose for example that we accept the concept that the radio button represents the same thing as a "problem is interactive" checkbox.  Then what happens when one of the other radio buttons in the `ButtonGroup` is selected?  The (purported) "problem is interactive" button gets _deselected_.  Does that mean the problem suddenly is NOT interactive?   (Not if I understand the interface as presented.)

As I said elsewhere, I think the concepts of "checkbox indicating the problem is interactive" and "problem uses the CLICS interactive interface" need to be separated.

Having said that, I will concede that this review possibly comes to late to make it feasible to make changes to address this issue... in which case, so be it.  But my process for doing code reviews is, when I see what I consider a flaw, I will point it out.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -3288,12 +3351,13 @@ private boolean loadCCSTSVFiles(IInternalContest contest, File cdpConfigDirector
         boolean loaded = false;
 
         String teamsTSVFile = cdpConfigDirectory.getAbsolutePath() + File.separator + LoadICPCTSVData.TEAMS_FILENAME;
+        String teamsTSV2File = cdpConfigDirectory.getAbsolutePath() + File.separator + LoadICPCTSVData.TEAMS2_TSV;
```
clevengr, line 3354: It's a bit confusing that this variable is named `teamsTSV2File` when what it's referring to is a `teams2.tsv` file.  Why not name the variable `teams2TSVFile`?

(I can see a -- very slight -- argument that the position of the `2` in `teams2TSVFile` could be slightly misleading since use of `2` like that is commonly used to replace the word `to`; for example, `tex2pdf`,   `xml2json`, etc.  However, there's already a `2` in the existing name -- it reads like _teamsTSV to File_, which could just as easily be misinterpreted using that same logic.  I think it would be better to help avoid confusion between `teams` and `teams2` by renaming the variable.)

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -3288,12 +3351,13 @@ private boolean loadCCSTSVFiles(IInternalContest contest, File cdpConfigDirector
...
 
         String groupsTSVFile = cdpConfigDirectory.getAbsolutePath() + File.separator + LoadICPCTSVData.GROUPS_FILENAME;
 
         // only load if both tsv files are present.
```
clevengr, line 3358: The above comment should be updated to reflect the new code below.  It should probably say something like `only load if both a teams tsv file and the groups tsv file are present`.

src/edu/csus/ecs/pc2/util/OSCompatibilityUtilities.java
```diff
@@ -134,7 +134,7 @@ public static boolean areSandboxesSupported(IInternalContest contest, Log log)
         String home = versionInfo.locateHome();
         
         //point to the PC2 Internal Sandbox file (under "/sandbox" in the home, i.e. installation, directory)
```
clevengr, line 136: The above comment should be changed to reference the `scripts` folder instead of the `sandbox` folder.

clevengr (approved):
I have run the PR code against all three problems in the `interactive` sample contest, submitting all `accepted` as well as all "rejected" submissions defined in the problems.  The expected results were obtained 100% of the time.  (I do note that there are NO submissions which test MLE, for any of the problems...)

I have also done a code review of all 213 files comprising the PR (ok, I didn't really look at all the data files in the added problems, nor did I spend appreciable time looking through the "submissions" code  :)   I have made my usual (large) number of comments on the code, which I hope will be reviewed and perhaps acted upon.  However, none of the comments rise to the level where I would want to require them to be fixed before merging; hence, I've approved the PR.

I do sincerely hope we can, "someday", get to the point of taking non-blocking code review comments seriously...

samps/contests/interactive/config/system.pc2.yaml
clevengr: We did determine that for my testing this was explained.  What I was saying was that, in general, it seems to me that our "sample contests" should not contain lines like this, which override the default setting of `true`.

samps/contests/interactive/config/system.yaml
clevengr: Again, I based my comment on the concept that our sample contests should not be platform-specific, just like I believe they should not (normally) override defaults.  I think it would be better if the FQP's were made "non-qualified" (i.e. so they rely on the PATH, which is something we tell users right at the beginning of the Admin Guide that they must set -- at least for Java (and I think for all languages...)

src/edu/csus/ecs/pc2/core/Utilities.java
clevengr: Understood, and I can't disagree.  I just wanted to flag it as "something that needs to be addressed... someday".

src/edu/csus/ecs/pc2/core/execute/Executable.java
clevengr: > I modeled the format after the security results file that is currently created by PC2 (`#sssXRSAM.#`). (see `Executable.validateProgram()`. It is mean to be obscure to avoid someone guessing it. ...

Ok, good point.  Although I worry that it may be even "too obscure for US"...  Feel free to mark this comment "Resolved" if you wish.

scripts/pc2validate_interactive.sh
clevengr: R.E. Point 1:  note that I didn't say anything about "users" in my initial comment; I was talking about helping _developers_ keep things straight.  

R.E. Point 2:   I guess my argument would be that "PC2 style validation" is itself an ambiguous term.  We have a "PC2 Internal Validator" and a "CLICS Validator"; does the former do "pc2 style validation" and the latter does not?  The question is, what is the "validator"?   Generally, the contest configuration defines (selects, specifies) a "validator" and PC2 _executes_ that validator after each submission's execution against a given test case.   Now, we have a different concept:  PC2 is going to "execute the 'validator' _in order to generate the test case output_, and then is going to _subsequently_ invoke some (new) code, after the "validator" is run, to check the results.  That's why I think calling it "PC2validate_interactive" is misleading: the script does not contain the "validator" code which gets executed (in this case, to generate test case output); the script gets executed _after_ the "validator" gets run.  If we're going to require that the user provide a program, called a "validator", which we will run, then we shouldn't (IMHO) call the script which runs _after_ the validator by a name which sounds (to me) like it is actually a validator.  Maybe a better name would be something like `PC2_interactive_results_validator` or `PC2_post_interactive`.  However, see my next response.

R.E. Point 3: I don't think it's significant that it's a private script; all of the source code is "private" (to the developers).  But if you really think the name should remain as is, I'll accept your decision.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: You must not have been looking at the most recent commit in your review.  The javadoc for alternateCDPPath is, in fact, there.  However, I am adding more info to the judgeDataFile "not used" param.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: No code changes.

scripts/pc2validate_interactive.sh
johnbrvc: Will be fixed in next commit.

src/edu/csus/ecs/pc2/core/Constants.java
johnbrvc: Good point.  Will be fixed in next commit.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: Will be fixed in next commit.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: Comment updated.  Changed to /scripts.

src/edu/csus/ecs/pc2/core/Utilities.java
clevengr: > You must not have been looking at the most recent commit in your review. The javadoc for alternateCDPPath is, in fact, there. However, I am adding more info to the judgeDataFile "not used" param.

I pulled from your fork, then immediately rebuilt the system and proceeded with my testing -- so I'm not sure how I could have not been looking at the "most recent commit", unless you made a commit in your local repo that you didn't push to your fork?

johnbrvc: Aha.  When Develop was merged into it yesterday, apparently a DIFFERENT PR fixed that missing javadoc, and it got added then.  
image

src/edu/csus/ecs/pc2/core/model/Problem.java
johnbrvc: Yes.  This logic has been changed.  If the value(s) is(are) null, it is dealt with in updateSandboxInfo().
There's a comment there that explains it.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
johnbrvc: Will be fixed in the next commit.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
johnbrvc: No action being taken.  Code works as-is.  If we find a problem later for some unknown reason, we can create an issue.

src/edu/csus/ecs/pc2/core/model/Problem.java
johnbrvc: The method is private.  It is used only by  other sandbox routines in the same module.  It does not matter if it "works" or not.  It just updates those 2 member variables is necessary.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
johnbrvc: Actually, that comment is wrong.  It's a radio button as you indicate.  IV's can ONLY be (custom)  **CLICS** type validators (at least, that is all we are supporting since that is what we have a spec for).  They can not be PC2 type validators.  Selecting a PC2 type validator correctly means  it is "non-interactive".  There are 3 choices for a  custom validate: **pc2**, **clics**, **clics interactive**.  That is it.  Adding a checkbox next to "_Use CLICS Standard Interface_" makes the interface confusing to use (IMHO).  The radio button makes it crystal clear what you are getting.  And, if  the user wants more explanation, there's always the What's this? question mark.

If this is sticking point, then after the merge an issue can be created discussing the merits of changing the UI.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: Yes.  Agreed.  Changed in next commit.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: Fixed in next commit.

src/edu/csus/ecs/pc2/util/OSCompatibilityUtilities.java
johnbrvc: Changed in next commit.

scripts/pc2validate_interactive.sh
johnbrvc: Fixed Usage message in comments to contain correct name in next commit.

Point 1 and 3 go hand-in-hand.  I do not consider these scripts source code.  These files are part of the standard distribution in the folder scripts.  We do not include 'java' source code with a distribution, but we do include these.  The fact these are in a separate folder (scripts) and not (bin) means they are none of the user's business to be monkeying around with (this will be in the administrators guide).  These are more "support" files, and i toyed with the idea of stuffing them in the support folder, but there is other seemingly unused stuff in there.

Point 2 is philosophical.  Probably better for a meeting discussion.

src/edu/csus/ecs/pc2/core/model/Problem.java
johnbrvc: See my comments in the comment about the reasoning for a radio group.

johnbrvc:
> I have run the PR code against all three problems in the `interactive` sample contest, submitting all `accepted` as well as all "rejected" submissions defined in the problems. The expected results were obtained 100% of the time. (I do note that there are NO submissions which test MLE, for any of the problems...)
> 
> I have also done a code review of all 213 files comprising the PR (ok, I didn't really look at all the data files in the added problems, nor did I spend appreciable time looking through the "submissions" code :) I have made my usual (large) number of comments on the code, which I hope will be reviewed and perhaps acted upon. However, none of the comments rise to the level where I would want to require them to be fixed before merging; hence, I've approved the PR.
> 
> I do sincerely hope we can, "someday", get to the point of taking non-blocking code review comments seriously...

I gave them all thoughtful review (maybe too much thought in some cases ;-)).  The ones I did not comment on can be discussed "someday".  In the case of changing the "interactive" sample contest CDP to use load data files and remove full paths
I will create an issue for this after the merge.

Issue #801, opened 2023-09-04
Per-problem output limit in problem.yaml has wrong units


johnbrvc:
**Describe the issue**: please provide a clear and concise description of what the bug is.  
The CLICS Problem Package format specification states that the per-problem output limit property is in MiB.  PC2 interprets this property as KiB.

![image](https://github.com/pc2ccs/pc2v9/assets/5657363/f8b2be8f-27ae-4a9c-9780-2113a719ca27)

Needless to say, this could lead to errant "output limit exceeded" judgements (WA).

**To Reproduce**:
Specify a per-problem output value of 16.  Note that PC2 reports this as 16KB:

![image](https://github.com/pc2ccs/pc2v9/assets/5657363/b8b2c706-158a-4c4d-b4e5-f32eb0be7779)

**Expected behavior**: 
The value should be converted to MiB.

**Actual behavior**: 
The value is converted to KiB.

**Environment**:
All.

**Log Info**:  

**Screenshots**:  

**Additional context**: 
Observed while testing wfsharm/wf46/systest1 contest.  The ETA problem has a custom output limit.

troy2914:
missed the fixes #801 in your commit message.  Closing.

PR #802, opened 2023-09-04
I801 Fix wrong units for CLICS per-problem output limit

johnbrvc:
CLICS Problem Package Formats states the output limit units are MiB.  PC2 was using KiB.

Interprets the output property of the limits object in the per-problem problem.yaml file as **MiB** instead of **KiB**.  This may be related to Issue #564 and PR #639.  
#801 

Java(TM) SE Runtime Environment (build 1.8.0_351-b10)
Windows 11

1) Delete any prior contest configuration you may have loaded from the `profiles `folder.  And delete `profiles.properties`.
2) Load the `samps/contests/sumithello/config/mini` sample contest.  Note that the `limits:` object looks like this in the `problem.yaml `file:

```
limits: 
   timeout: 30
   output: 1

```
3) Start an `admin `client
4) View the `sumit`problem and note the output limit of 1024 as opposed to 1.

### DIFF
--- a/samps/contests/mini/config/sumit/problem.yaml
+++ b/samps/contests/mini/config/sumit/problem.yaml
@@ -15,7 +15,7 @@ rights_owner:
 max-output-size-K: 162
 limits: 
    timeout: 30
-   output: 778
+   output: 1
 
 validator: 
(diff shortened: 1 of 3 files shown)


troy2914 (approved):
Tested this PR,  worked fine.

Issue #810, opened 2023-09-19
Sandbox execution does not clean up straggler processes


johnbrvc:
**Describe the issue**:
If a user submission forks itself (which is in and of itself against contest policy - but, nevertheless...) and the original process is killed or otherwise goes away, the child processes can remain around and continue forking even after the sandbox has terminated.  This can jam up an autojudge (AJ) by causing future runs by that AJ to fail due to too many processes.  This was noticed when running wf46/systest0 during testing.

**To Reproduce**:
Submit and judge the Linked C program below.
[stress-test-fork-setsid.txt](https://github.com/pc2ccs/pc2v9/files/12661621/stress-test-fork-setsid.txt)

The C program forks, then creates a new session ID making it hard to keep track of and the stragglers will keep on multiplying.

**Expected behavior**: 
When a sandbox run finishes, all children should be terminated and disappear.

**Actual behavior**:
Submitting the above C code will cause straggler "./a.out" processes to linger forever, jamming up the AJ and causing random errors such as "too many processes" since the sandbox limits it by UID.

**Environment**:
Ubuntu 22.04 WF Contest Image
PC2 with sandbox support

**Log Info**:  

**Screenshots**:

**Additional context**:
`cgroupv2 `provides a way to kill of ALL process in the **cgroup**.  The sandbox scripts should use this method when they exit (See https://docs.kernel.org/admin-guide/cgroup-v2.html#:~:text=A%20write%2Donly%20single%20value,will%20be%20killed%20via%20SIGKILL.)

clevengr:
This Issue appears to have been fixed by PR 812 ([I810 Make sure to kill off all processes in cgroup at start #812](https://github.com/pc2ccs/pc2v9/pull/812)).  It should have been closed after that PR was approved and merged.  The reason it did NOT get closed is likely that the PR comment in the PR section regarding "Issue which this PR addresses" just has `#810` in it.  In order for GitHub to automatically close an issue from a PR, the PR must contain either the word `Closes` or the word `Fixes` in front of the Issue number -- not just the Issue number by itself.

Closing the issue since the PR for it was approved and merged.

PR #812, opened 2023-09-21
I810 Make sure to kill off all processes in cgroup at start

johnbrvc:
1) Put submission in its own progress group (session) to make it easier to kill off later
2) Kill off all members of the process group upon termination
3) Kill off all processes that have the script as its parent (catches some stragglers)
4) Purge cgroup at startup to prevent left over stragglers from previous runs
5) CI: Add shell function to nicely print out resource usage for each test case
ex:
```
Resources used for this run:
   CPU ms  Limit ms    Wall ms   Memory Used Memory Limit
 2011.360  2000.000   2012.259      98971648   2147483648
The command was killed because it exceeded the CPU Time limit

```
6) CI: If process was killed due to a signal, show the signal number in the sandbox log
7) More robust reporting of time limit exceeded

Put the Issue Number in this section.
#810 
Ubuntu 22.04
Linux 5.19.0

1) Load up the `wfsharm/contests/wf46/systest0` contest
2) Replay the contest (exceedingly time consuming)
3) Note that contest plays through (previously it would hang after all the AJ's were occupied with infinite loop fork bombs that created separate process groups
To test CI's:
1) Load up a contest that has problems that use sandboxes (preferably some that have interactive problems such as the sample contest **interactve**
2) Submit runs
3) Judge runs
4) Look in the `executesite*` folder at the sandbox.log file, and if it's an interactive problem look in the `executesite*/reports` folder for per-case log files.

A typical report file looks like this for an interactive problem:
```
$ cat reports/testcase_004.log
Setting memory limit to 2048 MB
Starting: ./output_validator_out /home/ubuntu/pc2/current/config/guessingprimes/data/secret/04-random_1000.in /home/ubuntu/pc2/current/config/guessingprimes/data/secret/04-random_1000.ans interactive_feedback > fin < fout &
Started interactive validator PID 583391
Setting cpu limit to 10000000 microseconds (ulimit -t 10 )
Setting maximum user processes to 171
Putting 583376 into /sys/fs/cgroup/pc2/sandbox_pc2_550201 cgroup
Executing setsid taskset 0x08 /usr/bin/pypy3 ragnar-manual-begin.py < fin > fout
Contestant PID 583397 finished with status 0
Resources used for this run:
   CPU ms  Limit ms    Wall ms   Memory Used Memory Limit
 1092.647 10000.000   1129.340      54059008   2147483648
The command terminated normally.
Finished executing /usr/bin/pypy3 ragnar-manual-begin.py
/usr/bin/pypy3 exited with exit code 0
Waiting for interactive validator to finish...
Validator finishes with status 42
________________________________________

----------------------------------
Results XML File for testcase #4:
----------------------------------
<?xml version="1.0"?>
<result outcome="Accepted" security="174613XRSAM.4.txt"></result>
```

A typical report file looks like this for a timelimit exceeded run:

```
$ cat sandbox.log
checking PC2 CGroup V2 installation...
...done.
Removing existing sandbox to start clean
Purging cgroup /sys/fs/cgroup/pc2/sandbox_pc2_550201/cgroup.kill of processes
Creating sandbox /sys/fs/cgroup/pc2/sandbox_pc2_550201
checking memory limit
setting memory limit to 2048 MB
setting cpu limit to 5000000 microseconds (ulimit -t 5 )
setting maximum user processes to 170
putting 583976 into /sys/fs/cgroup/pc2/sandbox_pc2_550201 cgroup
Executing setsid taskset 0x08 /usr/bin/java -Dfile.encoding=UTF-8 -XX:+UseSerialGC -Xss64m -Xms1920m -Xmx1920m Main
Killing off submission process group 583990 and all children
Resources used for this run:
   CPU ms  Limit ms    Wall ms   Memory Used Memory Limit
 5008.318  5000.000   5009.582      15613952   2147483648
The command was killed because it exceeded the CPU Time limit
Finished executing /usr/bin/java -Dfile.encoding=UTF-8 -XX:+UseSerialGC -Xss64m -Xms1920m -Xmx1920m Main
/usr/bin/java exited with exit code 244


```

### DIFF
--- a/scripts/pc2_interactive.sh
+++ b/scripts/pc2_interactive.sh
@@ -114,7 +114,7 @@ HandleTerminateFromPC2()
 	pkill -9 -P $$
 	if test -n "$submissionpid"
 	then
-		kill -9 -"$submissionpid"
+		pkill -9 -s "$submissionpid"
 	fi
 	DEBUG echo $0: Wall time exceeded - exiting with code $FAIL_WALL_TIME_LIMIT_EXCEEDED
 	exit $FAIL_WALL_TIME_LIMIT_EXCEEDED 
@@ -254,7 +254,7 @@ do
 		fi
 
 		# Kill off submission pgrp
-		kill -9 -"$submissionpid"
+		pkill -9 -s "$submissionpid"
 
 		if test "$wstat" -eq $EXITCODE_AC
 		then
@@ -310,9 +310,9 @@ REPORT "________________________________________"
 rm -f "$INFIFO" "$OUTFIFO"
 
 # Kill pgrp
-kill -9 -"$submissionpid"
+pkill -9 -s "$submissionpid"
 
-# Kill off stragglers
+# Kill off stragglers with us as a parent
 pkill -9 -P $$
 
 # TODO: determine how to pass more detailed pc2sandbox.sh results back to PC2... Perhaps in a file...
--- a/scripts/pc2_interactive.sh
+++ b/scripts/pc2_interactive.sh
@@ -108,11 +108,13 @@ KillValidator()
 HandleTerminateFromPC2()
 {
 	DEBUG echo "Received TERMINATE signal from PC2"
+	DEBUG echo "Killing off submission process group $submissionpid and all children"
 	KillValidator
-	if test -n "$submissionpid" -a -d /proc/$submissionpid
+	# Kill off all our kids
+	pkill -9 -P $$
+	if test -n "$submissionpid"
 	then
-		DEBUG echo "Killing off submission process $submissionpid"
-		kill -9 "$submissionpid"
+		kill -9 -"$submissionpid"
 	fi
 	DEBUG echo $0: Wall time exceeded - exiting with code $FAIL_WALL_TIME_LIMIT_EXCEEDED
 	exit $FAIL_WALL_TIME_LIMIT_EXCEEDED 
@@ -209,9 +211,9 @@ REPORT Started interactive validator PID $intv_pid
 starttime=`GetTimeInMicros`
 
 # run the command
-REPORT_DEBUG Executing "taskset $CPUMASK $COMMAND $* < $INFIFO > $OUTFIFO"
+REPORT_DEBUG Executing "setsid taskset $CPUMASK $COMMAND $* < $INFIFO > $OUTFIFO"
 
-taskset $CPUMASK $COMMAND $* < $INFIFO > $OUTFIFO  &
+setsid taskset $CPUMASK $COMMAND $* < $INFIFO > $OUTFIFO  &
 # Remember child's PID for possible killing off later
 submissionpid=$!
 
@@ -244,13 +246,16 @@ do
 			if test -d /proc/$contestantpid
 			then
 				REPORT Contestant PID $submissionpid has not finished - killing it
-				kill -9 "$submissionpid"
 			fi
 			# This is just determines if the program ran, not if it's correct.
 			# The result file has the correctness in it.
 			# We only do this if the contestant program has not finished yet.
 			COMMAND_EXIT_CODE=0
 		fi
+
+		# Kill off submission pgrp
+		kill -9 -"$submissionpid"
+
 		if test "$wstat" -eq $EXITCODE_AC
 		then
 			GenXML Accepted ""
@@ -303,6 +308,12 @@ DEBUG echo
 REPORT "________________________________________"
 
 rm -f "$INFIFO" "$OUTFIFO"
+
+# Kill pgrp
+kill -9 -"$submissinpid"
+
+# Kill off stragglers
+kill -9 -P $$
 
 # TODO: determine how to pass more detailed pc2sandbox.sh results back to PC2... Perhaps in a file...
 
--- a/scripts/pc2_interactive.sh
+++ b/scripts/pc2_interactive.sh
@@ -310,10 +310,10 @@ REPORT "________________________________________"
 rm -f "$INFIFO" "$OUTFIFO"
 
 # Kill pgrp
-kill -9 -"$submissinpid"
+kill -9 -"$submissionpid"
 
 # Kill off stragglers
-kill -9 -P $$
+pkill -9 -P $$
 
 # TODO: determine how to pass more detailed pc2sandbox.sh results back to PC2... Perhaps in a file...
 
--- a/scripts/pc2sandbox.sh
+++ b/scripts/pc2sandbox.sh
@@ -216,7 +216,7 @@ then
 	if ! rmdir $PC2_SANDBOX_CGROUP_PATH
 	then
 		DEBUG echo Cannot remove previous sandbox: $PC2_SANDBOX_CGROUP_PATH
-		exit $FAIL_INVALID_CGROUP_INSTALLATION
+		exit $FAIL_SANDBOX_ERROR
 	fi
 fi
 
--- a/scripts/pc2sandbox.sh
+++ b/scripts/pc2sandbox.sh
@@ -66,7 +66,7 @@ _DEBUG="on"   # change this to anything other than "on" to disable debug/trace o
 DEBUG_FILE=sandbox.log
 function DEBUG()
 {
-  [ "$_DEBUG" == "on" ] && $@ >> $DEBUG_FILE
+  [ "$_DEBUG" == "on" ] && "$@" >> $DEBUG_FILE
 }
 
 # ------------------------------------------------------------
@@ -124,6 +124,24 @@ GetTimeInMicros()
         ret=$((sec*1000000))
         ret=$((ret+$us))
         echo $ret
+}
+
+# Show run's resource summary in a nice format, eg.
+#   CPU ms  Limit ms    Wall ms   Memory Used   Memory Limit
+#    3.356  5000.000      4.698       1839104     2147483648
+ShowStats()
+{
+	cpuused=$1
+	cpulim=$2
+	walltime=$3
+	memused=$4
+	memlim=$5
+	DEBUG echo Resources used for this run:
+	DEBUG printf "   CPU ms  Limit ms    Wall ms   Memory Used Memory Limit\n"
+	DEBUG printf "%5d.%03d %5d.%03d %6d.%03d  %12s %12d\n" $((cpuused / 1000)) $((cpuused % 1000)) \
+		$((cpulim / 1000)) $((cpulim % 1000)) \
+		$((walltime / 1000)) $((walltime % 1000)) \
+		$((memused)) $((memlim))
 }
 
 # ------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


scripts/pc2_interactive.sh
```diff
@@ -108,11 +108,13 @@ KillValidator()
 HandleTerminateFromPC2()
 {
 	DEBUG echo "Received TERMINATE signal from PC2"
+	DEBUG echo "Killing off submission process group $submissionpid and all children"
 	KillValidator
-	if test -n "$submissionpid" -a -d /proc/$submissionpid
+	# Kill off all our kids
+	pkill -9 -P $$
```
troy2914, line 114: I am confused by the -P 
```
      -P, --parent ppid,...
              Only match processes whose parent process ID is listed.
```

If a process forks does it's parent process ID not get request to 1?

scripts/pc2_interactive.sh
```diff
@@ -108,11 +108,13 @@ KillValidator()
...
+	if test -n "$submissionpid"
 	then
-		DEBUG echo "Killing off submission process $submissionpid"
-		kill -9 "$submissionpid"
+		kill -9 -"$submissionpid"
```
troy2914, line 117: i also don't understand this change of adding a - before the pid?

scripts/pc2_interactive.sh
```diff
@@ -244,13 +246,16 @@ do
...
-				kill -9 "$submissionpid"
 			fi
 			# This is just determines if the program ran, not if it's correct.
 			# The result file has the correctness in it.
 			# We only do this if the contestant program has not finished yet.
 			COMMAND_EXIT_CODE=0
 		fi
+
+		# Kill off submission pgrp
+		kill -9 -"$submissionpid"
```
troy2914, line 257: again the - here on the pid?  are you expecting the submissionpid to 1?

scripts/pc2sandbox.sh
```diff
@@ -65,16 +82,35 @@ timelimit, in seconds
...
+}
+
+# Kill active children and stragglers
+KillChildProcs()
+{
+	DEBUG echo "Killing off submission process group $submissionpid and all children"
+	# Kill off process group
+	if test -n "$submissionpid"
+	then
+		kill -9 -$submissionpid
```
troy2914, line 102: again with - before the pid,  I've looked at bash kill and /bin/kill and can find no reference for anything but the -1 case to kill 
```
 it indicates all processes except the kill  process  itself and init.
```

troy2914: I would rather all those be just written as -1 if that's the case.

troy2914 (commented):
if you mean -1 just do -1.

scripts/pc2_interactive.sh
johnbrvc: that kills all processes that has the shell script pid ($$) as it's parent, but, doesnt kill the killer.
If it forks, it does not change its parent.  If it does a setsid() then yes, it becomes one, in which case this pkill would be a no-op.

scripts/pc2_interactive.sh
johnbrvc: kill with a negative PID kills the process group with that ID.  If we put the submission in it's own session (setsid taskset ...), then we can kill that whole process group when we're done to make sure anyone in that group is killed.  This does NOT catch the case where a forked process does it's own setsid() call, but it will catch those who don't.  The ones that do the setsid() call will be killed when we write a 1 to the cgroup kill file, which we do before we initialize the cgroup.

scripts/pc2_interactive.sh
johnbrvc: No.  It' is the process group of the submission.

scripts/pc2sandbox.sh
johnbrvc: I suppose, but i thought that killing the pgroup was safer, just in case it was accidently run by root.  Doesn't kill -9 -1 kill everyone in that case?  Just killing the PGRP is safer, I would think.

scripts/pc2_interactive.sh
troy2914: Where is that documented?

scripts/pc2sandbox_interactive.sh
```diff
@@ -38,7 +38,13 @@ FAIL_INTERACTIVE_ERROR=$((FAIL_RETCODE_BASE+55))
 MAXPROCS=32
 
 # taskset cpu mask for running submission on single processor
-CPUMASK=0x01
+cpunum=${USER/judge/}
+if [[ "$cpunum" =~ ^[1-5]$ ]]
```
troy2914, line 42: doesn't this restrict one use judge numbers 1 to 5 ?

scripts/pc2_interactive.sh
johnbrvc: `man kill`
```
DESCRIPTION
       The default signal for kill is TERM.  Use -l or -L to list available signals.  Particularly useful signals include HUP, INT, KILL, STOP, CONT, and 0.
       Alternate signals may be specified in three ways: -9, -SIGKILL or -KILL.  **Negative PID values may be used to choose whole  process  groups**;  see  the
       PGID column in ps command output.  A PID of -1 is special; it indicates all processes except the kill process itself and init.
```

scripts/pc2sandbox_interactive.sh
johnbrvc: It does.  it should probably be much smarter and figure out how many CPUS's are available and only use the number that are.  This was more for testing on the same system though, to skew across the cpus.  In practice we use 'ubuntu' which pins to cpu 0 (default) and only one per system.

scripts/pc2_interactive.sh
troy2914: hmm,  you are referencing kill though which is bash internal.    
```
       NOTES  Your shell (command line interpreter) may have a  built-in  kill
              command.   You  may  need  to  run the command described here as
              /bin/kill to solve the conflict.
```

johnbrvc: Oh.  Excellent point.

scripts/pc2sandbox.sh
```diff
@@ -33,23 +33,40 @@ FAIL_SANDBOX_ERROR=$((FAIL_RETCODE_BASE+54))
 MAXPROCS=32
 
 # taskset cpu mask for running submission on single processor
-CPUMASK=0x01
+cpunum=${USER/judge/}
```
clevengr, line 36: It's probably due to the fact that my Bash knowledge has weakened over time, but I don't understand what the above line is doing.  Specifically, what is the value of the variable `USER/judge/`?    ( know what `${USER}` means; I don't know what the meaning of adding the `/judge/` to the argument is...)

And what happens if the submission is being executed by an Admin rather than a judge?

troy2914: It removes judge from the USER variable

johnbrvc: basically: cpunum=echo $USER | sed -e 's/judge//'
erases judge from the USER variable so it just gets the judge number.  This was added for running multiple judges on the same machine different judge unix login accounts, and pinning them to different CPU's.  If the account name is not "judge#", for example, it will use CPU 0, which is fine if only one judge is running (or admin) on the system.  We usually only run one client application at-a-time on a machine, so that's ok.

clevengr: Holy Cow, no wonder I couldn't find any documentation on that; it's pretty esoteric.  (I'm sure it's documented _somewhere_... but I couldn't find it.)

So now I have a new question:  it appears this is dependent on the ubuntu user login being named "judgeX" -- is that true?

And again, what happens if it is an Admin executing the submission?

scripts/pc2sandbox.sh
```diff
@@ -158,7 +212,12 @@ DEBUG echo ...done.
 if test -d $PC2_SANDBOX_CGROUP_PATH
 then
 	DEBUG echo Removing existing sandbox to start clean
-	rmdir $PC2_SANDBOX_CGROUP_PATH
+	KillcgroupProcs
+	if ! rmdir $PC2_SANDBOX_CGROUP_PATH
+	then
+		DEBUG echo Cannot remove previous sandbox: $PC2_SANDBOX_CGROUP_PATH
+		exit $FAIL_INVALID_CGROUP_INSTALLATION
```
clevengr, line 219: Is this really an "Invalid CGroup Installation" issue?  It seems to me that this is more of a NEW condition introduced by the updates of the PR... in which case I would think a separate error/exit code indicating more precisely the condition for the termination would be better.

scripts/pc2sandbox_interactive.sh
```diff
@@ -308,8 +361,9 @@ TIMELIMIT_US=$((TIMELIMIT * 1000000))
...
 ulimit -t $TIMELIMIT
 
-REPORT Setting maximum user processes to $MAXPROCS + whatever the user is currently using
-ulimit -u $((MAXPROCS+`ps -T -u $USER | wc -l`))
+MAXPROCS=$((MAXPROCS+`ps -T -u $USER | wc -l`))
+REPORT Setting maximum user processes to $MAXPROCS 
+ulimit -u $MAXPROCS
 
 # Remember wall time when we started
 starttime=`GetTimeInMicros`
```
clevengr, line 369: With the above command being located here, the user will be charged for all the time it takes to execute the statements between here and the `setsid taskset` command (below).  If REPORT and/or REPORT_DEBUG are enabled, that could be significant. it seems like this statement  belongs _right before_ the call to `setsid taskset`...

scripts/pc2sandbox_interactive.sh
```diff
@@ -391,35 +447,46 @@ do
 		# Get wall time - we want it as close as possible to when we fetch the cpu time so they stay close
 		# since the cpu.stat includes the time this script takes AFTER the submission finishes.
 		endtime=`GetTimeInMicros`
```
clevengr, line 449: The comment above notwithstanding, the placement of the call to `GetTimeInMicros` here means that (especially with the addition of the call to `KillChildProcs`, above) the user is being charged for not only the time it takes to kill the child processes, but also the time it takes to do the REPORT calls, above.  It seems like the call to `GetTimeInMicros` should come as close as possible to the `wait` (inside the `while` loop, above).   (I'm not sure of the ramifications of moving it up, but I _am_ sure that with the call where it is the user is begin charged for system overhead...)

scripts/pc2sandbox.sh
```diff
@@ -227,32 +290,47 @@ COMMAND_EXIT_CODE=$?
...
 kills=`grep oom_kill $PC2_SANDBOX_CGROUP_PATH/memory.events | cut -d ' ' -f 2`
 
+KillChildProcs
+
+# Get cpu time used.
+cputime=`grep usage_usec $PC2_SANDBOX_CGROUP_PATH/cpu.stat | cut -d ' ' -f 2`
+
+# Get wall time - we want it as close as possible to when we fetch the cpu time so they stay close
+# since the cpu.stat includes the time this script takes AFTER the submission finishes.
+endtime=`GetTimeInMicros`
```
clevengr, line 300: With this call to `GetTimeInMicros` located here, the user is being charged for the time taken by `KillChildProcs` and also the `grep` which gets the cputime.  It seems like this call (and hence probably also the call to compute walltime) belong closer to the `wait`, above.

scripts/pc2sandbox.sh
```diff
@@ -189,8 +248,9 @@ TIMELIMIT_US=$((TIMELIMIT * 1000000))
...
 ulimit -t $TIMELIMIT
 
-DEBUG echo setting maximum user processes to $MAXPROCS + whatever the user is currently using
-ulimit -u $((MAXPROCS+`ps -T -u $USER | wc -l`))
+MAXPROCS=$((MAXPROCS+`ps -T -u $USER | wc -l`))
+DEBUG echo setting maximum user processes to $MAXPROCS
+ulimit -u $MAXPROCS
 
 # Remember wall time when we started
 starttime=`GetTimeInMicros`
```
clevengr, line 256: With this call being located here, the user is being charged for all of the time taken below to execute debugging output (if enabled), but also for the time to do stuff like set the `trap` handler.  It seems like this call should be made as closely as possible to the `setsid taskset`...

clevengr (approved):
I have completed a code review of the PR. I've made a number of suggestions for improvement (including a couple that relate to inaccurate computation of submission resource usage), but nothing that rises to "refuse to approve" level.  I'm sure you (JohnB) will look at the comments and consider them as appropriate.

I have NOT done an execution test of the PR.  I intend to do so, but given the amount of time it's likely to take me to do this, and the fact that we are in "crunch time" with respect to producing a distribution which can be tested at GNY, I'm going ahead and marking the PR "Approved".  I would rather see it merged and used/tested in a complete PC2 Distribution at GNY than have to deal with post-GNY merges which we then have no real live contest on which to test.

scripts/pc2sandbox.sh
johnbrvc: From above: "**If the account name is not "judge#", for example, it will use CPU 0, which is fine if only one judge is running (or admin) on the system. We usually only run one client application at-a-time on a machine, so that's ok**."

scripts/pc2sandbox.sh
johnbrvc: How about "FAIL_SANDBOX_ERROR" which is one of the errors?  That error is used if the PID can't be added to a cgroup,  Since the case we are talking about deals with an opposite problem (namely, can't remove the cgroup because there are processes in it, presumably), it indicates some sort of general "sandbox error".

scripts/pc2sandbox_interactive.sh
johnbrvc: This is wall time, not CPU time.  We use ulimit -t to limit CPU time, so submission has it's own "cpu time" separate from script (and every other process).  So, yes, the wall time may be off slightly, but it won't cause the submission to be terminated.

scripts/pc2sandbox_interactive.sh
johnbrvc: Again, the user is not "charged" for the time.  We **REPORT** the time in the summary is all; no child process will be killed due to this value.

scripts/pc2sandbox.sh
johnbrvc: See above.  Not "charged".  It is for reporting to the log file only.

scripts/pc2sandbox.sh
johnbrvc: Repeat

johnbrvc:
> s

The "`wf_candidate5`" branch on my fork has these fixes in it, and we used it to run the EAPC contest on 09/23/23 successfully.
I am going to make some changes suggested by @troy2914 w/r/t `kill `vs. `/bin/kill`

troy2914 (approved):
seems good.

Issue #607, opened 2022-11-13
Add an interim compile time limit per language


lane55:
**Is your feature request related to a problem?** 

Yes, some team's source takes an excessive amount of time to compile
(ex. large source files, or requiring large amounts of time to compile)

**Feature Description**: 

Add a global compile time limit in ms.

Return a CE ??  if a team's srouce takes too long to compile

**Have you considered other ways to accomplish the same thing?** 

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2

**Additional context**:

lane55:
Analysis: what judgement/response should pc2 give for the limit exceeded?

lane55:
Add Next Contest to provide a way like DOMJudge to limit compile time.

clevengr:
I thought we were going to use "Next Contest" to mean "this item is critical to be done before the next contest".  I don't consider this item "critical", in that (1) we have a known, easy workaround (the Primary will report "CE" but the Shadow will continue with the execution cycle, so we'll get a mismatch which is easy to understand), and (2) it happens _EXTREMELY RARELY_ (I think I've seen it maybe once in all the time we've been using a CCS that supports it.   These two factors lead me to believe this issue does not rise to the level of "critical to fix before the next contest".

johnbrvc:
> I thought we were going to use "Next Contest" to mean "this item is critical to be done before the next contest". I don't consider this item "critical", in that (1) we have a known, easy workaround (the Primary will report "CE" but the Shadow will continue with the execution cycle, so we'll get a mismatch which is easy to understand), and (2) it happens _EXTREMELY RARELY_ (I think I've seen it maybe once in all the time we've been using a CCS that supports it. These two factors lead me to believe this issue does not rise to the level of "critical to fix before the next contest".

I believe there were 5 or 6 instances of it during the NAC Dress, and 1 during the NAC Real this year.

clevengr:
>> I thought we were going to use "Next Contest" to mean "this item is critical to be done before the next contest". I don't consider this item "critical", in that (1) we have a known, easy workaround (the Primary will report "CE" but the Shadow will continue with the execution cycle, so we'll get a mismatch which is easy to understand), and (2) it happens EXTREMELY RARELY (I think I've seen it maybe once in all the time we've been using a CCS that supports it. These two factors lead me to believe this issue does not rise to the level of "critical to fix before the next contest".

[...]

Issue #425, opened 2022-05-30
CCS add time limit on compilation step 


lane55:
**Is your feature request related to a problem?** 

Yes, during the NAC a team had a compilation that took more than 5 min (then it was killed)

**Feature Description**: 

Add compilation timeout (global) 
If the timeout occurs then return a Compilation Error

**Have you considered other ways to accomplish the same thing?** Please provide a clear and concise description of any alternative solutions or features you've considered.  
...

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, please describe.  
...

**Additional context**: please provide any additional information that will help us in considering your feature request.  
...

troy2914:
Also this should address memory limit as well.

PR #813, opened 2023-09-21
I607 Interim fix for issue #607 "Add a compile time limit".  

johnbrvc:
This is a temporary fix to deal with runaway compiles.  It makes use of a shell script shim to be inserted before any compile command that you wish to limit.

Adds a shell script to the scripts folder called `dotimedcompile.sh`.  This script provides a mechanism to limit the CPU time and file sizes for compiles.  It is meant to be prepended to existing compile commands, eg:

`dotimedcompile.sh 30 128 /usr/bin/g++ -g -O2 -std=gnu++20 -static ...
`
This temporarily addresses #607.  607 will be closed by this PR, leaving PR 425 still open which will be addressed later when a non-interim fix is coded.

Ubuntu 22.04
Java 1.8

1) From the administrator client, modify a language as shown above.  
2) Make a C++ submission using the attached program.
[stress_test.txt](https://github.com/pc2ccs/pc2v9/files/12690511/stress_test.txt)
3) Note that it will compile for a while but will eventually be killed due to exceeding the time limit


troy2914 (approved):
Seems reasonable for now.

clevengr:
As I understand it, the usage of this script would be to prepend it to a given _language_ definition.  However, it is accepting a hard-coded numeric argument for compilation time, which is a per-_problem_ attribute (see the CLICS specification for the Problem Package Format; specifically, the [limits section](https://icpc.io/problem-package-format/spec/problem_package_format#limits) of the `problem.yaml` file).  If a contest is configured with problems having different compilation time limits, the approach taken in the PR will not work.

Is there an existing substitution variable that could be used (instead of a hard-coded time limit) to solve this issue?

troy2914:
> As I understand it, the usage of this script would be to prepend it to a given _language_ definition. However, it is accepting a hard-coded numeric argument for compilation time, which is a per-_problem_ attribute (see the CLICS specification for the Problem Package Format; specifically, the [limits section](https://icpc.io/problem-package-format/spec/problem_package_format#limits) of the `problem.yaml` file). If a contest is configured with problems having different compilation time limits, the approach taken in the PR will not work.
> 
> Is there an existing substitution variable that could be used (instead of a hard-coded time limit) to solve this issue?

There is a {:timelimit} expansion.

johnbrvc:
> As I understand it, the usage of this script would be to prepend it to a given _language_ definition. However, it is accepting a hard-coded numeric argument for compilation time, which is a per-_problem_ attribute (see the CLICS specification for the Problem Package Format; specifically, the [limits section](https://icpc.io/problem-package-format/spec/problem_package_format#limits) of the `problem.yaml` file). If a contest is configured with problems having different compilation time limits, the approach taken in the PR will not work.
> 
> Is there an existing substitution variable that could be used (instead of a hard-coded time limit) to solve this issue?

There is currently no way use a specific compile time limit as specified in the `problem.yaml` file.  As the title of the PR indicates, this is an **_INTERIM_** fix.  When the real fix implemented, it will, of course, obey the CLICS spec for per-problem limits of `compilation_time `and `compilation_memory`.  Until then, you'll have to use the `timelimit `as @troy2914 pointed out, or, hardcode a constant compile time for all problems in the language definition.

troy2914:
> > As I understand it, the usage of this script would be to prepend it to a given _language_ definition. However, it is accepting a hard-coded numeric argument for compilation time, which is a per-_problem_ attribute (see the CLICS specification for the Problem Package Format; specifically, the [limits section](https://icpc.io/problem-package-format/spec/problem_package_format#limits) of the `problem.yaml` file). If a contest is configured with problems having different compilation time limits, the approach taken in the PR will not work.
> > Is there an existing substitution variable that could be used (instead of a hard-coded time limit) to solve this issue?
> 
> There is currently no way use a specific compile time limit as specified in the `problem.yaml` file. As the title of the PR indicates, this is an **_INTERIM_** fix. When the real fix implemented, it will, of course, obey the CLICS spec for per-problem limits of `compilation_time `and `compilation_memory`. Until then, you'll have to use the `timelimit `as @troy2914 pointed out, or, hardcode a constant compile time for all problems in the language definition.

as you noted there is no support to load the compile timeout or store the compile output,  it should be set to the highest of the system and problem level compile limits.

clevengr (approved):
Approving based on the understanding that this is an "interim" fix, for #607 and we still have #425 regarding the need for a full implementation (including per-problem, and including both YAML and UI versions).

PR #825, opened 2023-10-11
copy non xsl files from data/xsl to the html dirs.

### LINKED ISSUE #688: Automate copying of non-xsl files into target scoreboard directories

Yes.  When the scoreboard generates the html there are
needed files .css and .png files.  Those files must
be manually copied.

**Feature Description**: 

For all non .xsl files in the data/xml copy those files into
the html and public_html directories, same directories as
the directories where the html files are created.

Manually done.

Each time the html is generated, copy the non xsl files.

Consider not copying the files if the files are identical.
For exmaple
02/04/2023  07:42 PM             2,070 standings.css

if the file in xsl is a differnt time or size, then update the standings.css
02/04/2023  07:55 PM             2,555 standings.css

**Additional context**:

troy2914:
with some intelligence.

eg into ./html and ./public_html by default.

fixes #688

start server loading a contest from samps/contests
start board
you should see the standing.css that is in data/xsl be copied to html and public_html directories.


src/edu/csus/ecs/pc2/ui/board/ScoreboardCommon.java
```diff
@@ -102,6 +189,11 @@ public void generateOutput(String xmlString, String xslDir, String outputDir, Lo
                 } catch (Exception e) {
                     log.log(Log.WARNING, "Trouble transforming " + xslFilename, e);
                 }
+            } else {
+                if (new File(xslDir + File.separator+ xslFilename).isFile()) {
+                    // not xsl, and a file, so just copy it**
```
johnbrvc, line 194: What does the trailing ** mean?  "copy it**" ?

johnbrvc (approved):
Looks good to me.
Seems to work.  Maybe the `index.xsl `should be changed to include the `standings.css` (link href ...) instead of "including the standings.css" file contents, like it does now.  Since we have this feature, might as well use it.

src/edu/csus/ecs/pc2/ui/board/ScoreboardCommon.java
troy2914: ** means only if needed :)

johnbrvc (approved):

data/xsl/standings.css
```diff
@@ -0,0 +1,117 @@
+// $Id$
```
clevengr, line 1: Isn't the above line a reference to an obsolete source control system?  Does it actually get used somewhere?

clevengr (approved):
I reviewed the code changes and they all seem to make sense; I started a Server loaded with a sample contest and submitted several runs (some no, some yes) and observed that the `standings.css` file was copied into both the `html` and `public_html` folders, and that the`index.html` files in both those folders used the appropriate .css coloring for No, Pending, and Yes.  I approved the PR.

PR #836, opened 2023-10-30
i835: add EWTeam deprecation notices to ewrun and ewrun.bat scripts.

### LINKED ISSUE #835: Deprecate EWTeam in preparation for removal
**Feature Description**: 

The `EWTeam` project has outlived its usefulness; we have no intention of supporting it and should be targeting it for removal.  

However, respect for our users (if indeed there actually are any still using it) would require that we first provide a "deprecation notice" prior to actually removing the project from distributions.  The "issue" being tracked here is "reporting EWTeam deprecation".

There are two options other than providing a "deprecation notice":
- Continue to leave the EWTeam project in distributions.
- Remove the EWTeam project without warning.

The first option doesn't seem reasonable because there is an implication (at least in some people's minds) that we would still be _supporting_ it -- which we have neither the resources nor a reason (given the existence of the WTI project) to do.
The second option would be exactly what we would do if we just wanted to say "we don't care about our users"... which we don't want to do.  So no, there aren't any _**good**_ options.

1. Add the following message to the console output produced by the `ewrun` and `ewrun.bat` files in the `scripts` folder of the `EWTeam` project:

```
"Warning: EWTeam is deprecated and will be removed in an upcoming version of PC^2.  It has been superseded by the "Web Team Interface (WTI)" project; see Appendix N in the PC^2 Contest Administrator's Guide for details on using WTI as a web-based interface for teams".
```

2. Add the updated files to the V9.9develop branch, so that they will become part of the very next public distribution (i.e., when V9.9 is actually published).

3. Submit a new Issue titled "Remove EWTeam from distribution", and schedule that to happen with V9.10 (i.e., a future release).

clevengr:

Adds "deprecation warning" messages to `ewrun` and `ewrun.bat`, the scripts which are used to start the EWTeam (actually, the "JavaBridge" which is the server for EWTeam browser requests).

Fixes #835 

Windows 11.

Execute the `ewrun` and `ewrun.bat` scripts at Linux and Windows command lines, respectively; verify that a "Warning" message about EWTeam deprecation is displayed prior to the script continuing.

### DIFF
--- a/projects/EWTeam/scripts/ewrun.bat
+++ b/projects/EWTeam/scripts/ewrun.bat
@@ -5,9 +5,9 @@ rem Purpose: Start the ew Team server/java bridge
 rem Author : pc2@ecs.csus.edu
 
 echo.
-echo Warning: EWTeam is deprecated and will be removed in an upcoming version of PC^2.  
+echo Warning: EWTeam is deprecated and will be removed in an upcoming version of PC^^2.  
 echo It has been superseded by the "Web Team Interface (WTI)" project; 
-echo   see Appendix N in the PC^2 Contest Administrator's Guide for details on using WTI as a web-based interface for teams.
+echo   see Appendix N in the PC^^2 Contest Administrator's Guide for details on using WTI as a web-based interface for teams.
 echo.
 
 rem Windows 2000 and beyond syntax
--- a/projects/EWTeam/scripts/ewrun.bat
+++ b/projects/EWTeam/scripts/ewrun.bat
@@ -4,6 +4,12 @@ REM Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, 
 rem Purpose: Start the ew Team server/java bridge 
 rem Author : pc2@ecs.csus.edu
 
+echo.
+echo Warning: EWTeam is deprecated and will be removed in an upcoming version of PC^2.  
+echo It has been superseded by the "Web Team Interface (WTI)" project; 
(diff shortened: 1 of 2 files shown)


troy2914 (approved):
looks good.

projects/EWTeam/scripts/ewrun.bat
```diff
@@ -4,6 +4,12 @@ REM Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane,
 rem Purpose: Start the ew Team server/java bridge 
 rem Author : pc2@ecs.csus.edu
 
+echo.
+echo Warning: EWTeam is deprecated and will be removed in an upcoming version of PC^2.
```
johnbrvc, line 8: You should use PC^^2 if you want to see the caret in echo commands.  Caret is the escape char for the windows command interpreter.  If you "quote" the string, you'll see the quotes... So, ^^ works.

johnbrvc (changes requested):
Change PC^2 to PC^^2 in batch file (2 places).

projects/EWTeam/scripts/ewrun.bat
clevengr: > You should use PC^^2 if you want to see the caret in echo commands. Caret is the escape char for the windows command interpreter. If you "quote" the string, you'll see the quotes... So, ^^ works.

Done.  (In case anyone wonders, I DID test the batch file before submitting the PR.  However, I can't swear that it didn't say "PC2" instead of "PC^2"; I'm so used to seeing both forms that I could easily have overlooked that detail.)  I've submitted an update to the PR.

johnbrvc (approved):

PR #851, opened 2023-11-06
I850 - Validate that memory limit is non-negative in sandbox configuration for a problem

### LINKED ISSUE #850: Negative value for Memory Limit in EditProblemSandboxPane silents reverts to 0
**Describe the issue**:  
if you enter a negative value for the **Memory Limit** when editing the sandbox parameters for a problem, it silently converts that to zero when it updates the problem.  It should probably pop up a dialog saying "_Memory limit value must be greater than or equal to zero_".

**To Reproduce**: 

1. Edit a problem.
2. Select the **Sandbox** tab.
3. Specify a negative memory limit.
4. Update the problem.
5. View the **Memory Limit(MB)** column in the Problems tab on the _Administrator_ client.
6. Note the limit is 0.

**Expected behavior**: 
An error dialog should be presented to the user informing him that negative values are illegal.

**Actual behavior**: 
Negative values are silently changed to 0 in `Problem.setMemoryLimitMB()
`

**Environment**: 
All.

**Log Info**:  

**Screenshots**: 

**Additional context**: 
This bug was reported by John Clevenger while updating the Administrator's Guide for 9.9.

johnbrvc:
Add a test in `EditProblemPane.validateProblemFields() `to make sure that the memory limit specified on the **Sandbox** tab is valid.
Note that this test is ALWAYS performed, even if a sandbox is not selected for the problem.  It is possible that the user selected a sandbox, entered a negative value, then de-selected a sandbox.  We STILL want to issue an error here so we don't have any future issues with the memory limit.  Note that this scenario would require the user to perform the above sequence, so they went out of their way to put a negative value there, which we simply will not permit.

Fixes #850 
All.

1. Start an administrator client
2. Edit a problem.
3. Select the **Sandbox** tab.
4. Select the "_Use PC^2 Sandbox (not available on Windows)_" radio button.
5. Enter a negative value in the _Memory Limit (MB):_ edit text box.
6. Press the **Update** button.
7. Observe the error dialog that is presented.
image

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
+++ b/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
@@ -1883,6 +1883,17 @@ public class EditProblemPane extends JPanePlugin {
         }
         
         
+        // verify that the memory limit is a non-negative number 
+        try {
+            int memLimit = Integer.parseInt(getProblemSandboxPane().getPC2SandboxOptionMemLimitTextbox().getText().trim());
+            if (memLimit < 0) {
+                showMessage("Memory limit value must be greater than or equal to zero.");
+                return false;
+            }
+        } catch (NumberFormatException e) {
+            showMessage("Memory limit must be a non-negative number");
+            return false;
+        }
 
         // verify that if the PC2 Validator is selected, an option has been chosen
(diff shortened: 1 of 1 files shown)


clevengr (approved):
Verified that the change is very small and that it works.

PR #830, opened 2023-10-13
I 760 add results pane report

lane55:
testdata/FileComparisonUtilitiesTest/resultscompwork

### Description of what the PR does

On Admin, adds Results Compare tab for comparing primary ccs and pc2 results files.

### Issue which the PR addresses

Fixes #760

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)

test 1 - compare primary to pc2

start server 
start admin
On Results Compare
For Primary CCS Results Select Directory select testdata\FileComparisonUtilitiesTest\resultscompwork\results\domjudge
For PC2 Results Select Directory select testdata\FileComparisonUtilitiesTest\resultscompwork\results\pc2
Click Show Comparison button

The following will appear In Compare Results Summary text area
```
Summary:   All PASS - no differences found

 51 rows scoreboard.json
 19 rows awards.json
 51 rows results.tsv

Primary CCS results dir: C:\repos\laned55forkwip\testdata\FileComparisonUtilitiesTest\resultscompwork\results\domjudge
PC2 results dir        : C:\repos\laned55forkwip\testdata\FileComparisonUtilitiesTest\resultscompwork\results\pc2
```

test 2 - compare primary to pc2 with missing rows/data

start server 
start admin
On Results Compare
For Primary CCS Results Select Directory select testdata\FileComparisonUtilitiesTest\resultscompwork\results\domjudge
For PC2 Results Select Directory select testdata\FileComparisonUtilitiesTest\results\pc2data1
Click Show Comparison button

The following will appear In Compare Results Summary text area
```
Summary:   FAIL - there are 33  differences found

  0 differences scoreboard.json
  9 differences awards.json
 24 differences results.tsv

Primary CCS results dir: C:\repos\laned55forkwip\testdata\FileComparisonUtilitiesTest\resultscompwork\results\domjudge
PC2 results dir        : C:\repos\laned55forkwip\testdata\FileComparisonUtilitiesTest\resultscompwork\results\pc2data1

```

Issue #760, opened 2023-05-30
Add a feature to compare results files, export files, produce a results comprison report


lane55:
**Is your feature request related to a problem?** 

Yes, the EOC steps to compare the CCC results file pc2
export and compare files is not automated.

**Feature Description**: 

Ultimately create two tasks/output
1.  an export results file feature
2.  a report to compare the pc2 results with another CCS results files.
   2a.  input would be the other CCS results dir, ex nac23real/results/domjudge.

Optional is to compare existing results file by providing two
results dir, ex results/pc2 results/domjudge

Current results files to compare, from EOC doc
  https://docs.google.com/document/d/1uwAagoskpQ8TGqqt3yDsGXcwEZbCsvg5yYOcG_hwnRQ/edit#
are:
results.tsv
(results.csv file if/when available #351)
scoreboard.json
awards.json

**Have you considered other ways to accomplish the same thing?** 

Very manual, some script solutions.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?**

Add a new Report to export the files listed above.

On the Finalize tab add feature to compare the pc2 results snd domjudge results.

**Additional context**:

lane55:
Design requirements
Provide a new Pane that has
title and text field for output drectoy for results reports/files, add button with open dialog to pick dir
title and text field for other directory, primary results file, , add button with open dialog to pick dir
Add button to produce results compare report
Add What's this icon/component for the two text fields.

lane55:
For both text fields, save any changes local to user (add items to ClientSettings)

clevengr:
It appears that this issue, and the corresponding PR work being done on it, duplicates (actually, is a super-set of) Issue #592.  If so, #592 should probably be marked as a duplicate, and closed in favor of this issue.

clevengr:
It appears that this issue, and the corresponding PR work being done on it, duplicates (actually, is a super-set of) Issue #615. If so, #615 should probably be marked as a duplicate, and closed in favor of this issue.

PR #880, opened 2023-12-01
I878: Correct test data for results comparison

### LINKED ISSUE #878: Test data is wrong for PR #830 - add results pane report
**Describe the issue**:
The test procedure described in PR #830 does not give correct results for "**test 1 - compare primary to pc2**"
It reports that the scoreboards do not match.

**To Reproduce**: 
Just follow the "**test1**" procedure in PR #830 

**Expected behavior**:
 The results should indicate that the scoreboards match.

**Actual behavior**: 
The results indicate that the scoreboards do not match.  On Oct 19 when PR #830 was first reviewed, it did pass just fine.  The reason was, at the time, there was a bug in the code, specifically, `core.report.FileComparisonUtilities`:

```
            for (int rowNum = 0; rowNum < minLines; rowNum++) {
                TeamScoreRow firstRow = firstTeamScoreRows.get(rowNum);
                TeamScoreRow teamScoreRow = firstTeamScoreRows.get(rowNum);

```
Note that the code was comparing the SAME file's rows (`firstTeamScoreRows`) so there was no way it was going to fail.
This was **LATER** fixed to be:

```
            for (int rowNum = 0; rowNum < minLines; rowNum++) {
                TeamScoreRow firstRow = firstTeamScoreRows.get(rowNum);
                TeamScoreRow secondScoreRow = secondTeamScoreRows.get(rowNum);

```
on Nov 6, 2023 by commit [i760: PR fixes, improved error handling/display](https://github.com/pc2ccs/pc2v9/pull/830/commits/892141a4b24a93133f9f1b5ef0d9b057b58d1b80) :

This is what broke the test procedure since now `scoreboard.json` will never ever compare correctly.  Evidently, the PR was never fully  re-tested AFTER that commit.  

In addition, the PR test steps demonstrate how to show a failure case, ONLY had differences in "`awards.json`" and "`results.tsv`", not "`scoreboard.json`".  The `scoreboard.json`'s would ALWAYS compare correctly since it was comparing the same file.

The reason why the compare failed AFTER the above commit was because the PC2 version of the `...\pc2v9\testdata\FileComparisonUtilitiesTest\resultscompwork\results\pc2` file was generated using the (non-CLICS compliant) "**JSON 2016 Scoreboard**" report and not the (CLICS compliant) "**Scoreboard JSON**".  The primary difference in these is that the team ID property uses a different name:  `team `vs. `team_id`.  This would cause every line with a team ID to compare differently, thereby breaking the test procedure.  (This was not an issue before the above commit, since the code was comparing the exact same entry - the PC2 one - to itself).

**Environment**: 
All

**Log Info**:   

**Screenshots**:

**Additional context**: 
@johnbrvc hand-edited the  `...\pc2v9\testdata\FileComparisonUtilitiesTest\resultscompwork\results\pc2` file changing all "team" properties to "team_id" properties.  After doing this, and repeating the test procedure, the scoreboards compare correctly.  Really, what should happen is a REAL "**Scoreboard JSON**" `scoreboard.json` should be generated from the same contest from whence the **DOMjudge** `scoreboard.json` was generated.  It shouldn't make a difference since the only fields compared for the scoreboard are rank, team_id, total_time and num_solved.  No other properties (like problems[]) are even looked at - which is a whole **OTHER** issue.  @johnbrvc will create a PR including the hand-modified file.  The PR can be denied or accepted as others see fit, or, if someone wants to supply a genuine CLICS compliant scoreboard.json from PC2, that could be used.

johnbrvc:
The PC2 version of "correct" test data was generated using the wrong "_2016 JSON scoreboard_" report.  It should have simply used "_Scoreboard JSON_", which produces a **CLICS** compliant scoreboard JSON file.  Major difference: "**team**" property is "**team_id**" for **CLICS** compliant scoreboard.  The file in the commit was hand-edited to make that simple change.
See also #830 and #760 

This PR does not affect the public distribution package since the testdata folder is not included in that.  This PR only affects manual testing of #830.

Fixes #878 
Windows 11
JDK 8.1.321

See #830.
The instructions for "**test 1 - compare primary to pc2**" can be used to verify this PR.


troy2914 (approved):
I really don't like that that was not generated by code.

PR #895, opened 2023-12-10
i888 Use correct java compiler for JUnit tests

### LINKED ISSUE #888: Test units fail when run as part of a package build on Windows/MacOS
**Describe the issue**: 
It has been observed that several test units fail (see **Log Info** below) when run as part of the build process (as opposed to running them individually).  This only happens if the system you are running has the default executable path set so that the _JDK 1.8/JRE1.8_ is not before any newer version of a JDK or JRE.  You can determine this if you create a command prompt and type "`java -version`", for example:
```
C:\Windows\System32>java -version
openjdk version "11.0.16.1" 2022-08-12 LTS
OpenJDK Runtime Environment Microsoft-40648 (build 11.0.16.1+1-LTS)
OpenJDK 64-Bit Server VM Microsoft-40648 (build 11.0.16.1+1-LTS, mixed mode)
```
Note this is NOT a problem for the CI _Workflows_ in **Github**, since the path only contains the **Java 1.8 JDK/JRE**.

**To Reproduce**:
On a system where the **JDK/JRE 1.8** is NOT first on the path:
1) Clone a new copy of the latest develop branch
2) Create a new Eclipse project and import the newly cloned branch.
3) Right click on the "`build.xml`" and choose "_Run as..._"
4) Choose only: "**JarSkipBuldSubprojects**".  This will cause Eclipse to build PC2 only, and then run all the JUnits.
5) Observe that some tests fail.

**Expected behavior**: 
 All JUnit tests should pass.

**Actual behavior**: 
Several JUnit tests fail.  See Log Info below.

Windows 11.0
pc2v9.10 develop
JDK 1.8.0_321 + JDK-11.0.16.101

**Log Info**:  
[juniterrors..txt](https://github.com/pc2ccs/pc2v9/files/13601855/juniterrors.txt)

**Screenshots**:

**Additional context**: 
The problem stems from the fact that the Ant JUnit task specified in the build.xml file:
```
		<junit errorProperty="test.failed" failureProperty="test.failed" fork="yes">
                        <jvmarg value="-Djdk.crypto.KeyAgreement.legacyKDF=true"/>

```
If this is changed to:
```
		<junit errorProperty="test.failed" failureProperty="test.failed" fork="yes" jvm="C:/Program Files/Java/jdk-1.8/bin/java.exe">
                        <jvmarg value="-Djdk.crypto.KeyAgreement.legacyKDF=true"/>

```
Then the **JUnits** will all run correctly.  It seems that the _Ant JUnit_ task (according to multiple sources on the internet) do not listen to the Eclipse's default JRE, rather, it just uses "java" as the command to run the JUnit tests.  Whatever "java" comes up first on your default path is used to run the JUnit tests.  The "jvm" attribute in for the Ant junit task allows you to specify a different JVM to use.  Obviously, the path to java.exe should be whatever it is on the target system.  The one above is just an example from my system.

johnbrvc:
Since _Eclipse 4.17 (2020-09)_, **Java 11** (or newer) is the default JDK that ships with Eclipse.  Previously, it was **Java 8** (1.8). This PR attempts to track down the correct JDK to use for compiling the _JUnit_ test programs.  The JDK must match the currently running JVM in order to qualify.  A search of  the `JAVA_HOME/bin` and folders on the `PATH `environment variable are checked for the presence of the corresponding JDK (`javac[.exe]`). The code only checks for the presence of the Java version string in the folder name.  It could be expanded to actually attempt to run "`javac -version`" in each folder, but that shouldn't be necessary, and is highly inefficient.

The reason for this PR is that the JUnits requiring compiles were using Eclipse's JDK (Version >11), but attempting to execute the produced class files using JRE 1.8.  This lead to the errors described in #888.

Fixed #888 

Windows 11
Eclipse 2021-12, with JVM org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_17.0.1.v20211116-1657
JDK 1.8.0_351

Run all JUnits.  They should pass now.   Previously, the ones in #888 (ExecuteTest, among others) did not pass.

### DIFF
--- a/test/edu/csus/ecs/pc2/core/execute/ExecutableTest.java
+++ b/test/edu/csus/ecs/pc2/core/execute/ExecutableTest.java
@@ -164,6 +164,8 @@ public class ExecutableTest extends AbstractTestCase {
     private Language createLanguage(String autoFillLanguageTitle) {
 
         String[] values = LanguageAutoFill.getAutoFillValues(autoFillLanguageTitle);
+        // Make sure to use correct javac for the JVM in use.
+        SampleContest.fixJavaJDKPath(values);
         Language language = new Language(values[4]);
         // displayNameTextField.setText(values[0]);
         // compileCommandLineTextField.setText(values[1]);
--- a/test/edu/csus/ecs/pc2/core/model/SampleContest.java
+++ b/test/edu/csus/ecs/pc2/core/model/SampleContest.java
@@ -1,10 +1,12 @@
 // Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.model;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -100,6 +102,8 @@ public class SampleContest {
             "Michael", "Mohammad", "Mohammed", "Morgan", "Muhammad", "Nathan", "Noah", "Oliver", "Oscar", "Owen", "Reece", "Reuben", "Rhys", "Riley", "Robert", "Ryan", "Sam", "Samuel", "Sebastian",
             "Stanley", "Taylor", "Theo", "Thomas", "Toby", "Tyler", "William", "Zachary", };
 
+    // remember the compiler if we found it.  avoid LOTS of extra "javac -version" commands
+    private static String javaCompileCmd = null;
 
     public SampleContest() {
         super();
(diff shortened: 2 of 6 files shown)


troy2914:
Tested with jdk17 as first in my PATH (and only java in my PATH),  generated a `version 61.0` PracticeA.class
added my Amazon Correta 8 to the end my PATH /Library/Java/JavaVirtualMachines/amazon-corretto-8.jdk/Contents/Home/bin,
still generated a version 61.0 class.

troy2914:
btw, standard location on MacOS is for the jdk to be under 
/Library/Java/JavaVirtualMachines
where I have
```
/Library/Java/JavaVirtualMachines/jdk1.8.0_144.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_151.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_201.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_45.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_60.jdk
/Library/Java/JavaVirtualMachines/jdk1.8.0_65.jdk
```  
along with the amazon-corretto one

johnbrvc:
Based on @troy2914 's comments, and what I do see on Linux (JDK not located in a folder that contains the version number), we may have to what I suggested in the PR comment, that is, execute "`javac -version`" to see what version each **javac** is.  I'm thinking perhaps if we make the lookup of **javac** _static_ and save it to a _static_ member variable, then we only have to find the right **javac** once per **JUnit** test.

johnbrvc:
> Tested with jdk17 as first in my PATH (and only java in my PATH), generated a `version 61.0` PracticeA.class added my Amazon Corretto 8 to the end my PATH /Library/Java/JavaVirtualMachines/amazon-corretto-8.jdk/Contents/Home/bin, still generated a version 61.0 class.

Changed so it actually runs "javac -version" on every javac or javac.exe on the JAVA_HOME/bin and PATH until it finds one that matches instead of just looking for the version in the folder string.

johnbrvc:
> Based on @troy2914 's comments, and what I do see on Linux (JDK not located in a folder that contains the version number), we may have to what I suggested in the PR comment, that is, execute "`javac -version`" to see what version each **javac** is. I'm thinking perhaps if we make the lookup of **javac** _static_ and save it to a _static_ member variable, then we only have to find the right **javac** once per **JUnit** test.

Implemented this.

PR #834, opened 2023-10-29
I 833 upgrade WTI angular components

clevengr:
### Description of what the PR does

Upgrades the WTI to use Angular 16 and corresponding support packages.  Specifically:
- updates the `package.json` file in the `WTI-UI` project to specify Angular 16 and the appropriate related dependencies.
- updates the `angular.json` file in the `WTI-UI` project to remove deprecated (no longer supported) references.
- updates various _typescript_ files (NgModules) in the WTI-UI project to comply with new Angular rules.

Does NOT change any functionality of the WTI; just upgrades the build process so that changes made for other WTI issues can be built properly.

The PR pushes the changes onto the "v9.10" sub-branch of develop.

### Issue which the PR addresses
Fixes #833

### Environment in which the PR was developed (OS,IDE, Java version, etc.)

Window 11, Java 1.8.0_271

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)

The test is to verify that the upgraded version of WTI in the PR works the same way as the old version did.  The first step is to obtain a new build of WTI.

#### If you are planning to test the ability to build on your local machine, following these steps:

- Verify you have `nodejs` version 16 installed (type `node -v`).   If you don't have version 16, install it from https://nodejs.org/dist/latest-gallium/
- Run the Ant script `package.xml` in the `pc2v9` project (this generates a complete new PC2 distribution; this is necessary to generate an updated `pc2.jar` file for use in the WTI distribution).

#### If you are planning to use the already-built version in GitHub, follow these steps:

Download the complete PC2 build (including the WTI) from the GitHub `actions` page for the latest successful build of Issue #833.  (Note that artifacts are only retained by GitHub for 90 days from the time they are built.)

#### Once you have an updated PC2 distribution from one of the above two methods, follow these steps to test it:

- Copy the new `WebTeamInterface-1.1.zip` (or `.tar.gz`) file from the `projects` folder in the distribution to an empty folder.
- Unzip the WTI file.
- Start a PC2 server, loading a sample contest such as "SumitHello" 
- Start a PC2 Admin.
  - Ensure there is a "scoreboard2" account (if not, create one).
  - Ensure there is at least one team account.
  - Ensure there is at least one judge account configured to "autojudge" the problems in the contest.
  - Ensure there is at least one judge account NOT configured to "autojudge" anything.
- Start an AutoJudge.
- Start a non-autojudge.
- Start the contest.
[...]

PR #901, opened 2023-12-16
I631 Updated WTI and Team GUI backend to prevent duplicate filename submissions in additional files

### LINKED ISSUE #631: Duplicate files in team submission throws exception in web services /submissions/xx/files endpoint
**Describe the issue**: 

If a team makes a submission in which they specify the same file multiple times (e.g. if the "main" file and one of the "Auxiliary" files are the same) it causes an exception when referencing the web services "/contest/submissions/xx/files" endpoint (which is supposed to return a ZIP file containing the files in the submission). 

The exception happens during the zipping step in the `SubmissionService` class; it occurs because a ZIP file cannot contain multiple copies of the same file.  Specifically, the exception occurs at line 265 of method `createZip()`.  

**To Reproduce**:

- Start a server with a contest containing at least one problem and one team.
- Start a team client.
- Submit a run where the same file has been added to the Submission file list more than once.
- Start an Event Feed client.
- Use the EF client "Web Services" tab to start the web server.
- Open a browser to the URL `https://localhost:50443/contest/submissions/xx/files`,  where "localhost" represents the IP address of the web server and "xx" is the number of the submission.

**Expected behavior**: 

The web service should return a zip of the files making up the submission.

**Actual behavior**:

The web service throws an exception (which can be seen in the FEEDER log), and returns "Not Found".

**Environment**:
All

**Log Info**:  

An example of the Event Feeder exception is shown below.

```
221203 121537.013|INFO|qtp294235476-45|log|Got a reply from the server...
221203 121537.013|INFO|qtp294235476-45|log|Returning runFiles: ["MainFile":"hello.java","OtherFiles":["name":"hello.java","name":"hello.java"]]
221203 121537.023|FINER|qtp294235476-45|getSubmissionFiles|THROW
|java.util.zip.ZipException: duplicate entry: hello.java
|    at java.util.zip.ZipOutputStream.putNextEntry (ZipOutputStream.java:232)
|    at edu.csus.ecs.pc2.services.web.SubmissionService.createZip (SubmissionService.java:265)
|    at edu.csus.ecs.pc2.services.web.SubmissionService.getSubmissionFiles (SubmissionService.java:215)
|    at sun.reflect.NativeMethodAccessorImpl.invoke0 (NativeMethodAccessorImpl.java:-2)
|    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
```

**Screenshots**: 

The following shows an example of a team submitting a Run with duplicate file names, which will cause the above exception in the web services `/contest/submissions/xx/files` endpoint.  (The example shows the same file submitted THREE times, but all it takes is TWO occurences of the same file to generate the exception.)

image

SamanwaySadhu:
- Updated class `SubmitRunPane` and put a check in function `testOrSubmitRun` to prevent submission of Files with duplicate names in Team GUI.
- Added a check inside `new-run.component.ts` of WTI-UI to prevent submission of Files with duplicate names in WTI.

Fixes #631 

Windows 10, Eclipse 2021-12 R, JDK 8u381 (1.8)

#### To test the WTI-UI.
- Run the Ant script `package.xml` in the `pc2v9` project (this generates a complete new PC2 distribution; this is necessary to generate an updated `pc2.jar` file for use in the WTI distribution). Make sure there is `nodejs` version 16 installed before, if not follow the instructions [here](https://github.com/pc2ccs/pc2v9/pull/834).
- Copy the new `WebTeamInterface-1.1.zip` (or `.tar.gz`) file from the projects folder in the distribution to an empty folder.
Unzip the WTI file.
- Start a PC2 server, loading a sample contest such as "SumitHello".
- Start a PC2 Admin.
    - Ensure there is a "scoreboard2" account (if not, generate one).
    - Ensure there is at least one team account.
- Start the contest.
- Go into the unzipped WTI folder and execute the command `./bin/pc2wti.bat` (or `./bin/pc2wti` in Linux) to start the WTI server (API).
- Open a browser to http://localhost:8080/. You should see the standard WTI "Balloon Login" screen.
- Login as a team.
- Try to submit a problem where you add some files with the same name as each other or the Main File name in Additional Files.
- It would not submit and display the message `You may not submit multiple files with the same name`.
Capture

#### To test the Team GUI
- Run the Ant script `build.xml` in the `pc2v9` project.
- Start a PC2 server, loading a sample contest such as "SumitHello".
- Start a PC2 Admin.
- Start the contest.
- Start a PC2 Team.
- Try to submit a problem where you add some files with the same name as each other or the Main File name in Additional Files.
- It would not submit and display the message `You may not submit multiple files with the same name`.
Capture1

### Additional Context
It is also possible to add the checks in `ServerConnection` class and throw an Exception but in that case the WTI-UI will display a generic error message and teams will not know why their submission was invalid. It can still be added as additional fail-safe to the system.

### DIFF
--- a/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
+++ b/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
@@ -203,22 +203,34 @@ export class NewRunComponent implements OnInit, OnDestroy {
   //returns true if there are duplicate filenames among any of the specified FileSubmissions; false if all unique.
   private filenameContainsDuplicates(mainfile: FileSubmission, additionalFiles: FileSubmission []): boolean {
     let mainfilename = mainfile.fileName;
-    let l = mainfilename.indexOf("/");
-    if (l == -1) l = mainfilename.indexOf("\\");
-    if (l > -1) mainfilename = mainfilename.substring(l + 1);
+    let l = mainfilename.lastIndexOf("/");
+    if (l == -1) {
+      l = mainfilename.lastIndexOf("\\");
+    }
+    if (l > -1) {
+      mainfilename = mainfilename.substring(l + 1);
+    }
     for (let i = 0; i < additionalFiles.length; i++) {
       let otherfilename = additionalFiles[i].fileName;
-      l = otherfilename.indexOf("/");
-      if (l == -1) l = otherfilename.indexOf("\\");
-      if (l > -1) otherfilename = otherfilename.substring(l + 1);
+      l = otherfilename.lastIndexOf("/");
+      if (l == -1) {
+        l = otherfilename.lastIndexOf("\\");
+      }
+      if (l > -1) {
+        otherfilename = otherfilename.substring(l + 1);
+      }
       if (mainfilename === otherfilename) {
         return true;
       }
       for (let j = i + 1; j < additionalFiles.length; j++) {
         let otherfilename2 = additionalFiles[j].fileName;
-        l = otherfilename2.indexOf("/");
-        if (l == -1) l = otherfilename2.indexOf("\\");
-        if (l > -1) otherfilename2 = otherfilename2.substring(l + 1);
+        l = otherfilename2.lastIndexOf("/");
+        if (l == -1) {
+          l = otherfilename2.lastIndexOf("\\");
+        }
+        if (l > -1) {
+          otherfilename2 = otherfilename2.substring(l + 1);
+        }
         if (otherfilename2 === otherfilename) {
           return true;
         }
--- a/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
+++ b/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
@@ -122,6 +122,9 @@ export class NewRunComponent implements OnInit, OnDestroy {
 		//pop up an error dialog
 	    this._uiHelper.alert('File names may not contain spaces');
 	    console.error('One or more submitted file contains a space in its filename');
+  } else if (this.filenameContainsDuplicates(this.mainFile, this.additionalFiles)) {
+      this._uiHelper.alert('You may not submit multiple files with the same name');
+	    console.error('One or more submitted file have the same filename');
 	} else {
 		//submit the run
 	    this._teamService.submitRun(model)
@@ -196,6 +199,34 @@ export class NewRunComponent implements OnInit, OnDestroy {
 	//none of the specified FileSubmission files contains a space in its name
 	return false;
 	}
+
+  //returns true if there are duplicate filenames among any of the specified FileSubmissions; false if all unique.
+  private filenameContainsDuplicates(mainfile: FileSubmission, additionalFiles: FileSubmission []): boolean {
+    let mainfilename = mainfile.fileName;
+    let l = mainfilename.indexOf("/");
+    if (l == -1) l = mainfilename.indexOf("\\");
+    if (l > -1) mainfilename = mainfilename.substring(l + 1);
+    for (let i = 0; i < additionalFiles.length; i++) {
+      let otherfilename = additionalFiles[i].fileName;
+      l = otherfilename.indexOf("/");
+      if (l == -1) l = otherfilename.indexOf("\\");
+      if (l > -1) otherfilename = otherfilename.substring(l + 1);
+      if (mainfilename === otherfilename) {
+        return true;
+      }
+      for (let j = i + 1; j < additionalFiles.length; j++) {
+        let otherfilename2 = additionalFiles[j].fileName;
+        l = otherfilename2.indexOf("/");
+        if (l == -1) l = otherfilename2.indexOf("\\");
+        if (l > -1) otherfilename2 = otherfilename2.substring(l + 1);
+        if (otherfilename2 === otherfilename) {
+          return true;
+        }
+      }
+    }
+    //none of the specified FileSubmission files contains duplicate names
+    return false;
+  }
 	
 	//returns a string intended to identify the platform on which the browser is running
 	private getOSName() : string {
--- a/src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
+++ b/src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
@@ -483,15 +483,13 @@ public class SubmitRunPane extends JPanePlugin {
         
         if (additionalFilesMCLB.getRowCount() > 0) {
 
-            int l = filename.lastIndexOf('/');
-            if (l == -1) l = filename.lastIndexOf('\\');
             String mainfilename = filename;
+            int l = filename.lastIndexOf(File.separatorChar);
             if (l > -1) mainfilename = filename.substring(l + 1);
 
             for (int i = 0; i < additionalFilesMCLB.getRowCount(); i++) {
                 String otherfilename = (String) additionalFilesMCLB.getRow(i)[0];
-                l = otherfilename.lastIndexOf('/');
-                if (l == -1) l = otherfilename.lastIndexOf('\\');
+                l = otherfilename.lastIndexOf(File.separatorChar);
                 if (l > -1) otherfilename = otherfilename.substring(l + 1);
                 if (otherfilename.equals(mainfilename)) {
                     showMessage("You may not submit multiple files with the same name");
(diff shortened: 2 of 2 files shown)


src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
```diff
@@ -481,7 +481,37 @@ protected void testOrSubmitRun(boolean submitTheRun) {
             return;
         }
         
-        if (additionalFilesMCLB.getRowCount() > 0){
+        if (additionalFilesMCLB.getRowCount() > 0) {
+
+            int l = filename.lastIndexOf('/');
```
johnbrvc, line 486: Instead of testing for both / and \, you can use File.separator or File.separatorChar

src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
```diff
@@ -481,7 +481,37 @@ protected void testOrSubmitRun(boolean submitTheRun) {
...
+            if (l == -1) l = filename.lastIndexOf('\\');
+            String mainfilename = filename;
+            if (l > -1) mainfilename = filename.substring(l + 1);
+
+            for (int i = 0; i < additionalFilesMCLB.getRowCount(); i++) {
+                String otherfilename = (String) additionalFilesMCLB.getRow(i)[0];
+                l = otherfilename.lastIndexOf('/');
```
johnbrvc, line 493: Same comment about `File.separator `and` File.separatorChar`

src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
```diff
@@ -481,7 +481,37 @@ protected void testOrSubmitRun(boolean submitTheRun) {
...
+                if (l > -1) otherfilename = otherfilename.substring(l + 1);
+                if (otherfilename.equals(mainfilename)) {
+                    showMessage("You may not submit multiple files with the same name");
+                    log.warning("Found multiple files with same filename");
+                    return;
+                }
+                
+                for (int j = i + 1; j < additionalFilesMCLB.getRowCount(); j++) {
+                    String otherfilename2 = (String) additionalFilesMCLB.getRow(j)[0];
+                    l = otherfilename2.lastIndexOf('/');
```
johnbrvc, line 504: Same comment about` File.separator` and` File.separatorChar`

projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
```diff
@@ -196,6 +199,34 @@ export class NewRunComponent implements OnInit, OnDestroy {
 	//none of the specified FileSubmission files contains a space in its name
 	return false;
 	}
+
+  //returns true if there are duplicate filenames among any of the specified FileSubmissions; false if all unique.
+  private filenameContainsDuplicates(mainfile: FileSubmission, additionalFiles: FileSubmission []): boolean {
+    let mainfilename = mainfile.fileName;
```
johnbrvc, line 205: I'm not the typescript expert, but you may wish to check if typescript has a similar thing regarding file separators for pathname components..

johnbrvc (changes requested):
Built and tested.  Tried many different combinations of files and it works as expected.  Both the WTI and the fat Team client both indicate you can not submit files with the same basename.

The only issues are the comments I made about the checking of / and \\ that should be fixed before this is approved.

clevengr:
> Built and tested. Tried many different combinations of files and it works as expected. Both the WTI and the fat Team client both indicate you can not submit files with the same basename.
> 
> The only issues are the comments I made about the checking of / and \ that should be fixed before this is approved.

I agree; Java code should use `File.separator` rather than checking explicitly for `\` or `/`.

projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
clevengr: I haven't tried it myself, but a quick search led me to [this page](https://dirask.com/posts/TypeScript-Node-js-get-platform-specific-path-segment-separator-pJAboD), which says that the Typescript equivalent to Java's `File.separator` property is the following:
```typescript
import * as path from 'path';
```
and then reference the property `path.sep` to get the platform-specific file separator.  You can also do things like
```typescript
filePath.split(path.sep)
```
to separate a full path into its components based on the value of `path.sep`.

I think this should be put to use in the changes to new-run.component.ts.

src/edu/csus/ecs/pc2/ui/SubmitRunPane.java
SamanwaySadhu: Updated and tested for Windows, need to be tested for Linux and Mac

projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
SamanwaySadhu: To use this, we need to install the package `npm install --save @types/node`. It can be added to the `package.json`. Without it the build fails.

johnbrvc: So, I guess you are asking if it's OK to add that to the packages that **node** installs.  Did you actually try what you are suggesting and confirm it works?  If so, then I don't see a problem with adding it unless @clevengr does.

SamanwaySadhu: @types/node is already present in the `package.json`. Need to check why it still fails.

SamanwaySadhu: It is already included in the package.json. I tried a few different things like moving the @types/node to dependencies from devDependencies in package.json, adding "types": ["node"] in compilerOptions of tsconfig.json but everytime got error TS2307: Cannot find module 'path' or its corresponding type declarations.

projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
```diff
@@ -196,6 +199,34 @@ export class NewRunComponent implements OnInit, OnDestroy {
...
+    let l = mainfilename.indexOf("/");
+    if (l == -1) l = mainfilename.indexOf("\\");
+    if (l > -1) mainfilename = mainfilename.substring(l + 1);
```
clevengr, line 208: I do not understand the testing for `/` and `/` characters in the file name.  My quick testing (I printed out the filenames on the browser console inside of this method) indicates that what is being passed into the component (and hence into this method) is the `basename` of the file (i.e., the file name with all paths elements removed).  If this is the case, there's no point in looking for a `/` or `\`.

On the other hand, if there WAS a `/` or `\` in the file name, it's quite likely there would be MULTIPLE such characters (e.g. /usr/me/mydir/filename).  In that case, `indexOf("/")` would return the index of the FIRST such character -- and the following `substring(l+1)` would return a string which still has all the remaining `/` (or `\`) characters in it -- which wouldn't end up comparing the basename.     What's the purpose of doing it this way?

(Same comments apply to the testing of the "additional file" names, below...)

Have I somehow misunderstood what this code is doing?

SamanwaySadhu: You're correct it is a bug. I had meant to use the function `lastIndexOf()` instead of `indexOf()`. However, when I tested it didn't have erroneous behavior possibly because as you mentioned it is only taking the `basename`.

So, I wanted to know will it always be the case that only `basename` of the file taken. If yes, then I can get rid of all the checks for file separator character.  Otherwise, I can change it to `lastIndexOf()`. Also if I have to use the `path.sep` it needs to be further explored how to solve the `error TS2307: Cannot find module 'path' or its corresponding type declarations` while building the `package.xml`

johnbrvc: We just got finished discussing this very issue on the PC2 call, and I think we decided that you will only get the main file basename, and the list of additional files will only be basenames as well, however, we'd like to leave the code in there that deals with a path name if there are separators.

clevengr: What I would suggest regarding the `path.sep` issue is to file a separate GitHub issue pointing out that it's a fix that should (eventually) be applied to `new-run.component.ts`, and don't worry about it any further for now.  I think it's likely to be a rabbit-hole that will just distract people (Samanway) from moving on to other things which are more important at the moment.

clevengr (approved):
I've reviewed the updated code (with the `lastIndexOf()` fix), and run tests using both the Fat Team Client and the WTI.  Looks good, works fine.  Marking approval.

johnbrvc (approved):
Looks good. Tested and works.

PR #904, opened 2023-12-25
I131: Update Cancelling invalid changes not to discard

### LINKED ISSUE #131: Cancelling an invalid problem edit then saying "save" discards all changes.
**Describe the issue**: 

If the Admin makes changes while editing a problem, then pushes the "Cancel" button, a dialog saying "**Problem modified, save changes?  [Yes] [No] [Cancel]**"  is presented.  If the user presses "Yes", the expectation is that the changes which were entered will be saved.  This works correctly _if all the changes were valid_ (that is, if the changes result in a "valid problem state".

If however the changes leave the problem in an _invalid_ state, pressing "Yes" _discards all changes -- even valid ones_.   

For example, if the user changes the problem time limit and also changes the Output Validator to "PC2" (from something other than "PC2") but _fails to select a "PC2 Validator Option"_, the Admin returns to the main screen _having discarded all changes, including the legitimate change of the problem time limit_.  Confusingly, it also subsequently displays a popup warning "PC2 Validator is selected; you must select a Validator Mode Option" -- even though the Edit screen is gone and the user cannot select that option any more without re-opening the Edit Problem pane -- but the user has lost all other changes made (even though they said "Yes, save changes").

**To Reproduce**:

- Start a server with a contest having at least one problem.  
- Start an Admin, select a problem, press the "Edit Problem" button.
- Select the "Output Validator" tab.  If the problem already has "Use PC2 Validator" selected:
  - select the "**Do not use Validator**" radio button.
  - press **Update** (this is to make sure the problem starts with NOT having the PC2 validator, so we can _switch **to**_ the PC2 validator).
  - Select "Edit Problem" again.

- Change the "Run Timeout Limit" (on the **General** tab) to some new value (remember the new value).
- Select the "Output Validator" tab.
- Select the **Use PC2 Validator** radio button.
- Make sure that the "PC2 Validator Mode" drop-down says "None selected" (change it to "None selected" if not).

At this point you have made two changes:  setting a new Run Timeout Limit and selecting the PC2 Output Validator (with option "None" selected).  You might want to make all sorts of additional changes -- judging type, data files, input validator, etc. -- these would represent all sorts of changes a user might also make.  However, no more changes are necessary to demonstrate the bug.  Continue:

- Press the **Cancel** button. This displays a dialog "Problem modified, save changes?".
- Press the **Yes** button.

**Expected behavior**: 

The system should prompt that the current configuration is invalid (because no PC2 Option has been selected), and should _**leave the user on the Edit Problem screen**_  allowing them to fix the invalid data while not losing all the other changes they've made.

**Actual behavior**: 

- The system discards ALL changes and returns to the "Problems Pane".  The fact that ALL changes (even valid ones) were discarded even though the user said "Yes, save changes" can be verified by looking at the "Run Timeout Limit" value -- it will have its OLD value, not the new changed value.
- A delayed Message dialog saying "PC^2 Validator is selected; you must select a Validator Mode option" pops up -- but dismissing that Message dialog doesn't give the user the option to fix the problem; rather, all changes have been discarded just as if the user had selected "No, do not save changes".

**Environment**:
 - This was observed on Windows 8.1 and 10; I'm pretty sure it will manifest on ANY OS platform.
 - This was observed in PC2 Version 9.7.0-develop (that is, on the Develop branch in my GitHub fork, which is up-to-date with Develop in the master repo.   

**Additional context**: 

  Method `EditProblemPane.handleCancelButton()` calls `updateProblem()`, which in turn calls `validateProblemFromFields()`, which contains the following logic block:

``` java
        if (getUsePC2ValidatorRadioButton().isSelected()) {
            if (getPc2ValidatorOptionComboBox().getSelectedIndex() < 1) {
                showMessage("PC^2 Validator is selected; you must select a Validator Mode option (\"Output Validator\" tab)");
                return false;
            }
        }

```
Method `showMessage()` calls `SwingUtilities.invokeLater()` to display the "invalid-state" message (which explains why the message shows up later).  `validateProblemFromFields()` then returns `false`, which causes `updateProblem()` to terminate the attempt to update the problem (thus discarding all changes, including legitimate ones).  

The issue, then, is why doesn't the Edit Problem pane remain visible at that point so the user can fix the problem?

Further note: the above uses "selecting the PC2 validator without selecting a validator option" as an example of the issue -- but the problem isn't strictly related to the PC2 Validator Option.  For example, on the "i118_Integrate_VIVA" branch, the exact same issue occurs if you select an invalid Input Validator combination -- for example, selecting the "Use VIVA" button but failing to enter a VIVA pattern -- then press "Cancel" then press "Yes, save changes":  ALL changes are lost and a delayed popup Message dialog appears.  I'm pretty sure that ANY combination of "multiple changes where one of the changes leaves the problem in an invalid state" will cause this bug to manifest.

SamanwaySadhu:
`handleCancelButton()` calls `updateProblem()` but it has no way to know whether it actually updated problems or exited due to some invalid changes. It simply runs the command to close the window after that. So, the return type of `updateProblem()` was changed to `boolean` and it returns `true` only if it successfully updates problem and `false` otherwise. Now `handleCancelButton()` will close the window only if `updateProblem()` returns `true`. Similar changes were made for `addProblem()`.

Fixes #131 

Windows 10, Eclipse 2021-12 R, JDK 8u381 (1.8)

- Run the Ant script build.xml in the pc2v9 project.
- Start a PC2 server, loading a sample contest such as "SumitHello".
- Start a PC2 Admin.
- Select a problem and click **Edit**.
- Make multiple changes with at least one invalid change.
- Click on **Cancel**.
- This displays a dialog "Problem modified, save changes?". Click on **Yes**.
- A prompt showing invalid configuration shows up. Click on **OK**
- Now the Edit problem pane does not close or discard any changes.
- Try the same for **Add** problem.
- Now if all changes are valid and user accepts changes then the pane will close.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
+++ b/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
@@ -425,11 +425,11 @@ public class EditProblemPane extends JPanePlugin {
      * message "Problem Modified - Save Changes?") and responding "Yes" to the message.
      * 
      */
-    protected void addProblem() {
+    protected boolean addProblem() {
 
         if (problemNameTextField.getText().trim().length() < 1) {
             showMessage("Enter a problem name (\"General\" tab)");
-            return;
+            return false;
         }
 
         if (!validateProblemFields()) {
@@ -437,7 +437,7 @@ public class EditProblemPane extends JPanePlugin {
             if (debug22EditProblem) {
                 System.err.println("DEBUG: validateProblemFields() returned false");
             }
-            return;
+            return false;
         }
 
         //warn if the problem has input data files but no Input Validator, unless the user has previously disabled this warning
@@ -464,7 +464,7 @@ public class EditProblemPane extends JPanePlugin {
                 showMissingInputValidatorWarningOnAddProblem = false;
             }
             if (!(response == JOptionPane.YES_OPTION)) {
-                return;
+                return false;
             }
         }
 
@@ -485,7 +485,7 @@ public class EditProblemPane extends JPanePlugin {
                 showMissingInputValidatorProgramNameOnAddProblem = false;
             }
             if (!(response == JOptionPane.YES_OPTION)) {
-                return;
+                return false;
             }
         }
 
@@ -532,11 +532,11 @@ public class EditProblemPane extends JPanePlugin {
             }
         } catch (InvalidFieldValue e) {
             showMessage(e.getMessage());
-            return;
+            return false;
         } catch (Exception e) {
             showMessage("Exeception while adding Problem; see log.");
             getLog().throwing("EditProblemPane", "addProblem()", e);
-            return;
+            return false;
         }
 
         if (!newProblem.getElementId().equals(newProblemDataFiles.getProblemId())) {
@@ -570,6 +570,7 @@ public class EditProblemPane extends JPanePlugin {
         if (getParentFrame() != null) {
             getParentFrame().setVisible(false);
         }
+        return true;
     }
 
     private int getIntegerValue(String s) {
@@ -2139,7 +2140,9 @@ public class EditProblemPane extends JPanePlugin {
 
             if (result == JOptionPane.YES_OPTION) {
                 if (getAddButton().isEnabled()) {
-                    addProblem();
+                    if (!addProblem()) {
+                        return;
+                    }
                 } else {
                     if (!updateProblem()) {
                         return;
--- a/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
+++ b/src/edu/csus/ecs/pc2/ui/EditProblemPane.java
@@ -1654,13 +1654,13 @@ public class EditProblemPane extends JPanePlugin {
      * 
      */
 
-    protected void updateProblem() {
+    protected boolean updateProblem() {
 
         // showStackTrace();
 
         if (!validateProblemFields()) {
             // problem defined by the GUI fields is invalid, just return ( error message was issued by validateProblemFields() )
-            return;
+            return false;
         }
 
         //warn if the problem has input data files but no Input Validator, unless the user has previously disabled this warning
@@ -1687,7 +1687,7 @@ public class EditProblemPane extends JPanePlugin {
                 showMissingInputValidatorWarningOnUpdateProblem = false;
             }
             if (!(response == JOptionPane.YES_OPTION)) {
-                return;
+                return false;
             }
         }
 
@@ -1709,7 +1709,7 @@ public class EditProblemPane extends JPanePlugin {
                 showMissingInputValidatorProgramNameOnUpdateProblem = false;
             }
             if (!(response == JOptionPane.YES_OPTION)) {
-                return;
+                return false;
             }
         }
         // all the GUI fields are valid; create a new Problem from them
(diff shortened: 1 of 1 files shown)


src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -425,19 +425,19 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
      * message "Problem Modified - Save Changes?") and responding "Yes" to the message.
      *
```
johnbrvc, line 426: I would add **javadoc** to the method comment to indicate what this method returns now.  eg.
`@return true if problem was added, false if not`

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -425,19 +425,19 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
...
+    protected boolean addProblem() {
 
         if (problemNameTextField.getText().trim().length() < 1) {
             showMessage("Enter a problem name (\"General\" tab)");
-            return;
+            return false;
         }
 
         if (!validateProblemFields()) {
             // new problem is invalid, just return; message issued by validateProblemFields
```
johnbrvc, line 436: Change comment to indicate `false `is returned, since it no longer just "returns".

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -1654,13 +1655,13 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
      * 
      */
```
johnbrvc, line 1656: I would add **javadoc** to the method comment to indicate what this method returns now.  eg.
`@return true if problem was successfully updated, false if not`

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
```diff
@@ -1654,13 +1655,13 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
...
 
-    protected void updateProblem() {
+    protected boolean updateProblem() {
 
         // showStackTrace();
 
         if (!validateProblemFields()) {
             // problem defined by the GUI fields is invalid, just return ( error message was issued by validateProblemFields() )
```
johnbrvc, line 1663: Change comment to indicate `false `is returned since it no longer just "returns".

johnbrvc (approved):
Built and tested.  Looks good!
The suggestions I made are minor, but should probably be addressed.  
Since it's only comments, I will approve the changes now, so when the 2nd approval comes in it can be merged without having to review it again.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: Agree; comments should be updated to reflect new functionality.

src/edu/csus/ecs/pc2/ui/EditProblemPane.java
clevengr: Agree.

johnbrvc (approved):

clevengr (approved):
I've reviewed the code, and also built and tested it.  Everything seems to work as expected/intended; I'm adding an approval.

I DID discover that this bug FIX actually EXPOSES several bugs which were previously hidden.  I will write those up as separate new issues.  (None of the new issues are critical for the upcoming Regional contest...)

PR #910, opened 2024-01-08
i_395 scoring properties settings now on main page

### LINKED ISSUE #395: Update for Settings -> Edit Scoring Properties does not save Scoring Properties
**Describe the issue**: 

When saving Settings -> Scoring Properties Update does not update the values.

Every other Update in pc2 saves the values, except Scoring Properties, that inconsistency
can cause confusion.

One is required to also Update on the Settings tab.

**To Reproduce**: 

On admin, select Settings, select Scoring Properties
change any property, the Update button will enable.
Click Update.

the change is not saved until the Settings Tab Update button is selected.

**Expected behavior**: 

Scoring Properties should be saved when Update selected.

**Actual behavior**: 

The Scoring Properties change is not saved until the Settings Tab Update button is selected.

**Environment**: 

All.

**Log Info**:  

**Screenshots**:  

**Additional context**:

kkarakas:
Judging Settings Pane in Settings for Admin has been updated so that Scoring Properties are displayed without a new frame and saves without additional button.

Fixes #395 

On admin, select settings, see that there is a Scoring Properties Pane, try to update the value. 

Screenshot 2024-01-08 032649
Screenshot 2024-01-08 032951

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
@@ -35,6 +35,8 @@ public class ScoringPropertiesPane extends JPanel {
     
     private JButton cancelButton;
     
+    private int lengthOfColumns;
+    
     public ScoringPropertiesPane(JButton updateButtona,JButton cancelButtona) {
         super();
         updateButton = updateButtona;
@@ -45,14 +47,16 @@ public class ScoringPropertiesPane extends JPanel {
     
     private void initialize() {
         this.setLayout(new BorderLayout());
-        this.setPreferredSize(new Dimension(275, 185));
+        this.setPreferredSize(new Dimension(350, 200));
         this.add(getPropertyListBox(), BorderLayout.CENTER);
+        this.lengthOfColumns = 167;
+        
 
     }
     
     protected void updateProperties(Properties properties) {
         propertyUpdater.updateProperties(properties);
-        propertyListBox.removeAllRows(); //WHYY??
+        propertyListBox.removeAllRows();
 
     }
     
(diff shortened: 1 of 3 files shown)


troy2914 (approved):
This looks reasonable,  and tested fine.

johnbrvc (changes requested):
Code looks good and works as described, however, the GUI does not look right on my system (probably a result of the flow layout and my default fonts/resolution).  See the attached image.  Resizing the window obviously does not change anything.

image

In addition, I don't see any reason not to make the property sheet be a little wider to start with.  The reason being, in practice, the Value's for the HTML directories tend to be longer than the defaults, so it would be nice to see a little more context and not have the scroll-bars pop up almost immediately when you start typing in one of those fields.  Maybe make the **Value** column as wide as the **Property** column?

clevengr:
I agree that the layout needs to be fixed so it looks reasonable on all (or at least most) platforms.  I'm pretty sure it's a matter of something like putting the components into their own sub-JPanels, and perhaps changing the LayoutManager for the pane.  Unfortunately, it's not something I have time to work on myself right now :(

src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
```diff
@@ -115,7 +121,8 @@ protected void refreshProperties() {
             propertyListBox.addRow(objects);
         }
         
-        propertyListBox.autoSizeAllColumns();
+        propertyListBox.setColumnSize(0,this.lengthOfColumns);
```
johnbrvc, line 124: Same as above.  Why are you using "this." ?

src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
```diff
@@ -69,8 +73,10 @@ private MCLB getPropertyListBox() {
             HeapSorter sorter = new HeapSorter();
             propertyListBox.setColumnSorter(1, sorter, 1);
 
-            propertyListBox.autoSizeAllColumns();
-
+            propertyListBox.setColumnSize(0,this.lengthOfColumns);
```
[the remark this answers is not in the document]
kkarakas: I think I pickup that habit in Python. Java implicitly adds preprends that i think so as you suggest, there is no need for that here.

src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
kkarakas: I think I pickup that habit in Python. Java implicitly adds preprends that i think so as you suggest, there is no need for that here.

johnbrvc (approved):
Looks good.

PR #907, opened 2023-12-30
I209: Display team stderr to judges

SamanwaySadhu:
### Description of what the PR does
- Updated `Executable` class to now save Team stderr for each test case in a file and return the list of files
- Updated `SelectJudgmentPaneNew` to receive list of team stderr files from executable and forward it to `TestResultsPane`
- Updated `TestResultsFrame` to forward list of team stderr files to `TestResultsPane`
- Updated `TestResultsPane` to add a new column for team stderr and show the stderr upon button click
- Updated `TestCaseResultsTableModel` to handle new column for team stderr
- Updated `TestResultsRowData` to handle new column for team stder

### Issue which the PR addresses
Fixes #209 

### Environment in which the PR was developed (OS, IDE, Java version, etc.)
Windows 10, Eclipse 2021-12 R, JDK 8u381 (1.8), C++ 13.2.0

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)
- Run the Ant script build.xml in the pc2v9 project.
- Start a PC2 server, loading a sample contest such as "SumitHello".
- Start a PC2 Admin.
- In `Problems` pane make sure one of the problems have judging type set to Manual, if not edit and set to Manual.
- In `Auto Judge` pane make sure one of the judges have Problems `(none selected)`, if not edit and set.
- Start the contest.
- Start a PC2 team.
- Make a submission for a problem which had manual judging type set.
- It is usually hard to get to receive execution time STDERR. I set C++ and simulated using `cerr`. Need to be tested for other languages.
- Starte a PC2 judge. Make sure to use the one which no problems set for auto-judge.
- Select the submission from `New Runs` pane and request run.
- In `Select Judgement for Run {:id}` window click on Execute Run.
- In `Test Results for Run {:id}` you should see a new column name `Team StdErr`. Click on View to view the StdErr.
image

### Additional Context
- CI: In All Runs pane if we try to Rejudge, then View Outputs and Data and then try to view any Team Output/Stderr, etc. it throws NullPointerException error. This is because in TestResultsPane lines 2199 and 2209 it checks for the condition `currentTeamOutputFileNames != null || currentTeamOutputFileNames.length >= row` and because the condition has OR it checks for the length even if the variable is null.
Resolution: Changed the condition to AND i.e. `&&`.
[...]

PR #918, opened 2024-01-28
I39: Add Execute All in judge button to bypass stop on first failure

### LINKED ISSUE #39: On Judge UI add feature to execute all test cases if stop-on-first-failed-test-case is true
**Feature Description**: 

On Judge UI add a feature to execute all tests cases if stop-on-first-failed-test-case is true.

The only way to execute all test cases would be to change
the Problem def and that potentially could cause more problems
and more load on auto judges if they start judging the same 
problem.

Add a button in the Test Results frame at the bottom named
Execute all with a tool tip of Execute all test cases.

**Additional context**: 

This was requested by a judge at the local CSUS contest.

SamanwaySadhu:
Add `Execute All` button on `TestResultsPane`. It can bypass stop on first failure condition and run the code on all testcases if a judge desires.

Fixes #39 

Windows 10, Eclipse 2021-12 R, JDK 8u381 (1.8), C++ 13.2.0

- Run the Ant script build.xml in the pc2v9 project.
- Start a PC2 server, loading a sample contest such as "mini".
- Start a PC2 Admin.
- In `Problems` pane make sure one of the problems have judging type set to Manual, if not edit and set to Manual. Also check the box for stop-on-first-failure.
- In `Auto Judge` pane make sure one of the judges have Problems `(none selected)`, if not edit and set.
- Start the contest.
- Start a PC2 team.
- Make a submission for a problem which had manual judging type set. Make sure it fails on the first testcase.
- Starte a PC2 judge. Make sure to use the one which no problems set for auto-judge.
- Select the submission from `New Runs` pane and request run.
- In `Select Judgement for Run {:id}` window click on Execute Run.
- It will fail on first test case and the rest are not executed.
- Now click on `Execute All` button in Test Results Pane. It should now execute all test cases despite failures.

image

image

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
+++ b/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.BorderLayout;
--- a/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
+++ b/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
@@ -628,7 +628,7 @@ public class SelectJudgementPaneNew extends JPanePlugin {
         getAcceptValidatorJudgementButton().setEnabled(b);
         // getViewOutputsButton().setEnabled(b);
         getAcceptChosenSelectionButton().setEnabled(b && getJudgementComboBox().getSelectedIndex() != -1);
-        getTestResultsFrame().setExecuteAllButton(b);
+        getTestResultsFrame().enableExecuteAllButton(b);
     }
 
     /**
@@ -1529,7 +1529,7 @@ public class SelectJudgementPaneNew extends JPanePlugin {
         if (testResultsFrame == null) {
             testResultsFrame = new TestResultsFrame();
             // Add action listener whether Test results pane button is clicked
-            testResultsFrame.addActionListenerToExecuteAll(new ActionListener() {
+            testResultsFrame.addActionListenerToExecuteAllButton(new ActionListener() {
                 @Override
                 public void actionPerformed(ActionEvent e) {
                     new Thread(new Runnable() {
--- a/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
+++ b/src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
@@ -628,6 +628,7 @@ public class SelectJudgementPaneNew extends JPanePlugin {
         getAcceptValidatorJudgementButton().setEnabled(b);
         // getViewOutputsButton().setEnabled(b);
         getAcceptChosenSelectionButton().setEnabled(b && getJudgementComboBox().getSelectedIndex() != -1);
+        getTestResultsFrame().setExecuteAllButton(b);
     }
 
     /**
@@ -807,7 +808,7 @@ public class SelectJudgementPaneNew extends JPanePlugin {
                 public void actionPerformed(java.awt.event.ActionEvent e) {
                     new Thread(new Runnable() {
                         public void run() {
-                            executeRun();
+                            executeRun(false);
                         }
                     }).start();
                 }
@@ -859,8 +860,11 @@ public class SelectJudgementPaneNew extends JPanePlugin {
         showMessage("Would have extracted run");
         // TODO code extract run
     }
-
-    protected void executeRun() {
+    
+    /**
+     * Takes a boolean condition whether to override stop on first failure condition
+     */
+    protected void executeRun(boolean overrideStopOnFirstFailedTestCase) {
 
         executeTimeMS = 0;
         System.gc();
@@ -876,6 +880,11 @@ public class SelectJudgementPaneNew extends JPanePlugin {
         
        // getManualRunResultsPanel().clear();
         setEnabledButtonStatus(false);
+
+        // Set override to execute all testcases
+        if (overrideStopOnFirstFailedTestCase) {
+            executable.setOverrideStopOnFirstFailedTestCase(true);
+        }
         executable.execute();
         
         // Dump execution results files to log
@@ -1519,6 +1528,17 @@ public class SelectJudgementPaneNew extends JPanePlugin {
     private TestResultsFrame getTestResultsFrame() {
         if (testResultsFrame == null) {
             testResultsFrame = new TestResultsFrame();
+            // Add action listener whether Test results pane button is clicked
+            testResultsFrame.addActionListenerToExecuteAll(new ActionListener() {
+                @Override
+                public void actionPerformed(ActionEvent e) {
+                    new Thread(new Runnable() {
+                        public void run() {
+                            executeRun(true);
+                        }
+                    }).start();
+                }
+            });
             testResultsFrame.setContestAndController(getContest(), getController());
             FrameUtilities.centerFrame(testResultsFrame);
         }
--- a/src/edu/csus/ecs/pc2/ui/TestCaseResultsTableModel.java
+++ b/src/edu/csus/ecs/pc2/ui/TestCaseResultsTableModel.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.util.Vector;
--- a/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
+++ b/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import edu.csus.ecs.pc2.core.IInternalController;
--- a/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
+++ b/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
@@ -102,5 +102,18 @@ public class TestResultsFrame extends javax.swing.JFrame implements UIPlugin {
         getMultiTestSetOutputViewerPane().setValidatorOutputFileNames(filenames);
     }
 
+    /**
+     * Enables/Disables Execute All button
+    */
+    public void setExecuteAllButton(boolean b) {
+        getMultiTestSetOutputViewerPane().setExecuteAllButton(b);
+    }
+
+    /**
+     * Adds action listeners to Execute All Button
+    */
+    public void addActionListenerToExecuteAll(java.awt.event.ActionListener e) {
+        getMultiTestSetOutputViewerPane().addActionListenerToExecuteAll(e);
+    }
 
 } // @jve:decl-index=0:visual-constraint="10,10"
--- a/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
+++ b/src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
@@ -105,15 +105,15 @@ public class TestResultsFrame extends javax.swing.JFrame implements UIPlugin {
     /**
      * Enables/Disables Execute All button
     */
-    public void setExecuteAllButton(boolean b) {
-        getMultiTestSetOutputViewerPane().setExecuteAllButton(b);
+    public void enableExecuteAllButton(boolean b) {
+        getMultiTestSetOutputViewerPane().enableExecuteAllButton(b);
     }
 
     /**
      * Adds action listeners to Execute All Button
     */
-    public void addActionListenerToExecuteAll(java.awt.event.ActionListener e) {
-        getMultiTestSetOutputViewerPane().addActionListenerToExecuteAll(e);
+    public void addActionListenerToExecuteAllButton(java.awt.event.ActionListener e) {
+        getMultiTestSetOutputViewerPane().addActionListenerToExecuteAllButton(e);
     }
 
 } // @jve:decl-index=0:visual-constraint="10,10"
--- a/src/edu/csus/ecs/pc2/ui/TestResultsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/TestResultsPane.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.BorderLayout;
--- a/src/edu/csus/ecs/pc2/ui/TestResultsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/TestResultsPane.java
@@ -1013,13 +1013,6 @@ public class TestResultsPane extends JPanePlugin implements TableModelListener {
             resultsPaneButtonPanel.add(getCompareSelectedButton());
 
             //add space
-            Component horizontalStrut_4 = Box.createHorizontalStrut(20);
-            resultsPaneButtonPanel.add(horizontalStrut_4);
-            
-            // add a control button to invoke comparison of the team and judge output files for selected row(s)
-            resultsPaneButtonPanel.add(getExecuteAllButton());
-
-            //add space
             Component horizontalGlue_3 = Box.createHorizontalGlue();
             horizontalGlue_3.setPreferredSize(new Dimension(20, 20));
             resultsPaneButtonPanel.add(horizontalGlue_3);
@@ -1028,6 +1021,63 @@ public class TestResultsPane extends JPanePlugin implements TableModelListener {
             resultsPaneButtonPanel.add(getResultsPaneCloseButton());
         }
         return resultsPaneButtonPanel;
+    }
+
+    /**
+     * refreshes the resultspanebuttonpanel
+     * includes/removes execute all button based on stop-on-first-failure
+     */
+    private void updateResultsPaneButtonPanel() {
+        
+        if (resultsPaneButtonPanel == null) {
+            resultsPaneButtonPanel = new JPanel();
+        }
+
+        resultsPaneButtonPanel.removeAll();
+            
+        //add a checkbox to allow filtering between all test cases and only failed test cases
+        resultsPaneButtonPanel.add(getShowFailuresOnlyCheckbox());
+
+        //add space
+        Component horizontalStrut_1 = Box.createHorizontalStrut(20);
+        resultsPaneButtonPanel.add(horizontalStrut_1);
+
+        //add a button to select all test cases in the grid
+        resultsPaneButtonPanel.add(getSelectAllButton());
+            
+        //add space
+        Component horizontalStrut_3 = Box.createHorizontalStrut(20);
+        resultsPaneButtonPanel.add(horizontalStrut_3);
+            
+        //add a button to unselect all the test cases in the grid
+        resultsPaneButtonPanel.add(getUnselectAllButton());
+
+        //add space
+        Component horizontalStrut_2 = Box.createHorizontalStrut(20);
+        resultsPaneButtonPanel.add(horizontalStrut_2);
+            
+        // add a control button to invoke comparison of the team and judge output files for selected row(s)
+        resultsPaneButtonPanel.add(getCompareSelectedButton());
+
+        if (currentProblem != null && currentProblem.isStopOnFirstFailedTestCase()) {
+            //add space
+            Component horizontalStrut_4 = Box.createHorizontalStrut(20);
+            resultsPaneButtonPanel.add(horizontalStrut_4);
+            
+            // add a control button to execute all test cases
+            resultsPaneButtonPanel.add(getExecuteAllButton());
+        }
+
+        //add space
+        Component horizontalGlue_3 = Box.createHorizontalGlue();
+        horizontalGlue_3.setPreferredSize(new Dimension(20, 20));
+        resultsPaneButtonPanel.add(horizontalGlue_3);
+
+        // add a control button to dismiss the frame
+        resultsPaneButtonPanel.add(getResultsPaneCloseButton());
+
+        resultsPaneButtonPanel.revalidate();
+        resultsPaneButtonPanel.repaint();
     }
     
     /**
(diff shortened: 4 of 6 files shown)


src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
johnbrvc, line 1: If the only change is the copyright notice, why is this file in the PR?

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
johnbrvc, line 1: If the only change is copyright, why is this in the PR?

src/edu/csus/ecs/pc2/ui/TestCaseResultsTableModel.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
johnbrvc, line 1: If the only change is the copyright date, why is this in the PR?

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
johnbrvc, line 1: If the only change is the copyright date, why is this in the PR?

src/edu/csus/ecs/pc2/ui/TestResultsRowData.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
johnbrvc, line 1: If the only change is the copyright date, why is this in the PR?

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
```diff
@@ -1003,6 +1012,13 @@ private JPanel getResultsPaneButtonPanel() {
             // add a control button to invoke comparison of the team and judge output files for selected row(s)
             resultsPaneButtonPanel.add(getCompareSelectedButton());
 
+            //add space
+            Component horizontalStrut_4 = Box.createHorizontalStrut(20);
+            resultsPaneButtonPanel.add(horizontalStrut_4);
+            
+            // add a control button to invoke comparison of the team and judge output files for selected row(s)
```
johnbrvc, line 1019: Comment not correct.  refers to comparison button, probably copied from above.  Should be "execute all" button?

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
```diff
@@ -1003,6 +1012,13 @@ private JPanel getResultsPaneButtonPanel() {
...
+            resultsPaneButtonPanel.add(getExecuteAllButton());
```
johnbrvc, line 1020: It might be a good idea to only add this button if the problem has "stop at first error" set.  If "stop at first error" is not set, then this button is meaningless (and confusing).

johnbrvc (changes requested):
See individual comments on each file for notes.

src/edu/csus/ecs/pc2/ui/TestCaseResultsTableModel.java
SamanwaySadhu: It should've been changed in the PR for I209, but it was already changed so changed here.

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
SamanwaySadhu: It should've been changed in the PR for I209, but it was already changed so changed here.

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
SamanwaySadhu: It should've been changed in the PR for I209, but it was already changed so changed here.

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
SamanwaySadhu: It should've been changed in the PR for I209, but it was already changed so changed here.

src/edu/csus/ecs/pc2/ui/TestResultsRowData.java
SamanwaySadhu: It should've been changed in the PR for I209, but it was already changed so changed here.

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
SamanwaySadhu: Updated

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
SamanwaySadhu: Updated

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
johnbrvc: The problem is, there is no traceability if we just change the date - we have no idea why it changed.  In general, if it's just the copyright date, we won't just modify that in a different PR as it's confusing.  However, now that it's done, and we have this log attached to it, we'll know why it was done.  In the comment for the commit "Update copyright year on updated files", it may have helped to say something about the fact it was forgotten in #907.  I won't comment on the other ones since they're all the same.

johnbrvc:
One general question I have before I actually start testing the PR  is: I noticed you duplicated all the "`executeRun`" code in the `TestResultsPane `from `SelectJudgementPaneNew.executeRun()`.  Is there any reason you couldn't leverage the code already in `SelectJudgementPaneNew `directly instead of duplicating it?  The problem with duplicating it is if a change is required in `executeRun`, we have to make that change in 2 places now.  I believe that `RestResultsFrame`/`TestResultsPane` is modeless, so maybe some sort of interface could be used to call `executeRun `from `SelectJudgementPaneNew`?   I haven't thought about it much, what are your thoughts on this?

clevengr:
The "Steps for Testing the PR" shows a sample GUI screenshot, which is apparently supposed to be related somehow to testing the PR.  However, as it shows in the GUI header, there is only ONE total test case for the problem.  Using a problem with only one test case doesn't really help in explaining how to test the ability to execute multiple test cases after the first one fails...

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
```diff
@@ -1519,6 +1528,17 @@ protected void viewOutputsAndData() {
...
+            testResultsFrame.addActionListenerToExecuteAll(new ActionListener() {
+                @Override
+                public void actionPerformed(ActionEvent e) {
+                    new Thread(new Runnable() {
+                        public void run() {
+                            executeRun(true);
+                        }
+                    }).start();
+                }
+            });
```
clevengr, line 1541: This block of code (adding an `actionListener()` to the `ExecuteAll` button) is a bit awkward; normally, code like this belongs in the singleton accessor/constructor (in this case, in `TestResultsPane.getExecuteAllButton()`, right after the `executeAllButton = new JButton()` statement).  As presently coded, this block makes calls through a series of methods in different classes (`TestResultsFrame.addActionListenerToExecuteAll()` --> `TestResultsPane.addActionListenerToExecuteAll()` --> `TestResultsPane.getExecuteAllButton()`).  If changes are made later to methods in that sequence, it can result in an error in the calling class (`SelectJudgementPaneNew`).   In other words, there is a lot of tight coupling between these classes, which in general is not good.

I do recognize that moving this block of code into the singleton accessor/constructor would raise the problem of how the ActionListener code is going to be able to execute the statement `executeRun(true);` since the `execute()` method is in the `SelectJudgementPaneNew` class, not in the `TestResultsPane` class where the `ExecuteAll` button is defined.  I suspect that moving the code to the `ExecuteAll` constructor would require adding code to allow it to reference back to the `SelectJudgementPaneNew.execute()` method, so I'm not necessarily requesting a change.  It just seems awkward to have this long sequence of `addActionListenerToExecuteAll()` method calls.

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
```diff
@@ -102,5 +102,18 @@ public void setValidatorOutputFileNames(String[] filenames) {
         getMultiTestSetOutputViewerPane().setValidatorOutputFileNames(filenames);
     }
 
+    /**
+     * Enables/Disables Execute All button
+    */
+    public void setExecuteAllButton(boolean b) {
```
clevengr, line 108: As the COMMENT above says, this method is enabling/disabling the button.  However, the _name_ of the method doesn't say anything about "enabling"; it talks about "setting".  It's not completely clear what it means to "set" a button.  It could mean "position it in a specific location"; it could mean "set the text on the button"; it could mean other things.  

I think the method should be named (something like) `setExecuteAllButtonEnabled(boolean)`, or perhaps `enableExecuteAllButton(boolean)`.

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
```diff
@@ -102,5 +102,18 @@ public void setValidatorOutputFileNames(String[] filenames) {
...
+        getMultiTestSetOutputViewerPane().setExecuteAllButton(b);
```
clevengr, line 109: Similar to the above comment, I think method `setExecuteAllButton()` (in class `TestResultsPane`) should be renamed to reflect the fact that what it is doing is managing the _enabled state_ of the button.

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
```diff
@@ -102,5 +102,18 @@ public void setValidatorOutputFileNames(String[] filenames) {
...
+    }
+
+    /**
+     * Adds action listeners to Execute All Button
+    */
+    public void addActionListenerToExecuteAll(java.awt.event.ActionListener e) {
```
clevengr, line 115: As the above comment says, this method is adding something (a listener) to a BUTTON.  However, the method name doesn't make that clear;  "ExecuteAll" could refer to some other type of object.  I would recommend changing the name of this method (and the one it calls in the following line) to `addActionListenerToExecuteAllButton()`.

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
```diff
@@ -1086,6 +1145,22 @@ public void actionPerformed(ActionEvent e) {
...
+
+    private JButton getExecuteAllButton() {
+
+        if (executeAllButton == null) {
+            executeAllButton = new JButton("Execute All");
+        }
+        return executeAllButton;
+    }
+
+    public void addActionListenerToExecuteAll(java.awt.event.ActionListener e) {
```
clevengr, line 1157: As noted elsewhere, I would recommend changing the name of this method to `addActionListenerToExecuteAllButton()`, for clarity.

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
```diff
@@ -1086,6 +1145,22 @@ public void actionPerformed(ActionEvent e) {
...
+        getExecuteAllButton().addActionListener(e);
+    }
+
+    public void setExecuteAllButton(boolean b) {
```
clevengr, line 1161: As noted elsewhere, I would recommend changing the name of this method to `setExecuteAllButtonEnabled()`, (or perhaps `setExecuteAllButtonEnabledState()` or `enableExecuteAllButton()`) for clarity.

clevengr (changes requested):
I have built the code and done a run-time test; it appears to work as intended (and is a nice improvement).

I have made a number of comments in the code about things I'd "recommend" being fixed/updated.  Most of them are minor and I could live with them as-is. However, there is one change which I think needs to be made prior to approval:  see the comment in class `TestResultsPane` regarding the inappropriate duplication of code in method `updateResultsPaneButtonPanel()`.  That method should not be recreating the entire button panel from scratch (for the reasons I listed in the code-review comment). Rather, it should only be "updating" the button panel by adding/removing the "Execute All" button, as appropriate.

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
SamanwaySadhu: Updated

src/edu/csus/ecs/pc2/ui/TestResultsFrame.java
SamanwaySadhu: Updated

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
SamanwaySadhu: I kind of figured out a way to fix this however one small issue that arises is that the Execute All button now comes up in the end after the Close button. So far, I'm unable to figure out how to change the ordering of the buttons. I'm still looking into it.

image

SamanwaySadhu:
> The "Steps for Testing the PR" shows a sample GUI screenshot, which is apparently supposed to be related somehow to testing the PR. However, as it shows in the GUI header, there is only ONE total test case for the problem. Using a problem with only one test case doesn't really help in explaining how to test the ability to execute multiple test cases after the first one fails...

Updated the screenshots

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
```diff
@@ -1013,6 +1015,63 @@ private JPanel getResultsPaneButtonPanel() {
         }
         return resultsPaneButtonPanel;
     }
+
+    /**
+     * refreshes the resultspanebuttonpanel
+     * includes/removes execute all button based on stop-on-first-failure
+     */
+    private void updateResultsPaneButtonPanel() {
```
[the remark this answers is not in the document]
SamanwaySadhu: I kind of figured out a way to fix this however one small issue that arises is that the Execute All button now comes up in the end after the Close button. So far, I'm unable to figure out how to change the ordering of the buttons. I'm still looking into it. Basically, the issue is that there is a JPanel in FlowLayout with a bunch of components, how to now add some new components in between some existing components.

image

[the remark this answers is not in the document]
clevengr: One solution, admittedly not necessarily ideal, would be to create two sub-panels in the "Button Panel":  one to hold the "permanent" buttons (the checkbox, plus the `Select All`, `Unselect All`, `Compare Selected`, and `Execute All` buttons), and a second panel to hold (just) the `Close` button.  Then you could add/remove the `Execute All` button from the "left sub-panel" and the "right sub-panel" containing (just) the `Close` button would always appear to the right of whatever was in the "left sub-panel".  This however would certainly add some additional (perhaps undesirable) complexity to the GUI.  Another solution might be to use a `GridBagLayout` for the button panel -- but if you've never fooled with those, fair warning: they are pretty complicated (but, very powerful for controlling layouts).

A better solution of course would be if there's a way to manage the insertion point of a component in a JPanel with a flow layout -- but I'm not sure that's possible without rebuilding the entire panel (which as I pointed out isn't really desirable -- at least, there shouldn't be two different places where it's done...)

If none of the above helps, my own feeling is that it's better to have proper code organization, at the expense of having the `Execute All` button show up on the far left (others may feel differently...)  Overall, I would investigate the "two sub-panels" idea first; it might be relatively clean and easy to do.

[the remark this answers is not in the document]
SamanwaySadhu: Updated with the "two sub-panels" approach

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
SamanwaySadhu: One workaround for it can be that `TestResultsPane` object should have an object of `SelectJudgementPaneNew` as a member variable. When an object of `TestResultsPane` is created inside `SelectJudgementPaneNew` we can set the `TestResultsPane` 's member with `this` and use the `executeRun` function from there.

clevengr: Yes, this is one approach to solve the problem -- instead of having the parent (composing) class (`SelectJudgementPaneNew`) have references downward to the composed classes (like `TestResultsFrame` and `TestResultsPane`), support the ability of the composed class (`TestResultsPane`) to have a reference upward to its composing class.  This would be slightly cleaner, although it still represents tight coupling between the classes.

In any case, if you take that approach I don't think method `SelectJudgementPaneNew.executeRun()` should be _public_; it should be _protected_ -- which will allow classes in the same package to access its methods while prohibiting general public access.  This will work since both `SelectJudgementPaneNew` and `TestResultsPane` are in the same package (`edu.csus.ecs.pc2.ui`).

src/edu/csus/ecs/pc2/ui/SelectJudgementPaneNew.java
```diff
@@ -1519,6 +1528,17 @@ protected void viewOutputsAndData() {
...
+            testResultsFrame.addActionListenerToExecuteAllButton(new ActionListener() {
+                @Override
+                public void actionPerformed(ActionEvent e) {
+                    new Thread(new Runnable() {
+                        public void run() {
+                            executeRun(true);
+                        }
+                    }).start();
+                }
+            });
```
clevengr, line 1541: I still do not understand why the code to add the `actionListener()` to the `Execute All` button is here, in the `SelectJudgementPaneNew` class.  The `Execute All` button is created as an object when  method `TestResultsPane.getExecuteAllButton()` is invoked.  That method is invoked from only three places:  methods `addActionListenerToExecuteAllButton()`, `enableExecuteAllButton()`, and `updateResultsPaneButtonPaneLeft()`.

The FIRST time one of those methods calls `getExecuteAllButton()`, the button will be created.  The button does not go away (cease to exist) just because method `updateResultsPaneButtonPanelLeft()` chooses to add it to, or remove it from, the "left panel" -- it remains as a valid object in memory, with its current state intact.  If method `updateResultsPaneButtonPanelLeft()`  removes it from the panel, and then subsequently adds it back, it is _the exact same button (object)_ which is being "added back".  Therefore, there's no reason to continually "update" the `actionListener` attached to the button.

The singleton accessor/constructor for the button (method `getExecuteAllButton()`) should add the `actionListener` to the button once, when the JButton is constructed.  This avoids all the unnecessary code which is constantly updating the button's `actionListener`.    (It does of course require implementing the facility for the `TestResultsPane` to have an upward access (reference) to the `SelectJudgementPaneNew` which contains the `executeRun()` method which the `actionListener` needs to call.  But that seems a small price to pay for eliminating all the "action listener setting", and for being able to keep the entire construction of the JButton in a single place.)

clevengr (commented):
I have reviewed the updated code and also done a run-time test.  The updates fixed the main issue I had requested a change on, and the new code works fine.

I made one comment about the (IMHO) unnecessary updating of the `actionListener` for the `Execute All` button; I think it makes more sense to add that `actionListener` when the button is first constructed (in `TestResultsPane.getExecuteAllButton()`).  See my comments in the "conversation" for details.

I'm prepared to approve this PR if you will either fix the issue with the `actionListener`, or provide a description of why the changes required to fix it (that is, the code to give the `TestResultsPane` access to the parent `SelectJudgementPaneNew` object so the `actionListener` can be added in the singleton constructor/accessor method) are more complicated than what would be gained by removing the complicated `update action listener` methods.

clevengr (approved):
On further consideration, I am changing my mind regarding my previous review comments.  While I don't like the fact that the `actionListener` for the `Execute All` button is being set from a composing class (specifically, `SelectJudgementPaneNew`) rather than being set in the singleton accessor/constructor for the button, there are other considerations. 

Specifically, changing it so the `ActionListener` is set in the constructor method would require adding the same type of code to `TestResultsPane` that already exists in `SelectJudgementPaneNew` and `TestResultsPane` -- code to access UP the hierarchy of the composition (instead of DOWN the hierarchy as it presently does).  So making the change is probably a net-zero gain.

In addition, this hierarchy is a classic example of a _composition_.  In general, the design rules for a composition include that the parent controls the lifetime of its children.  This can be extended to say that the parent "knows about" its children, but that the children in general do not know about their parents.  While there are valid reasons for making exceptions to this rule, there doesn't seem to be one here; and as coded, the parents (`SelectJudgementPaneNew` and `TestResultsFrame`) do know about the "child" (`TestResultsPane` and its subcomponents).  So I think the current structure is fine (I wish I could come up with a better one, but I can't...)

So I'm adding an Approval to this PR.

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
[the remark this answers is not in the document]
johnbrvc: I'm a little late to the party, but I have been doing some testing on this branch.
Instead of a left/right panel, why not just always add the execute all button, but just set its _visiblity_:
`getExecuteAllButton().setVisible(isExecuteAllButtonVisible) `?  Basically the button will always "be there", just not visible for the cases we don't want to see it.  I suppose you could do the same for the spacing strut too.  I quickly tried this and it seems to work.  Basically, as a test, I just added:
            
```
            //add space
            getResultsPaneButtonPanelLeft().add(getExecuteAllHorizontalStrut());
            
            // add a control button to execute all test cases
            getResultsPaneButtonPanelLeft().add(getExecuteAllButton());
            
            getExecuteAllButton().setVisible(false);

```
to your "`getResultsPaneButtonPanelLeft()`" method.  And changed `updateResultsPaneButtonPanelLeft()` to:

```
    private void updateResultsPaneButtonPanelLeft() {

        if (currentProblem != null && currentProblem.isStopOnFirstFailedTestCase()) {
            isExecuteAllButtonVisible = true;
        }
        else {

            isExecuteAllButtonVisible = false;
        }
        getExecuteAllButton().setVisible(isExecuteAllButtonVisible);
    }

```
I think this would work as it was before you added the left and right panels.  Or, did you try this and find it didn't work?  It seems cleaner, assuming of course it works without the left/right panels.

johnbrvc (commented):
Looks good and seems to work as advertised.  The "how to test" instructions were a bit misleading since they imply there will be more than one problem to select from when turning off computer judging.  When I loaded 'mini' there was exactly one problem, **sumit**. No big deal, just pointing it out.

I would like a to see a response to the comment I made about left/right panels and the `setVisible() `business I suggested before I approve this.

src/edu/csus/ecs/pc2/ui/TestResultsPane.java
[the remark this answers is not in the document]
SamanwaySadhu: Updated as per suggestion. Earlier I had tried this approach, must have done something wrong as I was seeing a large space when the Execute All button set to be invisible.

johnbrvc (approved):
Looks good.  Works as I would expect it to.

clevengr (approved):
I reviewed the updated code, then built and ran the system.  Works great.  @johnbrvc 's solution (toggling visibility instead of using two sub-panels) was a better suggestion than mine; I'm glad you went ahead and implemented it.
Marking PR as "Approved".

PR #909, opened 2024-01-07
i_900: add support for multiple groups per team

johnbrvc:
### Description of what the PR does
This PR provides built-in support for an account to belong to multiple groups, instead of just one.  The allows per-group display of standings, scoreboard generation and report generation.

The following changes were made to support multiple groups:

1. Add `HashSet<ElementId> groupIds` to `Account `object - this is a hash table containing all groups the account is a member of
2. Rename the old group id to `primaryGroupId` which is now the single **CMS** supplied group ID.  These is currently used by several unit tests, and the **DSA**/**NSA** for group ranking.  (Note: Group ranking has to be fixed so it ranks the team for each group the team is a member of.  This is a non-trivial change).
3. Add methods to `Account` to support adding a group to an account, fetching the groups for an account, clearing the groups.
4. Change PC2 API `ITeam`, `ITeamImplementation` to support multiple groups for the PC2 api (**WTI**) interface.
5. Change **NSA**  to add all groups to generated XML for standings
6. `PacketFormatter `shows all groups for an account.
7. In `ContestXML`, when generating contest xml, add all groups for each account
8. In `ExportAccounts`, include all groups for each account
9. Change `ICPCAccount` to support multiple groups, although currently, only one is ever present in an ICPC "_tab_" file.  Change made for consistency.
10. Change `ICPCImportData` to pass all groups when creating an `ICPCAccount`
11. Fix `Account `report generation to include all groups the team is a member of in the report
12. Add support in `ScoreboardUtilities` to support multiple groups for routines that previously only checked one group per account.
13. Add method `getGroupFilteredRuns()` to `ScoreboardUtilities` to return runs appropriate for selected groups only. (used by **DSA**/**NSA**)
14. Change `JSONTool` to generate multiple groups per account
15. In `PacketHandler`, change the routine that calculates which problems a team can see (it's now based on multiple groups, not just one)
16. Change `LoadAccounts` to accept a CSV list of groups on importing accounts instead of just a single group.
17. In `CLICSAwardUtilities`, change getGroupForTeam to getGroupsForTeam
18. In `InternalContest`, change `addProblem` to account for all groups the team belongs to
19. `ProblemGroupAssignmentReport` was changed to check all groups for a team.
20. In `EventFeedXML`, `EventFeedXML2013` and `ResolverEventFeedXML` add all groups a team is a member of to XML feed as separate group tags
21. In `TeamData`, use renamed `primaryGroupId` - it's ok to only support one group here since this is a legacy feed object
[...]

Issue #900, opened 2023-12-14
Allow teams to be members of multiple groups


johnbrvc:
**Is your feature request related to a problem?** 
Yes.  Teams may only be a member of a single group.  This makes doing things like #250 very difficult. 
...

**Feature Description**: 
This would be a part of a larger feature to support generation of scoreboards for a specific group.  If a team can be a member of, say, a **_site_** group (_Hawaii_, _Oregon_, _Washington_, etc.), a **_level_** group (_D1_, _D2_, _sophomores_, etc) and a _**site-level**_ group, then a scoreboard could be requested for any **_site_**, **_level_**, or **_site-level_** group and include the teams in the appropriate group.

Code has already been added to the DSA to calculate a score based on group since the _CLICS 2023-06 API_ supports that.  The problem is, PC2 only allows a single group per team.  
...

**Have you considered other ways to accomplish the same thing?** 
Yes.  **PacNW** has a workaround by massaging the scoreboards using **_XSLT_** for showing division scoreboards.  With the addition of multiple groups per team, that would no longer be necessary.
...

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?**
Instead of a single group in the `Account `object, allow an array / collection of groups.  Change all code that access the group property to allow multiple groups.  The GUI should be changed right away to show all groups a team belongs to. 
...

**Additional context**: 
#250 #864 
...

johnbrvc:
PR #909 Fixes this.

PR #919, opened 2024-01-28
cleanup the xsl based on group stuff done in #900

### LINKED ISSUE #917: Standings index XSL files need to be updated for multiple groups
> > > > also maybe the samps/web/xsl xsl should be updated like index.xsl was changed?
> > > 
> > > 
> > > actually, thinking when the groups gets committed those samps/web/xsl are unneeded?
> > 
> > 
> > Probably not.
> > The real "pc2" provided `data/xsl/index.xsl` will have to be fixed to provide links to other groups' standings pages (assuming, such other groups exist). In addition, it will have to be aware - somehow - of what group **IT** is, so it doesn't include a link to itself in the list of links. That is, we can't simply `for-each` group, generate a standings link - or it would generate a link to itself. For **_PacNW_**, it looks like we'll need a custom one as well, since you add other things to the standings page (technical notes link, sample data link), as well as don't include the "other" division's groups. A D1 page will not include all the D2 links, only a link to the "D2" page. I'm not sure how that will work. At GNY we provide a link on the desktop to the "problem set page", which has links to the problem statements (all zipped or individual pdf), as well as links to the sample data (all zipped up, or individual files) on THAT page. You guys do it a bit differently. image
> 
> Okay, in that case, can we update the samps/web stuff to follow the changes made to the data/xsl bits.

I guess the question is, should we make this a separate issue?  These XSL changes are only relevant for "group" type standings, which is really not the scope of this issue/PR.

_Originally posted by @johnbrvc in https://github.com/pc2ccs/pc2v9/issues/915#issuecomment-1908050028_

troy2914:
fixes #917

can be tested like this
java -cp ../pc2v9/dist/pc2.jar edu.csus.ecs.pc2.core.util.XSLTransformer index.xsl results_D1.xml index_D1.html
java -cp ../pc2v9/dist/pc2.jar edu.csus.ecs.pc2.core.util.XSLTransformer group.xsl results_Hawii - D1.xml group_Hawaii  - D1.html

with the results file from #909 (possibly modified to remove the site (non-D1/D2 groups).


johnbrvc (approved):
Since these files are in the **samps** folder, they have no direct impact for the average user.  These are specific to those folks looking to do group specific rankings and will require changes by anyone who wishes to do something similar.

PR #944, opened 2024-03-20
i_943 Fix Group tab on Edit Problem

### LINKED ISSUE #943: When editing a problem, the Groups tab operates poorly
**Describe the issue**: 
The **Group** tab is supposed to block you from checking/unchecking groups if the **Show to all groups** radio button is selected.  It does not.  
When there are many groups (20+) and many (or big) test case data files, the **Edit Problem** dialog takes an inordinately long amount of time to come up.

**To Reproduce**: 
Part 1 (non-disabled checkboxes):
1) Load a contest that has some groups defined
2) Go to the **Problems** tab
3) Add a problem (or edit an existing one)
4) Go to the **Groups** tab
5) Select the **Show to all groups** radio button
6) Attempt to check some of the groups in the group box check list
7) even though they're disabled you can check them still.

Part 2 (slowness loading):
1) Load a contest (such as `wfluxor/contests/wf47/systest1`) that has lots of groups defined (>20) and problems with lots of test cases and/or big test case files.
2) Attempt to edit one of these problems  (note that it takes a long long time)

**Expected behavior**: 
Part 1:  If **Show to all groups** is selected, then the checkboxes should be disabled.
Part 2:  The edit problem box should come up in a reasonable amount of time. 

**Actual behavior**: 
Part 1:  If **Show to all groups** is selected, then the checkboxes are not disabled (despite what the code expects)
Part 2:  The edit problem box takes a long time to come up.  Going to the **Groups** tab and selecting groups on and off shows
   a noticeable lag between clicks.

**Environment**: 
Windows 11
Java jdk1.8.0_321

**Log Info**: 

**Screenshots**: 

**Additional context**: 
In a **JCheckboxJList** control, enabling/disabling the individual checkboxes appears to have no effect.  Instead, you should enable/disable the entire **JCheckboxJList** control itself.  That has the desired effect.
Fixing the problem as described above, also fixes the slow loading of the problem by not causes excessive check box state change events to be processed.  This prevents excessive re-reads of the problem data when enabling the Update button for each checkbox click.

johnbrvc:
Do not enable/disable individual group check boxes. This causes excesseive overhad when checking for the **Update** button.
Instead, enable/disable the entire `JCheckBoxJList `control which CONTAINS the checkboxes. This appears to be the original author's intent.

Left some commented-out code with new comments to help in code review. Will be removed prior to merging after approved.

Fixes #943 
Windows 11
jdk1.8.0_321

Part 1:

1. Load a contest that has some groups defined
2. Go to the **Problems** tab
3. Add a problem (or edit an existing one)
4. Go to the **Groups** tab
5. Select the **Show to all groups** radio button
6. Attempt to check some of the groups in the group box check list
7. Note you can not change the state of the check boxes
8. Select the Show to only these groups radio.
9. Attempt to check some of the groups in the group box check list
10. Note you can now change the state of the check boxes

Part 2 (slowness loading):

1. Load a contest (such as `wfluxor/contests/wf47/systest1`) that has lots of groups defined (>20) and problems with lots of test cases and/or big test case files.
2. Attempt to edit one of these problems (note that that the performance is now reasonable)

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ProblemGroupPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ProblemGroupPane.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.Color;
--- a/src/edu/csus/ecs/pc2/ui/ProblemGroupPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ProblemGroupPane.java
@@ -113,28 +113,6 @@ public class ProblemGroupPane extends JPanePlugin {
             getGroupListBox().setForeground(Color.BLACK);
         }
         getGroupListBox().setEnabled(enable);
-
-        // This loop is wrong and does not do what you think it does.
-        // Apparently, enabling/disabling a checkbox within a JCheckboxJList has no effect (it appears to always be enabled
-        // if the JCheckboxJList is enabled.)
-        // Instead you disable the entire JCheckboxJList component as we do above.  Leaving
-        // this here for code review, but should be removed.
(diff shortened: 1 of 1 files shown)


kkarakas (approved):
LGTM!

troy2914 (approved):
seems reasonable.

Issue #869, opened 2023-11-20
Executable method substituteAllStrings has a typo causing substitution failure for "{:files}"


clevengr:
**Describe the issue**:

In method `Executable.substituteAllStrings(Run, String, int)`, there is the following set of code lines to perform "variable substitutions":
```
            newString = replaceString(origString, "{:mainfile}", runFiles.getMainFile().getName());
            newString = replaceString(newString, "{files}", ExecuteUtilities.getAllSubmittedFilenames(runFiles));
            newString = replaceString(newString, "{:basename}", removeExtension(runFiles.getMainFile().getName()));
            newString = replaceString(newString, "{:package}", packageName);
```

Notice that in the second line, the colon is missing in the second parameter (`"{files}"`).  This is causing a substitution failure for the `{:files}` variable.

**To Reproduce**: 

- Start a server using **--load sumithello** 
- Start an Admin and do the following:
  - Edit the Java language by _replacing_ the substitution string {:mainfile} with the substitution string `{:files}` in the **Compile Cmd Line**.
  - Edit problem `hello` by changing the "Judging Type" from "computer" to "manual" (this is so you can see the results without an auto-judge interfering).
- Start a team and submit  the `hello.java` program from the `samps/src` folder as a Java solution to the "Hello" problem.
- Start a judge _which is not configured to autojudge problem "hello"_ (in the `sumithello` sample contest, Judge1 will work).
- Fetch the submitted run and execute it.

**Expected behavior**:  
For the case of a single submitted file (such as `hello.java`), the "main file" is the same as the "list of files".  Therefore, the run should compile and execute correctly. 

**Actual behavior**: 
Compilation fails with the error message
```
Compilation error when compiling program "hello.java" using the command "javac {:files}"
```
and the following message on the compiler stderr:
```
error: invalid flag: {:files}
Usage: javac <options> <source files>
use --help for a list of possible options
```
Clearly, the substitution of "hello.java" for `{:files}` did not work as it should have.

**Environment**: 
 - Windows 11, Eclipse Version: 2023-03 (4.27.0), Java version 1.8.0_371
 - PC2 Version 9.9-develop  

**Additional context**: 
This is a one-line (one-character, actually) fix:  add the missing colon to the `{:files}` substitution in method `Executable.substituteAllStrings()` (around line 2946).  I've actually tested this and verified the fix;  I'll try to get A Round Tuit for submitting a PR for it.

clevengr:
[...]

PR #945, opened 2024-03-20
i_942 Fix entry points with periods

### LINKED ISSUE #942: Submissions with entry points containing periods are not processed properly
**Describe the issue**: 
This was noticed during the WF Luxor wf47/systest1 pre-contest test where PC2 was shadowing DOMjudge.  If a submission (such as a Java submission) arrives on the event feed and includes an **entry_point** property of `nwerc2021.Knitpicking`, PC2 will trim this to `nwerc2021`, which causes the run to get a compile error since the appropriate class file can not be found.  A similar issue with Python submissions where the supplied **entry_point** is specified as `flatlandolympics.py`, PC2 will trim it `flatlandolympics`, causing it to fail.

**To Reproduce**:
Requires a great deal of set-up.  You will need a primary that can feed very specific _submission_ events to the shadow.  For example, the events have to look something like this:
```
[
   {
      "external_id" : null,
      "entry_point" : "nwerc2021.Knitpicking",
      "team_id" : "47037",
      "contest_time" : "0:10:50.578",
      "id" : "359",
      "problem_id" : "knitpicking",
      "files" : [
         {
            "filename" : "submission.zip",
            "mime" : "application/zip",
            "href" : "contests/wf47_systest1/submissions/359/files"
         }
      ],
      "time" : "2024-03-16T14:10:50.578+02:00",
      "language_id" : "java",
      "import_error" : null
   },
   {
      "problem_id" : "flatlandolympics",
      "files" : [
         {
            "href" : "contests/wf47_systest1/submissions/1346/files",
            "mime" : "application/zip",
            "filename" : "submission.zip"
         }
      ],
      "contest_time" : "1:23:50.568",
      "id" : "1346",
      "entry_point" : "flatlandolympics.py",
      "team_id" : "47043",
      "external_id" : null,
      "language_id" : "python3",
      "import_error" : null,
      "time" : "2024-03-16T15:23:50.568+02:00"
   }
]

```
This can only be achieved over the event feed and not the the pc2team fat client.  This may mean hand-crafting events.  

**Expected behavior**:
The **entry_point**'s provided in the event-feed should be utilized correctly, and not stripped after the period.

**Actual behavior**: 
Debugging reveals that the first **entry_point**, `nwerc2023.Knitpicking` gets changed to `nwerc2023`.  And the second, `flatlandolympics.py` gets changed to `flatlandolympics`.  

**Environment**: 
Windows 11
Ubuntu 22.04
jdk1.8.0_321

**Log Info**:
 

**Screenshots**:  

**Additional context**:
The `setEntryPoint()` method in **Run.java** is the one responsible for stripping the **entry_point** of the required information by calling the `Utilities.getFileBaseName()` method.  This should be fixed to possibly adjust the **entry_point** based on the language supplied.

johnbrvc:
If an `entry_point `was specified in a "**submissions**" event on the event-feed that contained a package (for Java) or an explicit file (for Python3), the run would fail (_compile error_ for Java and _Run time Exception_ for Python3). Add code to handle the special cases for each language that can specify an entry point. Code works the same as before if no `entry_point `specified.
Remove pruning of `entry_point `when it is set in the **Run**.   This was removing important information for Java entry points with package information specified and extensions on Python files.

CI: Add constant for {files} command substitution string and change in places it's used.
NOTE: This may relate to issue #869 in that the {files} substitution variable is mistyped in the code perhaps.  It should be {:files} probably.  That change is not part of this code, but it is now easier to fix that, should we decide to do it.  It's unclear if if {files} is a type or legacy nomenclature.

Fixes #942 

Windows 11
Java jdk1.8.0_321

Refer to the issue #942.  This is very difficult to test without a shadow system, and the procedure described there is lengthy and involved.

A good thorough code review is probably a good first step.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/Constants.java
+++ b/src/edu/csus/ecs/pc2/core/Constants.java
@@ -1,15 +1,15 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core;
 
 /**
  * Constants for pc2.
- * 
+ *
  * @author Douglas A. Lane, PC^2 Team, pc2@ecs.csus.edu
  */
 public final class Constants {
-    
+
     public static final String AWARDS_JSON_FILENAME = "awards.json";
-    
+
     public final static String SCOREBOARD_JSON_FILENAME = "scoreboard.json";
 
     public static final long HOURS_PER_DAY = 24;
@@ -19,7 +19,7 @@ public final class Constants {
     }
 
     // SerializedFile constants.
-    
+
     public static final int FILETYPE_BINARY = 1;
 
     public static final String FILETYPE_BINARY_TEXT = "Binary";
@@ -71,22 +71,22 @@ public final class Constants {
      * number of seconds in an hour.
      */
     public static final long SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR;
-    
+
     public static final long SECONDS_PER_DAY = 86400;
 
     /**
      * Default contest length.
      */
     public static final long DEFAULT_CONTEST_LENGTH_SECONDS = 18000; // 5 * 60 * 60
-    
-    
+
+
     public static final int DEFAULT_MAX_OUTPUT_SIZE_K = 512;
 
     /**
      * PC<sup>2</sup> Validator Program Name.
      */
     public static final String PC2_VALIDATOR_NAME = "edu.csus.ecs.pc2.validator.pc2Validator.PC2Validator";
-    
+
     /**
      * CLICS Validator Program Name.
      */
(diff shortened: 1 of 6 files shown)


troy2914 (commented):
These changes look okay.

troy2914 (approved):
new change look reasonable to me.

kkarakas (approved):
I think it looks good!

clevengr:
Regarding the comment in the PR description:
```
It's unclear if if {files} is a type or legacy nomenclature.
```
I'm pretty sure this is typo (which is what I assume "type" in the above meant).  I don't recall that we ever had "command substitution" that didn't use a colon.

src/edu/csus/ecs/pc2/core/Constants.java
```diff
@@ -96,18 +96,23 @@ private Constants() {
...
     /**
      * Command substitution variable for name of thing to execute (basename of source file)
      */
     public static final String CMDSUB_BASENAME_VARNAME = "{:basename}";
-    
+
+    /**
+     * Command substitution variable for all files submitted. Note the missing ':' - legacy.
+     */
+    public static final String CMSSUB_FILES_VARNAME = "{files}";
```
clevengr, line 114: "CMSSUB..." seems to me to be a weird name.  Was the intention that it be "CMDSUB...", as in "Command Substitution"? If so, I'd like to see that corrected, for the sake of keeping code clean and clear...

Also, I haven't yet reviewed the rest of the code but:  there are just one or two places where `{files}` (without the colon) is used; all other places use the proper form (`{:files}`).  It seems weird to me to provide a "code solution" like this to fix something that is (almost certainly) a typo.  Why not just correct the typo in the one or two places it occurs?
And on a related note (again, I haven't reviewed the rest of the code changes), are there OTHER places that are now using this substitution constant (when they should be using `{:files}`) ?

Having said this, I can see filing a "clean-up" Issue that suggests defining a string like "CMDSUB_FILES_VARNAME" -- and then replacing all hard-coded instances of `{:files}` with that string....  But I don't like having a string like this that uses the WRONG value.  It's too possible for someone in the future to think "oh, there's a CONST for this substitution so I should use that" -- thus getting the WRONG substitution.  Brrrr... brittle code.

johnbrvc: Heh.  Yeah, that is a typo **CMSSUB** vs **CMDSUB**.
w/r/t `{files} `vs. `{:files}`, I was just concerned that `{files}` has been there so long (since 2017 when Doug added it), that if any CDP counts on `{files}` will break and any 
users that may use the GUI and "know" to use `{files}`, will now find it no longer works.  Are we concerned about "backward" compatibility for stuff that was wrong in the first place and was done 7 years ago?

johnbrvc: BTW, I actually mention https://github.com/pc2ccs/pc2v9/issues/869 in the PR description, which is what you are referring to.  Quite frankly, I'm inclined to leave {files} there and ADD {:files} as well, but others may have a different opinion.  (I hate to break things by unbreaking things).

PR #947, opened 2024-03-23
i_946 Preserve problem title if no TEX files

### LINKED ISSUE #946: Problem Title is lost if there are no TEX files.
**Describe the issue**: 
If a **title:** property is found in the `problem.yaml `file, and there are no `problem_statement/problem.tex` and `problem_statement/problem.en.tex` files, the title is lost.

**To Reproduce**: 

1. Create a `problem.yaml` file that has a **title** property.
2. Make sure there are no TEX files in the `problem_statement` folder.
3. Load the contest.
4. It will fail indicating that the problem does not have a name.

**Expected behavior**: 
If a **title** is specified in the `problem.yaml`, it should not be lost if there are no `problem_statement `TEX files. 

**Actual behavior**: 
The **title** (name) is lost of there are no `problem_statement `TEX files.

**Environment**: 
All.

**Log Info**:  

**Screenshots**:  

**Additional context**: 
In the  **ContestSnakeYAMLLoader**, the name of the problem (**problemTitle**) is obtained from the **title** property in the `problem.yaml`.  Then, the existence of `problem_statement/problem.tex` is checked.  If it is found, then the file is searched for the problem name, and that is used for **problemTItle**, whether it is found or not.  If the file is **not** found, then if **problem_statement/problem.en.tex** can not be opened, the name of the problem (**problemTitle**) obtained from the `problem.yaml` is set to **null** because the file can not be opened, and the problem name can not be found.  The existence of the `problem_statement/problem.en.tex `should first be checked before blindly reading the (non-existent) file and setting the **problemTitle**.

johnbrvc:
Add check for existence of `problem_statement/problem.en.tex` before trying to get the problem name (**problemTitle**) from it.

Fixes #946 

Windows 11
jdk1.8.0_321

1. Create contest CDP with a single problem and a `problem.yaml` file that has a **title** property giving the name of the problem.
2. Make sure there are no TEX files in the `problem_statement `folder.
3. Load the contest.
4. It will succeed and use the title provided in the` problem.yaml `as the problem  name.

NOTE: I did NOT remove all the **typecast**'s throughout the code.  That was Eclipse.  I even tried putting them back, but it insisted on removing the redundant **typecast's**.  The REAL code changes are at line 1409 and 1411 - so you can skip the rest and go right there.  Also, all the space removals at the end of lines, etc. were a result of eclipse "cleaning up" the code.  If you want me to turn that option off, I can do that.   Personally, I find it extremely annoying, even for me - it makes finding the real change extremely difficult.  I guess a case could be made, that whoever did that originally (added all the spaces at the end of lines, etc.) is the culprit.  Nevertheless, it's still bothersome.

### DIFF
--- a/src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
+++ b/src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2023 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.imports.ccs;
 
 import java.io.ByteArrayInputStream;
--- a/src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
+++ b/src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
@@ -71,7 +71,7 @@ import edu.csus.ecs.pc2.validator.pc2Validator.PC2ValidatorSettings;
 
 /**
  * Load contest/model from Yaml and files in a CDP.
- * 
+ *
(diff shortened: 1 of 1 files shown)


troy2914 (commented):
need to update the copyright year.

troy2914 (approved):
looks good now.

PR #951, opened 2024-04-01
Only substring 5, if the string contains "No - ".

### LINKED ISSUE #950: When a judgement display name is less than 5 chars, the event feeder will cause StringIndexOutOfBounds exception
**Describe the issue**: 
The event feed does not get generated if a judgement type description is less than 5 characters.

**To Reproduce**: 

1. On the **Administrator** **Judgements** tab, create a judgement where the "**Judgement**" field is less than 5 characters. 
2. image
3. Then go to the **Reports** tab, and attempt to  **Export Report Contents** of the **Event Feed JSON**.
4. An exception will be generated.

**Expected behavior**: 
The report should be generated properly. 

**Actual behavior**: 
image

**Environment**: 
Ubuntu 22.04
Java 17.0.10

**Log Info**:  
See above.

**Screenshots**: 
See above

troy2914:
fixes #950

See the issue for testing.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/util/JSONTool.java
+++ b/src/edu/csus/ecs/pc2/core/util/JSONTool.java
@@ -312,7 +312,9 @@ public class JSONTool {
             solved = true;
             penalty = false;
         } else {
-            name = name.substring(5, name.length());
+            if (name.startsWith("No - ")) {
+                name = name.substring(5, name.length());
+            }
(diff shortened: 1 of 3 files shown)


johnbrvc (approved):
The original code was clearly wrong.  The intent was to check if it began in "No - ", and strip that off (for the event feed).  This fix addresses that.

PR #964, opened 2024-05-25
I963: scoreboard and compare runs frame are only restricted to one instance.

### LINKED ISSUE #963: Should not be allowed to open more than one compare runs window (or compare scoreboards) from event feed client
**Describe the issue**: 
The event feeder allows you to open more than one shadow runs compare window.  This can cause check-box state synchronization problems, and, there is no need to have more than one compare results window open. 

**To Reproduce**: 
On the event feeder client, while shadowing, simply press the "Compare Runs" twice.  Two compare runs windows will open.

**Expected behavior**:
Only one "Compare Runs" window should be shown.

**Actual behavior**: 
Multiple "Compare Runs" windows can appear - one for each compare button press.

**Environment**: 
All.

**Log Info**: 
**Screenshots**: 

**Additional context**:

kkarakas:

scoreboard and compare runs frame are only restricted to one instance. If user attempts to open the same frame one more time user is warned and the action is not completed.

fixes #963 

open feeder. Once connected to ccs api, attempt to open compare scoreboard and compare runs. Try to open them again without closing them. User should be warned by the system it is not possible and there is already compare scoreboard or compare runs. Also notice that the old compares will come to front if it is behind other stuff or will get unminimized if it was minimized. Notice that problem stated below does not exist anymore.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
@@ -77,16 +77,16 @@ import edu.csus.ecs.pc2.shadow.ShadowJudgementPair;
 public class ShadowCompareRunsPane extends JPanePlugin {
 
     private static final long serialVersionUID = 1L;
-    
+
     private static final int RUN_UPDATE_REQUEST_SERVER_TIMEOUT_MILLIS = 30000;
-    
+
     private static final String DEFAULT_REFRESH_INTERVAL = "5";
-    
+
     private ShadowController shadowController = null ;
-    
+
     //the current judgement information from the shadow controller
     private Map<String, ShadowJudgementInfo> currentJudgementMap = null;
-    
+
     private Map<String, ShadowJudgementInfo> filteredJudgementMap = null;
     
     //the table displaying the current results
@@ -102,16 +102,16 @@ public class ShadowCompareRunsPane extends JPanePlugin {
     private Run runWeRequestedServerToUpdate;
 
     private boolean serverHasUpdatedOurRun;
-    
+
     private JPanel dynamicallyRefreshPanel;
-        
+ 
     private JCheckBox mismatchCheckBox;
 
     @Override
     public String getPluginTitle() {
         return "Shadow_Compare_Pane";
     }
-    
+
     /**
      * This GUI class accepts a reference to a {@link ShadowController}, from which it obtains (by calling 
      * {@link ShadowController#getJudgementComparisonInfo()}) a
@@ -127,11 +127,11 @@ public class ShadowCompareRunsPane extends JPanePlugin {
         Dimension size = new Dimension(700,700);
         this.setPreferredSize(size);
         this.setMinimumSize(size);
-        
+
         this.shadowController = shadowController ;
-        
+
         this.log = shadowController.getLog();
-        
+
         this.setContestAndController(shadowController.getLocalContest(), shadowController.getLocalController());
 
         //add a run listener so we can be notified when run edits which we invoke (during the "Resolve Run" operation) are completed
@@ -141,13 +141,13 @@ public class ShadowCompareRunsPane extends JPanePlugin {
         JLabel header = new JLabel("Comparison of PC2 vs. Remote Judgements");
         header.setAlignmentX(Component.CENTER_ALIGNMENT);
         this.add(header);
-        
+
         //get the framework for the table which will be used to display comparison results
         resultsTable = getResultsTable();
         
         //put the current comparison results into the table model
         resultsTable.setModel(getUpdatedResultsTableModel());
-        
+
         //support sorting the table by clicking on the column headers
         TableRowSorter<TableModel> sorter = new TableRowSorter<TableModel>(resultsTable.getModel());
         resultsTable.setRowSorter(sorter);
@@ -156,13 +156,18 @@ public class ShadowCompareRunsPane extends JPanePlugin {
         //put the results table in a scrollpane on the GUI
         JScrollPane scrollPane = new JScrollPane(resultsTable, ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED,
                     ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
+
+        shadowController.setFilter(FILTERS.NONE);//Required to reset the state of the filter
+
         this.add(scrollPane);
 
         this.add(getSummaryPanel());
-        
+
         this.add(getdynamicallyRefreshPanel());
-        
+
         this.add(getButtonPanel());
+
+
     }
         
     /**
@@ -179,7 +184,7 @@ public class ShadowCompareRunsPane extends JPanePlugin {
 
         JTable resultsTable = new JTable() {
             private static final long serialVersionUID = 1L;
-            
+
             private Border outside = new MatteBorder(1, 0, 1, 0, Color.RED);
             private Border inside = new EmptyBorder(0, 1, 0, 1);
             private Border highlight = new CompoundBorder(outside, inside);
@@ -207,7 +212,7 @@ public class ShadowCompareRunsPane extends JPanePlugin {
                      (remoteJudgement != null && remoteJudgement.toLowerCase().contains("pending"))) {
                     c.setBackground(new Color(255, 255, 153));
                 }
-                
+
                 // update font to bold & italic if row is selected
                 if (isRowSelected(row)) {
                     c.setFont(new Font("Arial Bold", Font.ITALIC, 14));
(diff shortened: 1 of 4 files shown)


johnbrvc:
Seems to work, however one question:  Is there a reason you can't just check if `shadowCompareRunsFrame != null && shadowCompareRunsFrame.isVisible()` instead of creating a new member variable (`isOpen`)?  Same for the scoreboard compare?  Is it possible for a user to actually minimize the window causing it to not be visible?  If that's the case, then you could check` isDisplayable() `and if so, make it visible (bring it out of hiding).  `isDisplayable()` will return false if the window has been closed (disposed of).  Using the exising Swing framework might be better, assuming there is no other reason why it wouldn't work, right?

kkarakas:
> Seems to work, however one question: Is there a reason you can't just check if `shadowCompareRunsFrame != null && shadowCompareRunsFrame.isVisible()` instead of creating a new member variable (`isOpen`)? Same for the scoreboard compare? Is it possible for a user to actually minimize the window causing it to not be visible? If that's the case, then you could check`isDisplayable()`and if so, make it visible (bring it out of hiding). `isDisplayable()` will return false if the window has been closed (disposed of). Using the exising Swing framework might be better, assuming there is no other reason why it wouldn't work, right?

Good ideas let me look at them.

johnbrvc:
During testing today (NOT using your PR, so I'm not sure if this bug is still there or not, but please check):
If I have one compare runs frame up and I have "**Only Mismatched**" checked, then I close that frame.  Then I reopen the compare runs frame, it shows only mismatched runs, but the checkbox is not selected (see below).  If I then select the box, nothing new happens, but if I deselect it, then all the runs appear.  Basically, it looks like the state of the checkbox is not set upon making the frame visible?
image

kkarakas:
I was unable to replicate the problem you mentioned in the last comment in the new PR so I guess it is solved.

johnbrvc:
> I was unable to replicate the problem you mentioned in the last comment in the new PR so I guess it is solved.

What problem are you referring to?  I don't think I specifically mentioned a problem in this PR?  If it's in a different PR, you can comment there.

kkarakas:
Sorry I should've tagged and be more clear. Yes I am referring the comment you made just above mine.

johnbrvc:
> Sorry I should've tagged and be more clear. Yes I am referring the comment you made just above mine.

The issue happens with the latest "develop" branch.  I did not try it with your new branch in this PR.  If you put code in to fix that, then, of course, it will be fixed.  Right now I am running a test contest (wfastana/contests/systest1).   After the systest1 is done, I'll try to run it using your branch in this PR.

johnbrvc:
> I was unable to replicate the problem you mentioned in the last comment in the new PR so I guess it is solved.

I just tested it with your `i_963_should_not_be_allowed_multiple_compare_runs` branch and the problem occurs.
Basically, start shadowing, bring up **Compare Runs**.  Check the "**Only Mismatched**" check box.  Close the **Compare Runs** window.  Bring up a new **Compare Runs**.  Note that the **Only Mismatched** box is **NOT** checked, and there are no runs being shown until you check the box, and then uncheck it.

johnbrvc (changes requested):
Please make the indicated changes. (split up those lines).

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -126,12 +126,11 @@ public String getPluginTitle() {
     public ShadowCompareRunsPane(ShadowController shadowController) {
         Dimension size = new Dimension(700,700);
         this.setPreferredSize(size);
-        this.setMinimumSize(size);
-        
+        this.setMinimumSize(size);shadowController.setFilter(FILTERS.NONE);//Required to reset the state of the filter
```
johnbrvc, line 129: Break this up into 2 lines.  We don't like to put multiple statements on a single line.

src/edu/csus/ecs/pc2/ui/ShadowCompareRunsPane.java
```diff
@@ -141,28 +140,33 @@ public ShadowCompareRunsPane(ShadowController shadowController) {
         JLabel header = new JLabel("Comparison of PC2 vs. Remote Judgements");
         header.setAlignmentX(Component.CENTER_ALIGNMENT);
         this.add(header);
-        
+ 
+        this.setMinimumSize(size);shadowController.setFilter(FILTERS.NONE);//Required to reset the state of the filter
```
johnbrvc, line 144: Break this up into 2 lines. We don't like to put multiple statements on a single line.

kkarakas: Solved

johnbrvc (changes requested):
I'm not sure if this bug was added as a result of reusing the compare runs frame (and probably, similarly, the compare scoreboards frame).   If you start shadowing, bring up the compare runs frame/pnae and enable the refresh (checkbox).  Then, you Stop Shadowing, while the compare frame is still open, it looks like the timer is still running for refresh, and every 5 seconds, you get an exception in the log.  There does not appear to be a way to stop this, except by exiting from the event feeder client and restarting it.   It seems if the compare runs frame is closed the timer should be stopped/cancelled.  This can happen at the end of contests when the feed is finished.  We sometimes stop the shadowing, but leave the compare windows up.  Please confirm this is a result of not recreating the compare frame each time.  Either way, this should probably be fixed before this is approved.

kkarakas:
> I'm not sure if this bug was added as a result of reusing the compare runs frame (and probably, similarly, the compare scoreboards frame). If you start shadowing, bring up the compare runs frame/pnae and enable the refresh (checkbox). Then, you Stop Shadowing, while the compare frame is still open, it looks like the timer is still running for refresh, and every 5 seconds, you get an exception in the log. There does not appear to be a way to stop this, except by exiting from the event feeder client and restarting it. It seems if the compare runs frame is closed the timer should be stopped/cancelled. This can happen at the end of contests when the feed is finished. We sometimes stop the shadowing, but leave the compare windows up. Please confirm this is a result of not recreating the compare frame each time. Either way, this should probably be fixed before this is approved.

I am not sure what you are referring to by saying reusing the compare runs frame. In the code a new object of compare runs frame is created however shadowController is not recreated. I tried to test it and I observed the bug you have stated. Similarly to this bug, if you press refresh button while shadowing is stopped from the feeder window, it will also give an exception. And given that "refresh every: _ seconds" button calls a method that refresh button activates he bug itself is not new.

johnbrvc:
> > I'm not sure if this bug was added as a result of reusing the compare runs frame (and probably, similarly, the compare scoreboards frame). If you start shadowing, bring up the compare runs frame/pnae and enable the refresh (checkbox). Then, you Stop Shadowing, while the compare frame is still open, it looks like the timer is still running for refresh, and every 5 seconds, you get an exception in the log. There does not appear to be a way to stop this, except by exiting from the event feeder client and restarting it. It seems if the compare runs frame is closed the timer should be stopped/cancelled. This can happen at the end of contests when the feed is finished. We sometimes stop the shadowing, but leave the compare windows up. Please confirm this is a result of not recreating the compare frame each time. Either way, this should probably be fixed before this is approved.
> 
> I am not sure what you are referring to by saying reusing the compare runs frame. In the code a new object of compare runs frame is created however shadowController is not recreated.
> 
> I tried to test it and I observed the bug you have stated. Similarly to this bug, if you press refresh button while shadowing is stopped from the feeder window, it will also give an exception. And given that "refresh every: _ seconds" button calls a method that refresh button activates he bug itself is not new.

We should probably fix that with this PR as a CI, rather than create a new issue and new PR for that.  THe fix is probably simple enough.  If shadowing is stopped, stop the timer if it is running?  Without looking at the code, I would imagine that would do it.

johnbrvc:
> > > I'm not sure if this bug was added as a result of reusing the compare runs frame (and probably, similarly, the compare scoreboards frame). If you start shadowing, bring up the compare runs frame/pnae and enable the refresh (checkbox). Then, you Stop Shadowing, while the compare frame is still open, it looks like the timer is still running for refresh, and every 5 seconds, you get an exception in the log. There does not appear to be a way to stop this, except by exiting from the event feeder client and restarting it. It seems if the compare runs frame is closed the timer should be stopped/cancelled. This can happen at the end of contests when the feed is finished. We sometimes stop the shadowing, but leave the compare windows up. Please confirm this is a result of not recreating the compare frame each time. Either way, this should probably be fixed before this is approved.
> > 
> > 
> > I am not sure what you are referring to by saying reusing the compare runs frame. In the code a new object of compare runs frame is created however shadowController is not recreated.
> > I tried to test it and I observed the bug you have stated. Similarly to this bug, if you press refresh button while shadowing is stopped from the feeder window, it will also give an exception. And given that "refresh every: _ seconds" button calls a method that refresh button activates he bug itself is not new.
> 
> We should probably fix that with this PR as a CI, rather than create a new issue and new PR for that. THe fix is probably simple enough. If shadowing is stopped, stop the timer if it is running? Without looking at the code, I would imagine that would do it.

A [new issue ](https://github.com/pc2ccs/pc2v9/issues/1011)has been created for this problem.  It should not preclude this PR from being approved.

johnbrvc (approved):

PR #985, opened 2024-06-30
i_984 Make sure validate stdout/stderr buffers are flushed

### LINKED ISSUE #984: Validator stdout and stderr are never copied correctly for each test case
**Describe the issue**:   
The individual validator **stdout** and **stderr** files for a run's test cases are almost always empty (0 bytes).  These files have the names `valout.#.txt` and `valerr.#.txt`, where # is replaced with the test case number.

**To Reproduce**:
Look in any autojudge or human judge's execute folder and observe the lengths of the `valout.#.txt` and `valerr.#.txt` - they are almost always 0.  I say _almost_ because if the validator generates enough output, buffers get flushed (see _Additional Context_ below).

**Expected behavior**: 
The `valout.#.txt` and `valerr.#.txt` files should contain the corresponding output and error outputs of the validator execution for each test case.

**Actual behavior**: 
The files are almost always empty.

**Environment**: 
All.

**Log Info**: 
**Screenshots**: 

**Additional context**:
The problem is related to the fact that the Buffered streams for each channel (stdout, stderr) of the validator are not flushed before the file contents are serialized.  When PC^2 saves the serialized file to each test case's respective files, the serialized files are empty.  A simple `flush()` of the `BufferedOutputStream`'s will fix the problem

johnbrvc:
Flush the validators stdout and stderr BufferedOutputStream buffers before serializing the files to disk.

Fixes #984 
Windows 11
java version "1.8.0_321"

1) Load a contest that has a custom output validator that produces output on stderr, for example.
2) Submit and judge a run.
3) Note that the valout.1.txt and/or valerr.1.txt are now not empty and contain the appropriate messages.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -1122,6 +1122,9 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
 
 
             executionData.setvalidateTimeMS(System.currentTimeMillis() - startTime);
+            // be sure to flush before serializing.
+            stdoutlog.flush();
+            stderrlog.flush();
             executionData.setValidationStdout(new SerializedFile(prefixExecuteDirname(VALIDATOR_STDOUT_FILENAME)));
             executionData.setValidationStderr(new SerializedFile(prefixExecuteDirname(VALIDATOR_STDERR_FILENAME)));
 


src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -1122,6 +1122,9 @@ protected boolean validateProgram(int dataSetNumber) {
 
 
             executionData.setvalidateTimeMS(System.currentTimeMillis() - startTime);
+            // be sure to flush before serializing.
+            stdoutlog.flush();
+            stderrlog.flush();
```
troy2914, line 1127: is there a reason to not close those logs instead of flush'ing them?

johnbrvc: Yes.  The finally {} clause below does that.

troy2914: why isn't the close done where the flush you added are.  There is no need to keep them open imo.  We are not doing flush's in other locations in Executable.

johnbrvc: The issue is the `finally {}` has the `close()` calls for each stream.  If we close the files where the `flush()`'s are, then we'd close them twice, causing an `IOException `on the second time (in the `finally(){}`), and stack traces galore.  We could change the code so that it removes the `finally(){} `clause altoether, then duplicates the `try{close/close}catch(){}` in 2 places (once where the `flush()`'s were added, and once in the `catch()`).  The files **are** closed almost immediately by the `finally(){}` right after their corresponding files have been serialized.

Can you suggest a better and cleaner way?  Is there anything wrong with using `flush()`?  If so, why have it as part of the API?

troy2914 (approved):

kkarakas (approved):
Seems good!

PR #989, opened 2024-07-18
i_988 Fix NPE in SubmitSampleRunspane

### LINKED ISSUE #988: RunsTablePane works but does not update runs
**Describe the issue**: 
If a run is selected from the `RunsTablePane`, and edited, and saved, the `RunsTablePane `no longer updates.

**To Reproduce**:
Follow the procedure in the issue description.  Quite simple to reproduce.

**Expected behavior**:
The `RunsTablePane `should continue to update after editing runs.

**Actual behavior**:
While the buttons work, and actually perform the desired tasks (eg. **Update**), the runs table never updates to show changes in a run's status.

**Environment**: 
All operating systems.
Versions of PC^2 that have had PR #979 applied (after July 3, 2024)

**Log Info**:  
NPE's in `SubmitSampleRunsPane.getSubmission()` - `submissionList `was null.

**Screenshots**:  

**Additional context**:
When a run was updated, it caused the `RunListener `to call `runChanged `in  `SubmitSampleRunsPane`, which caused an NPE, and stopped anything that caused the run to change from continuing (`RunsTablePane`, for example).
Found during WF48 testing.

johnbrvc:
If the `SubmitSampleRunsPane `was never used, and a run changed, an NPE would result when the `IRunListener `methods were called.   Fixed to simply check the value of offending member, `submissionList` before referencing.  Currently, this only happens for administrator accounts.

Fixes #988 

Windows 11 and Ubuntu 22.04.1 (contest image)
java version "1.8.0_321"

1) Load a contest
2) From an administrator account, judge a run
3) From the administrator account, edit the run and change its status to NEW
4) From the administrator account, judge the run again and note the `RunsTablePane ` now updates - previously, it did not.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/SubmitSampleRunsPane.java
+++ b/src/edu/csus/ecs/pc2/ui/SubmitSampleRunsPane.java
@@ -1263,38 +1263,40 @@ public class SubmitSampleRunsPane extends JPanePlugin {
 
     public void reloadRunList() {
 
-        if (filter.isFilterOn()){
-            getFilterButton().setForeground(Color.BLUE);
-            getFilterButton().setToolTipText("Edit filter - filter ON");
-            rowCountLabel.setForeground(Color.BLUE);
-        } else {
-            getFilterButton().setForeground(Color.BLACK);
-            getFilterButton().setToolTipText("Edit filter");
-            rowCountLabel.setForeground(Color.BLACK);
-        }
+        if(submissionList != null) {
+            if (filter.isFilterOn()){
+                getFilterButton().setForeground(Color.BLUE);
+                getFilterButton().setToolTipText("Edit filter - filter ON");
+                rowCountLabel.setForeground(Color.BLUE);
+            } else {
+                getFilterButton().setForeground(Color.BLACK);
+                getFilterButton().setToolTipText("Edit filter");
+                rowCountLabel.setForeground(Color.BLACK);
+            }
 
-        for (SubmissionSample sub : submissionList) {
-            ClientId clientId = null;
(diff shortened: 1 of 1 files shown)


SamanwaySadhu:
> Currently, this only happens for administrator accounts.

What do you mean by that? The issue was only happening from admin accounts, or the fix only works through the admin account?

SamanwaySadhu:
> 4. From the administrator account, judge the run again and note the `RunsTablePane ` now updates - previously, it did not.

By `RunsTablePane` updating, you mean the status and Judge tabs data updates for a particular submission?

johnbrvc:
> > Currently, this only happens for administrator accounts.
> 
> What do you mean by that? The issue was only happening from admin accounts, or the fix only works through the admin account?

If an administrator is logged in (eg. **administrator1**, **administrator2**, etc.) and a run is changed by any method (eg it is judged, or edited by another administrator), the **RunsTablePlane** would not update with the changed run(s).  It would just sit there in its original state.    This only happens on administrator accounts since those accounts are the only ones that have the **Submit Sample Runs** tab.

johnbrvc:
> > 4. From the administrator account, judge the run again and note the `RunsTablePane ` now updates - previously, it did not.
> 
> By `RunsTablePane` updating, you mean the status and Judge tabs data updates for a particular submission?

If you select the "Runs" tab on an administrator, it will never update when a run changes state due to this bug.

SamanwaySadhu (approved):
Works as described.

kkarakas (approved):
In most of the lines, a tab is prepended. Code looks good.

PR #996, opened 2024-07-25
i995, i968, i975 Scoreboard JSON file always shows 0 for the time of a solved problem

### LINKED ISSUE #968: Scoreboard JSON file always shows 0 for the time of a solved problem
**Describe the issue**: 
The `scoreboard.json` file produced for results (and I am guessing the API Endpoint) shows the "**time**" field as 0 for solved problems.

See also #975 and both these can fixed with one PR.

**To Reproduce**: 
Generate a _scoreboard_ report (or results export report) and look at the `scoreboard.json`.  Find a problem that was solved.  Note the time is 0.

**Expected behavior**: 
For solved problems, the specification indicates the time should reflect the number of minutes into the contest the problem was solved:
image

**Actual behavior**: 
0 is always reported as the time solved.

**Environment**: 
All

**Log Info**:  

**Screenshots**:  
image

**Additional context**: 
Noticed by DOMjudge team during scoreboard compare.

### LINKED ISSUE #975: Scoreboard JSON file should use problem short name for "problem_id" property, not the ElementId
**Describe the issue**:
The `scoreboard.json` file uses an internal ID for the `problem_id `property.  It should probably use the short name for the problem.
image

See also #968 and these can be fixed at the same time perhaps.

**To Reproduce**:
1) Generate a `scoreboard.json` file using the **Reports** tab, `Scoreboard JSON` report, on the _server_ client
2) **Export Report Contents**
3) **Save** (selection name and location for exported JSON)
4) View the generated file (see above in the Description).

image

**Expected behavior**: 
While not explicitly required, it makes more sense to use the problem _short name_, such as "arresteddevelopment" in the example above.    

**Actual behavior**: 
The internal PC2 ID is uses which is comprised of the _short name_ and a hash value (Unique ID).

**Environment**: 
All.

**Log Info**:  

**Screenshots**: 

**Additional context**: See `core/imports/clics/CLICSScoreboard.java`.  This is where the ID is created using `"Id = " + problemIdNumber`  There is already a routine in `core/util/JSONTool.java` called `getProblemId()` that probably gets the right thing.  Maybe use that instead of the whole `"Id = "` business?

### LINKED ISSUE #995: xml file is silently not converted to object (ContestStandings) properly.
The method createContestStandings(String xmlString) .......core/standings/ScoreboardUtilities.java attempts to convert a string in xml format to ContestStandings object. However it makes mistakes in header namely groupList and problems. 

Before the method returns ContestStandings object compare the object and xml string. Informations on groups and problems that are supposed to be in header are gone. That point of the code can be reached by attempting to export Scoreboard JSON report from admin or server.

Header should contain groups and problems.

ContestStandings object will have null for the attributes groupList and problems. 

**Extra**
This doesn't create any issues as these two attributes don't appear to be used at all. However, in order solve #975 , solving this issue helps.

kkarakas:

Fixes the three issue that will be mentioned below. shortName info is added to standing XML's by the scoring algorithms so that shortName can be displayed in ScoreboardJSON. 

ScoreboardJSON report no longer shows 0 for time of solved problems. 

Grouplist and problems are no longer null. They have the correct values. This is a intermediate XML file and it is not used so this can't be seen in UI.

fixes #968 and fixes #995 and fixes #975

For #968 open admin or server. Then select reports. Select Scoreboard JSON from reports. Then export report contents. Save the pop up. Check the file download. Observe that time for solved problems are not 0 anymore. 

For #995 Before the method createContestStandings(String xmlString) .......core/standings/ScoreboardUtilities.java returns ContestStandings object. Compare the object and xml string. Information on groups and problems that were not there is there. That point of the code can be reached by attempting to export Scoreboard JSON report from admin or server.

For #975 open admin or server. Then select reports. Select Scoreboard JSON from reports. Then export report contents. Save the pop up. Check the file download. Observe that problem_id's value is shortName.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/standings/GroupList.java
+++ b/src/edu/csus/ecs/pc2/core/standings/GroupList.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2021 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.standings;
 
 import java.util.List;
@@ -7,6 +7,11 @@ import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 
 /**
  * scoreboard XML group element.
@@ -19,7 +24,7 @@ import javax.xml.bind.annotation.XmlRootElement;
 public class GroupList {
 
     //    <groupList>
-    //    <group externalId="18475" id="4" title="ICPC North America South Division Championship"/>
+    //    <group externalId="18475" id="4" included = "1" title="ICPC North America South Division Championship"/>
     //    </groupList>
 
     public List<ScoringGroup> getGroups() {
@@ -31,6 +36,23 @@ public class GroupList {
     }
 
     @XmlElement(name = "group")
+    @JacksonXmlProperty(localName = "group")
+    @JacksonXmlElementWrapper(useWrapping = false)
     private List<ScoringGroup> groups = null;
-
+    
+    @Override
+    public String toString() {
+        
+        String retStr = "Undefined";
+        ObjectMapper mapper = new ObjectMapper();
+        
+        try {
+            retStr = mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            // TODO pass a log into this class so we can do proper logging
+            e.printStackTrace();
+        }
+        
+        return retStr;
+    }
 }
--- a/src/edu/csus/ecs/pc2/core/standings/GroupList.java
+++ b/src/edu/csus/ecs/pc2/core/standings/GroupList.java
@@ -2,6 +2,7 @@
 package edu.csus.ecs.pc2.core.standings;
 
 import java.util.List;
+import java.util.logging.Level;
 
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
@@ -12,6 +13,9 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+
+import edu.csus.ecs.pc2.core.log.Log;
+import edu.csus.ecs.pc2.core.log.StaticLog;
 
 /**
  * scoreboard XML group element.
@@ -49,7 +53,8 @@ public class GroupList {
         try {
             retStr = mapper.writeValueAsString(this);
         } catch (JsonProcessingException e) {
-            // TODO pass a log into this class so we can do proper logging
+            Log log = StaticLog.getLog();
+            log.log(Level.WARNING, "GroupList was unable to be turned into a JSON", e);
             e.printStackTrace();
         }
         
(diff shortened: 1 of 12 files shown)


SamanwaySadhu (approved):
Reviewed only code changes and not functionality.

src/edu/csus/ecs/pc2/core/standings/GroupList.java
```diff
@@ -31,6 +36,23 @@ public void setGroups(List<ScoringGroup> groups) {
...
+    @Override
+    public String toString() {
+        
+        String retStr = "Undefined";
+        ObjectMapper mapper = new ObjectMapper();
+        
+        try {
+            retStr = mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            // TODO pass a log into this class so we can do proper logging
```
johnbrvc, line 52: You should be able to use StaticLog.getLog().

johnbrvc (changes requested):
In the generated scoreboard json report, if a team attempts a problem, but does not solve it, the time property is set to **-1**.  This is probably because in the `DSA.calcProblemScoreData`, it initialized `solutionTime = -1`.  If it never finds a solved submissions for the problem, this is the value that gets set as the time for problem.  

The correct thing to do is if the `solved `property is `false`, **LEAVE OUT** the `time `property altogether. It should not appear.
```
{
   "rank": 122,
   "team_id": 46090,
   "score": {
      "total_time": 31,
      "num_solved": 1
   },
   "problems": [
   {
      "solved": false,
      "num_judged": 1,
      "time": -1,
      "problem_id": "turningred",
      "num_pending": 0
   },
```
would be:
```
{
   "rank": 122,
   "team_id": 46090,
   "score": {
      "total_time": 31,
      "num_solved": 1
   },
   "problems": [
   {
      "solved": false,
      "num_judged": 1,
      "problem_id": "turningred",
      "num_pending": 0
   },
```

Everything else seems to work, so once this is fixed, I will approve it.

kkarakas:
> In the generated scoreboard json report, if a team attempts a problem, but does not solve it, the time property is set to **-1**. This is probably because in the `DSA.calcProblemScoreData`, it initialized `solutionTime = -1`. If it never finds a solved submissions for the problem, this is the value that gets set as the time for problem.
> 
> The correct thing to do is if the `solved `property is `false`, **LEAVE OUT** the `time `property altogether. It should not appear.
> 
> ```
> {
>    "rank": 122,
>    "team_id": 46090,
>    "score": {
>       "total_time": 31,
>       "num_solved": 1
>    },
>    "problems": [
>    {
>       "solved": false,
>       "num_judged": 1,
>       "time": -1,
>       "problem_id": "turningred",
>       "num_pending": 0
>    },
> ```
> 
> would be:
> 
> ```
> {
>    "rank": 122,
>    "team_id": 46090,
>    "score": {
>       "total_time": 31,
>       "num_solved": 1
>    },
>    "problems": [
>    {
>       "solved": false,
>       "num_judged": 1,
>       "problem_id": "turningred",
>       "num_pending": 0
>    },
> ```
> 
> Everything else seems to work, so once this is fixed, I will approve it.

Implemented this, is this the case for DOMjudge? Do they ignore time if it is not solved in their JSON?

johnbrvc:
> > In the generated scoreboard json report, if a team attempts a problem, but does not solve it, the time property is set to **-1**. This is probably because in the `DSA.calcProblemScoreData`, it initialized `solutionTime = -1`. If it never finds a solved submissions for the problem, this is the value that gets set as the time for problem.
> > The correct thing to do is if the `solved `property is `false`, **LEAVE OUT** the `time `property altogether. It should not appear.
> > ```
> > {
> >    "rank": 122,
> >    "team_id": 46090,
> >    "score": {
> >       "total_time": 31,
> >       "num_solved": 1
> >    },
> >    "problems": [
> >    {
> >       "solved": false,
> >       "num_judged": 1,
> >       "time": -1,
> >       "problem_id": "turningred",
> >       "num_pending": 0
> >    },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would be:
> > ```
> > {
> >    "rank": 122,
> >    "team_id": 46090,
> >    "score": {
> >       "total_time": 31,
> >       "num_solved": 1
> >    },
> >    "problems": [
> >    {
> >       "solved": false,
> >       "num_judged": 1,
> >       "problem_id": "turningred",
> >       "num_pending": 0
> >    },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Everything else seems to work, so once this is fixed, I will approve it.
> 
> Implemented this, is this the case for DOMjudge? Do they ignore time if it is not solved in their JSON?

DOMjudge leaves it out if the problem hasn't been solved.

johnbrvc:
My only concern is that it is still possible that the "time" can still be set to -1 (even though the scoreboard json will never display it).  Is there any other place in the code where that could be an issue?  Another problem is, a time of value 0 is technically possible, so, it seems that the "time" field is only valid if solved is true, so it should be up to the code that would potentially use the time value to make sure that the problem was solved, similar to what you just added for the json scoreboard generation.  I dont think we have to change any more code to address this at the moment, so I am prepared to approve it as it, after I test it.

johnbrvc (approved):
Looks ok, and seems to work.

PR #1000, opened 2024-08-08
I999 Add configurable execute grace periods for sandbox runs

### LINKED ISSUE #999: Execution Grace Time for interactive and sandbox problems
With the proliferation of interactive problems at ICPC contests, and the imprecise specification of what a problem's _time limit_ means, it is necessary to add a _Grace Period_, similar to what DOMjudge does, for problems executing in a sandbox.  In addition, it is necessary to add even more of a _Grace Period_ for interactive problems in a sandbox, since the wall-time (which is what PC2 uses for timing runs) includes the time consumed by any interactive validator, as well as the sandbox setup overhead.
...

**Feature Description**: 
Add multiplier to be applied to the specified problem time-limit for the time out values for execution timers.  In addition, use the already existing **Extra Kill time**  value only for runs that use a sandbox, but are not interactive.  Both the ExecuteTimer and the TLE timer have to be adjusted to use these new _padded_ time limits.  Note that this will NOT affect TLE's since the sandbox handles CPU time TLE's.  These timers are a failsafe for runs that do not consume CPU time, but otherwise take a long time to run, eg. a contestant put a `sleep `in their submission, or inefficient I/O is spending lots of wall time but little CPU time.

The values for the multiplier and the normal sandbox _Grace Period_ should be configurable.
...

Manual review of each judgment that receives a TLE would be required.
...

This is a low level of effort fix.  Just involves increasing the times set for the timers.  
...

**Additional context**: 
When PC2 is acting as a shadow, the primary, DOMjudge in this case, allows roughly 3x the time limit for interactive problems to complete.  The CPU time is still the specified time limit, but the program is permitted to use 3x as much wall time.  In order to avoid mismatches during shadowing, PC2 should do something similar.  In addition, if PC2 is to act as primary, then these changes will be necessary as well so that the shadow will not run into the same problem.  (Both the primary and shadow have to _be on the ssame page_ @johnbrvc has already coded these changes.  PR to follow shortly.
...

johnbrvc:

1. Add 2 new values to **ContestInformation**: `sandboxGraceTimeSecs` and `sandboxInteractiveGraceMultiplier`.  These integer values may be configured using the `system.pc2.yaml` at contest load time, or the GUI **Settings** tab after a contest is configured.  Default values are: 1 and 3 respectively.
2. Change **Executable** to pad execution timers to wait additional time (grace time) for sandbox and interactive sandbox problems, where the sandbox will monitor the actual CPU time of the submission.  The timers in PC2, in this case, are for **Wall** time to protect against sleeping/wall time consuming processes.
3. Cleaned up some comments that were wrong or had errors in them
4. Rearranged some code to initialize the timer delays earlier on since they are used in 2 places.
5. CI: While testing the new sandbox time grace periods, a couple of bugs were found in the **SubmitSampleRunsPane** (which is used to submit the judge's sample solutions by an administrator.    The following were fixed/added to support **SubmitSampleRuns**:

    - Add alphanumeric sorting instead of pure numeric sorting for some columns since some columns may have both text and numbers in them.
    - Synchronize the add runs since we appear to get 2 runAdded notifications for each run from different worker threads.
    - Fixed bug in Filter where if a custom filter group is used alone, it would not allow you to enable filtering.

Basically, these new configuration parameters replace constants that were used in Executable previously.

This PR is one of those cases where there appear to be lots of changes, but again, Eclipse has made many of them related to spaces or other syntax fixes in `ContestInformation.java`.  Also, there were new modules added and one renamed, so they appear as a lot of differences as well.

Fixes #999 

Windows 11/Ubuntu 22.04.1 (tested with both)
java version "1.8.0_321"

1) Choose a CDP for testing
2) Edit the **system.pc2.yaml** (or **system.yaml**, if there is no **system.pc2.yaml**) and add the following properties:
```
     sandbox-grace-time-secs: 2
     sandbox-interactive-time-multiplier: 4

```
3) Load the contest
4) Start the pc2server an administrator client.
5) Go to the "**Configure Contest->Settings**" tab.
6) Notice that the values for the **Sandbox Grace Time added** and **Interactive problem time multiplier** reflect the values added to the system **yaml** file:
image

7) Using the GUI, change the values to 1 and 3 respectively.  (You can click the ? (what's this) buttons next to each value for a detailed description of what they do).
8) Shutdown the admin and the server
9) Restart the server and admin
10) Go to the "**Configure Contest->Settings**" tab and notice the values that were modified by the GUI have been saved.
image


SamanwaySadhu (approved):
Works as described. Code changes look good.

kkarakas (approved):
Code looks good. Did not test functionality.

Issue #79, opened 2020-05-09
WTI does not display pending runs or clars unless refreshed


clevengr:
**Describe the issue**: 
The Web Team Interface (WTI) project does not display "Pending Runs" in the "Runs" grid immediately after they've been submitted; it requires switching to another screen and then back to "Runs" to see a submitted run which hasn't yet been judged.  Similarly, a newly-submitted clar does not immediately show up; it requires switching to another screen (and back) to see that the clar is "pending" (not yet answerered).

**To Reproduce**:
- Start a contest (start a PC2 Server configured with a contest, then start a PC2 Admin and use it to start the contest)
- Do NOT start an AutoJudge
- Start the WTI server (see the instructions at [this page](https://github.com/pc2ccs/pc2v9/wiki/Web-Access))
- Use a browser to connect to the WTI server and login to the PC2 system (see the above link for help).
-Go to the "Runs" page and submit a run

**Expected behavior**: 
The run should show on the Runs grid as "Pending" (or at least, as "not having a judgement").

**Actual behavior**: 
The run does not appear on the Runs grid at all.
However:  click on the "Clarifications" tab, then click back to the "Runs" tab.  The run now appears on the Runs grid.  So the Runs grid is just not being updated when the run is first submitted.

Repeat the above for a clar (submit a clarification request); observe that it does not show up as expected in the Clars grid but WILL show up if you click on the "Runs" tab and then back to the "Clars" tab.  Same issue: the Clars grid is not being refreshed when the clar is submitted.

**Environment**:  This was  verified present using:
Version 9.7build 20200427 (Monday, April 27th 2020 01:53 UTC) build 5582
Java version 1.8.0_201
OS: Windows 10 10.0 (amd64)
 Chrome Version 81.0.4044.129 (Official Build) (64-bit)

PR #1021, opened 2024-12-11
i871: F5 in WTI causes logout

### LINKED ISSUE #871: F5 in WTI should not revert to login page
**Describe the issue**: 

When a team logs into the WTI (UI), they typically navigate around the several different pages comprising the SPA (Angular Single-Page-Application).  A common thing for a team to do is to wonder if a given page (say, the list of clarifications) is "up to date".  The common way of checking this (in almost any browser situation) is to hit F5, to "refresh" the page.  When F5 is hit on **ANY** WTI page, it doesn't "refresh" the page -- it immediately transfers to the login page, requiring the team to login again.  Hitting F5 should simply update the CURRENT page; it should not transfer to the login page.

**To Reproduce**:

- Start a server using --load to load a "sample contest" ("sumithello" will do nicely).
- Start an Admin and verify that there exists a "scoreboard2" account and at least one team account in the contest (if not, create them).
- Start the WTI server (unzip the zip file in the WTI project, then run the command "./bin/pc2wti").
- Open a browser to the URL `http://localhost:8080`.  This should present the WTI's "Balloon Login" screen.
- Login as a team.  This should default to presenting the "Runs" (Submisisons) screen.
- Hit the F5 key.

**Expected behavior**: 

The WTI should "refresh" the current (Submissions) page.

**Actual behavior**: 

The WTI switches to the Login screen and forces you to login again.

**Environment**: 
 - Windows 10, Java Version 119.0.6045.160 (Official Build) (64-bit)
 - PC2 Version 9.9build-6772~develop
 - Chrome Version 119.0.6045.160 (Official Build) (64-bit)

**Additional context**: 

The root cause here is almost certainly a failure to set up Angular _routing_ properly.  By default, an F5 on a component with no specified router route is probably going back to the root page (which for WTI is the login page).  What likely needs to happen is that a new "route" needs to be defined for each component (page) mapping F5 for that page back to that page.

clevengr:
Fixes the problem where an F5 (browser refresh) in the WTI-UI Angular browser TypeScript code causes a switch to the login page.  Now, hitting F5 keeps the user on the same page as before.

Note that _behind the scenes_ there is a WHOLE lot going on to make this apparently simple thing happen.  Generally, the following describes what the PR does:
- Every page in the Angular Single-Page-Application (SPA) now writes an entry into browser `sessionStorage` indicating that page is the "current page".  Transferring to a different page (e.g. from "Runs" to "Scoreboard") causes this `sessionStorage` value to be updated appropriately.
- Upon successful login, and before transferring to the "main screen" (which is in fact the "Runs" page), the login page writes two additional values into `sessionStorage`:  the logged-in user name (e.g. "team3") and the "token" which was returned from the WTI-API (a.k.a WTI Server) as part of the login-success operation.  These two values are normally kept in the `AuthService` class and are required to be included in the headers of any subsequent HTTP calls between the browser and the WTI Server.
- When a browser refresh (F5) occurs, the browser reloads _the entire SPA_ from the server, and re-executes the SPA initialization code.  (This is the way browser refresh works, and it cannot [reasonably] be changed.)
- The SPA initialization code has been modified to check for the existence of the above data in `sessionStorage`.  If not found, initialization proceeds as in the past.  However, if the above data IS found in `sessionStorage`, the initialization code reads the `sessionStorage` data and re-inserts it into the appropriate place(s) in the SPA objects.  For example, it restores the user name and token into the `AuthService` object.
- The initialization code then uses the Angular `Router` to force a navigation to the "current page" as listed in `sessionStorage`.

The overall effect of the above is that an F5 causes the SPA to be (re-)initialized as if the current user had just logged in, and transfers the user back to the page they were previously viewing.

Fixes #871 

Windows 10 [laptop] and 11 [desktop] ; 
java version "11.0.16.1" 2022-08-18 LTS ;
Eclipse Versions 2020-12 (4.18.0)  [laptop) and 2021-12 (4.22.0) [desktop]
Chrome Versions 130.0.6723.117 and 131.0.6778.109 

#### Setup:
- Start a PC2 server, loading the "sumithello" contest.  Do NOT start the contest clock yet.
- Start an Admin; verify that there exists a "scoreboard2" account (or create one if not).
- Start an AJ configured to auto-judge both "sumit" and "hello" (the "sumithello" contest predefines both judge3 and judge4 to do this).
- Start a manual judge for responding to clarifications (or, use the Admin for this).
- Go into the **projects/** folder and unzip the WebTeamInterface zip file.
- Go into the unzipped WebTeamInterface folder, start a command prompt, and run **./bin/pc2wti** to start the WTI Server (aka WTI-API).

#### Testing, Phase 1 (Verify that all previous functions still work correctly):
- Open a new browser window
  - Optionally, hit F12 to open the browser console so that you can see Angular debug/tracing messages.
- Enter http://localhost:8080 in the browser address bar; you should get the PC2 login page
- Login as one of the teams defined in the contest configuration (we'll assume you use "team1/team1").  This should display the WTI "Runs" page.
- Click the Submit Problem button on the Runs screen, then click the "Select Problem" dropdown.
    This should NOT show a list of the problems, because the contest is not started (see above).
- On either the Admin or the PC2 Server, select the "Times" tab and click the "Start All" button.
- On the PC2 Team browser, again click the "Select Problem" dropdown.  It should now show the contest problems ("Hello" and "Sumit").
- Cancel the Submit Problem operation, then click the "Logout" menu item on the Team browser screen.  This should return you to the PC2 Team Login screen.
- Log back into PC2, using the same account as before (team1/team1).
- Submit a correct run for one of the problems (hello or sumit).  This should return a "Yes" judgement from the AJ. You should also see a browser console message saying "marking standings out of date" (this is because the Yes run makes the scoreboard copy which is cached in the WTI out of date).
    - Note: if you choose the "sumit" problem and attempt to submit a "correct" run chosen from the PC2 samples, you may have to use the Admin to edit the problem and change the "Team Reads From" entry from "File" to "Stdin".  This is because the PC2 samples are apparently configured to default to "File I/O".
- Click the "Scoreboard" menu item on the WTI screen.  This should show the Yes run on the scoreboard, and you should see a browser console message indicating it fetched new (updated) scoreboard standings.
- Click "Clarifications" in the browser menu. This should display the "Clarifications" page.
- Click "New Clarification" and then submit a clarification for one of the problems.  
  - This should cause a new clar to appear in the manual judge (and/or the Admin). Note however that the clarification does NOT immediately display in the browser clarifications list (this is a known issue; see https://github.com/pc2ccs/pc2v9/issues/79).
- Click on another menu item (e.g. "Runs"), then click back on the Clarifications menu; verify that the submitted clarification DOES now appear in the clar list.
- Submit a SECOND clarification from team 1, following the same procedure as above.
- On the manual judge or the Admin, answer the first clar, but do NOT check "Send to All Teams".  This should display the clarification *answer* on the Clarifications screen in the browser.
- Again on the manual judge or Admin, answer the second clar, but this time DO check the "Send to All Teams" checkbox.  This should display the second clar *answer* on the WTI UI.
- Click the "Logout" menu item.  This should return to the "PC2 Team Login" screen.
- Login as a DIFFERENT team (we'll assume "team2").  This should put you at the "Runs" screen for team2.
- Click the "Scoreboard" menu.  This should show the correct scoreboard (a single "yes" for team 1).
- Click the "Clarifications" menu.  This should show the clarification which was "Sent to All Teams", but should NOT show the clarification which was only sent to the first logged-in team.

If all of the above works correctly it demonstrates that no previously-working functionality has been broken.

#### Testing, Phase 2 (test the real purpose of the PR: to make sure hitting F5 doesn't log the team out):

- While still logged in as team2, hit the "F5" key (or click the browser's "reload" button).  This should return you to the same UI page you were on (whereas prior to this PR such an action sent you to the Login screen).  
  - If you are watching the browser console, it should include a message like "Restarting Single-Page-Application after refresh navigation" (along with a lot of other tracing information).
- Select the "Runs" menu, then click "Submit Problem" and submit a solution for a problem; verify that a proper response is received from the PC2 Server.
- Select the "Clarifications" menu, then click "New Clarification"; verify that submitting a clarification works.
- On the Admin or manual Judge, answer the new clarification but only send the answer to team2; verify that the answer is received by team 2.
- Select the "Scoreboard" menu; verify that the correct scoreboard is displayed.
- Click "Logout"; this should return you to the PC2 Login screen.
- Login as team1.
- Click "Clarifications" in the header menu.  Verify that team1 can see the clar it originally sent, and the clar from team2 which was sent to "All Teams", but CANNOT see the most recent clar sent just to team2.
- Hit F5 (or the browser "Refresh" button); verify you remain on the same page (clarifications) with the same information.
- Click on the "Runs" menu, then submit a correct run for the other problem (whichever you didn't submit above); verify that you get back a correct response from the Judge.
- Click on the "Scoreboard" menu; verify that the scoreboard is correct.

If all of the above works, it should be sufficient demonstration that the PR correctly solves the problem of F5 logging the user out -- and that hitting F5 still allows all other functions to happen normally.


projects/WTI-UI/src/constants.ts
```diff
@@ -0,0 +1,43 @@
+//App-wide constants for the PC2 WTI-UI project
+
+/**
+ * Global flag indicating whether to display debug messages on the browser console.
+ * Set to null to disable debugging output; any non-null value causes debugging output.
+ * NOTE: this flag should probably be set to null when creating a production release.
+ */
+export let DEBUG_MODE = true ;
```
johnbrvc, line 8: Do we want DEBUG_MODE = true to be the default?

johnbrvc:
Once logged into a team, if a change is made on the "**Options**" page (eg. changing the state of the "Pop-ups" for Clars or Runs), this is not preserved if F5 is pressed.

This also happens on the "**Runs**" page and "**Clarifications**" page with filters.  If you have filters set, their state is lost.  However, on these pages, if you simply switch pages, the filter state is lost as well, so evidently that was the intent of the original code.  The "**Options**" page, however, should be preserved.

johnbrvc (changes requested):
The one comment about "Options" is the only sticking point.
I tried building with DEBUG set to false to make sure everything still worked, and it does appear to.
Reviewing the code consisted of me simply reading through it, as I am not a typescript person.  I was just looking for glaring issues - I didn't see any.  There are a TON of conditional DEBUG console logging messages, which I completely understand and see the need for.

projects/WTI-UI/src/constants.ts
clevengr: I would say we definitely want DEBUG_MODE to be FALSE in our releases.  I will make that change in my next push.

clevengr:
> Once logged into a team, if a change is made on the "**Options**" page (eg. changing the state of the "Pop-ups" for Clars or Runs), this is not preserved if F5 is pressed.
> 
> This also happens on the "**Runs**" page and "**Clarifications**" page with filters. If you have filters set, their state is lost. However, on these pages, if you simply switch pages, the filter state is lost as well, so evidently that was the intent of the original code. The "**Options**" page, however, should be preserved.

Excellent catch regarding "Options"; I totally overlooked that as being part of the "state" which needs to be saved/restored.  I'll push an update to the PR with that fixed.

Regarding the issue of filters losing their state, two comments.  One, I think we should fix that; it's very irritating.  Two, I think this should be filed as a separate Issue since it was a pre-existing condition and doesn't really directly relate to the "F5" issue (although, it will mean whoever tackles such a new issue will need to include save/restoring filter states on F5 refresh).  Also, I think we need to consider a complete redesign of the whole "filtering" issue (and it's layout), so we should probably do that in conjuction with -- or perhaps before -- tackling the "save filter state" issue.

clevengr:
I pushed five new commits that collectively solve (are intended to solve) the issue of Options Page values being lost on an F5 reload.

johnbrvc (approved):
Everything seems to work.
As I indicated, my knowledge of typescript is lacking, but it looks good.

projects/WTI-UI/src/constants.ts
kkarakas, line 1: Should constants be in .env vars or at least .json? Something to thing about. If it is in .env and if we include .env in gitignore there is no way DEBUG_MODE will be accidentally true  in prod.

projects/WTI-UI/src/app/modules/core/abstract-services/i-teams.service.ts
```diff
@@ -1,12 +1,23 @@
...
+import { DEBUG_MODE } from 'src/constants';
 
+@Injectable({
+	providedIn: 'root'   //forces the service to be a singleton across all app components ('root' == "root injector")
+})
 export abstract class ITeamsService {
   runsUpdated = new Subject<void>();
+  
+  constructor () {
+	  if (DEBUG_MODE) {
```
kkarakas, line 17: There are a lot of cases where DEBUG_MODE is checked for logging purposes. Could we maybe be create a logging class that when initialized checks the DEBUG_MODE. Having if cases all around makes it harder to read. Logs should be for example used as log.debug("blah blah"). Alternatively we can use winston.

kkarakas (approved):
Built and tested the program. It works well. Did some code review and nothing problematic came up to me.

projects/WTI-UI/src/app/modules/core/abstract-services/i-teams.service.ts
kkarakas: Reading some forums it is probably a bad idea to use winston on client side.

clevengr:
@kkarakas : I was under the impression that you had approved this PR (your suggestions about changing the way DEBUG_MODE is handled notwithstanding).  In fact, there is a comment in the discussion thread saying
```
[kkarakas](https://github.com/kkarakas) approved these changes [2 weeks ago](https://github.com/pc2ccs/pc2v9/pull/1021#pullrequestreview-2524869296)
```
However, under the "Changes Approved" list, it only shows a single approval (JohnB's).  Did you actually submit an "Approve" review for this PR?  If it was indeed your intention to approve it, could you go back and submit a review with "Approve" marked so that we can merge the PR?

kkarakas (approved):
I swear I approved it and I have a tick mark next to my photo on top right but I am repeating it.

Issue #1029, opened 2025-01-02
Allow WTI viewing of Clars (always) and Problems (any time after contest has been started)


clevengr:
**Describe the issue**: 

Currently the WTI code (specifically, the `ContestController` class in the WTI-API `controllers` package) supports HTTP endpoints `/clarifications` (which returns a list of current clarifications) and `/problems` (which returns a list of the current contest problems).  Both these endpoints have code at the beginning which requires two conditions to be met:  the requesting user must be logged in, and the contest must be "running" (i.e., has been _started_ and is not currently _stopped_ (a.k.a. "paused").  If either of these conditions is false, the endpoint returns "401 - Not Authorized".

There doesn't seem to be any logical reason for requiring that the contest be "running" in order for a user to see a list of clarifications -- for example, a contest admin might want to send a "clarification" (announcement) like "The start of the contest has been delayed; it will start at 9:20".  The requirement of the `/clarifications` endpoint that the contest be "running" means that logged in users won't be able to see this if it is sent before the contest starts.

Similarly, there doesn't seem to be any reason to prohibit viewing of the contest problems _after the contest has been started_  (for example, during a "pause"); once the contest has been started then teams have already seen (or at least, had the opportunity to see) the contest problems -- but the current `/problems` endpoint implementation won't allow them to continue to see them if the contest is paused.  

**To Reproduce**: 
- Start a PC2 Server loading a contest containing at least one problem; make sure the contest.yaml config does NOT indicate that the contest should be "running" (i.e., make sure that the contest has never been started).
- Start a WTI server.
- Start a PC2 Admin; send a clarification announcement "To All Teams".
- Open a browser and login to the WTI.
- Select the "Clarifications" menu item in the browser.
- On the Admin, "start" the contest, wait a moment, then "stop" (pause) the contest.
- On the team, select the "Runs" tab and then click the "Problems" dropdown.

**Expected behavior**: 

- The logged in team should be able to see the clarification announcement before the contest has started.
- The logged in team should be able to see the problem list once the contest has been started and then paused.

**Actual behavior**: 

The browser console will show multiple Errors ("401 - Not Authorized") from the "Clarifications" request;  the team will not be able to see the problem list when the contest is paused (after it has been started).

**Environment**: 

All.

**Additional context**:

Two suggested steps to fix this issue:
[...]

PR #1034, opened 2025-01-12
I1027 Add Contest Clock To WTI

### LINKED ISSUE #1027: Add a contest clock to WTI
**Feature Description**: 
The WTI screen should display a "contest clock" at all times.  It might be useful to have both "Elapsed:" and "Remaining:" displays.

It's probably going to involve changes to the **app-header** component in the **shared** module under the WTI **src** folder.

**Additional context**: please provide any additional information that will help us in considering your feature request.  

Angular already defines a library containing a countdown widget; this should probably be used.  See https://blog.devgenius.io/how-to-show-a-countdown-timer-in-angular-using-ngx-countdown-9835c98de101 for details.

clevengr:

Adds "Elapsed Time" and "Remaining Time" counters to the WTI header bar whenever a team is logged in.  

The counters are initialized with values read from the PC2 Server (via the WTI-API Server and its PC2 API connection) each time a team logs in.  The counters respond appropriately to changes in the "PC2 contest clock" state -- that is, when the Admin "Starts" the contest the counters begin counting (up/down respectively), and the counters stop (freeze) whenever the Admin "Stops" (pauses) the contest.  The counters also respond correctly to changes in the contest times; for example, if the Admin extends the contest (adds more time), then the "Remaining Time" counter automatically adjusts as appropriate.

Times are displayed in HH:MM:SS format, preceded by a number of days if the total time is greater than 24 hours.

The counters are driven by a JavaScript _setInterval()_ object which ticks at a rate of once per second.  To handle potential clock drift, the code automatically "resynchronizes" the on-screen counters with the current PC2 Server time, at a frequency defined by a constant in the WTI-UI `constants.ts` file (currently every 15 minutes).

A screenshot showing the clocks for a contest that was scheduled to run for 30 hours and has been running for just over an hour is shown here:

image

### Issue which the PR addresses:

Fixes #1027, #1029

- Windows 10 & 11
- Eclipse Versions 2021-12 (4.22.0) and 2020-12 (4.18.0)
- Java versions 1.8.0_381 and 11.0.16.1
- Chrome Version 131.0.6778.109
- MS Edge Version 131.0.2903.112 

### Precise steps for _testing_ the PR :

- Edit samps/contests/sumithello/config/contest.yaml:  change from running:true to running:false,then save the updated file.
- Start a PC2 server using --load sumithello.
- Start a PC2 Admin; verify that the contest is NOT running and has NEVER BEEN STARTED (check the "Times" tab"; under "Started?" it should say "No").
- In the PC2 "projects" folder, unzip the WebTeamInterface .zip file.
- Go into the resulting WebTeamInterface folder, open a terminal command prompt, and start the WTI server with the command `./bin/pc2wti`.
- Open a browser, then hit F12 (or the browser's equivalent) to open the browser devtools; select the "Console" tab.
- Enter "http://localhost:8080" in the browser address bar.  Verify:
   - you are presented with the PC2 Login screen, and 
  - no ERRORS appear on the browser console (there will be lots of debug/tracing messages, but none should say "Error").
- Login as "team1/team1".  Verify:
   - you are presented with the PC2 "Runs" screen
   - the Header displays a field "Elapsed Time: 00:00:00 (no contest time has elapsed because the contest hasn't been started)
   - the Header displays a field "Remaining Time: xx:xx:xx", where xx:xx:xx is the number of hours/mins/secs defined as the
      length of the contest (for the default values in the "sumithello" contest, that should be "05:00:00")
      Note: if you change the Contest Length in the contest.yaml file BEFORE STARTING THE SERVER, and the value you set for
        contest length is greater than 24 hours, the "Remaining Time" field should also include a count of the number of DAYS remaining.
   - neither the Elapsed Time nor the Remaining Time values are changing (because the contest isn't running).
   - there are no ERRORS on the browser console (again, there will be lots of tracing/debug messages, but there should be none marked ERROR)
- In the browser, click on the Clarifications menu item; verify there are no ERROR messages in the browser console.
- Click on the Runs menu item, then click on the "Select Problem" dropdown in the "Filters" pane.  Verify that the contest problems are NOT shown.
- On the PC2 Admin, START the contest, then verify:
   - the Elapsed Time field on the browser header starts counting up at a rate of once per second;
   - the Remaining Time field on the browser header starts counting down at a rate of once per second;
   - clicking on the "Select Problem" dropdown in the "Filters" pane now shows the contest problems (Sumit and Hello).
   - there are no ERROR messages on the browser console.
- On the PC2 Admin, STOP ("pause") the contest, then verify:
   - both clock fields on the browser freeze;
   - the sum of the Elapsed time plus the Remaining Time on the browser header equals the "Contest Length" shown on the Admin "Length" field.
   - clicking on the "Select Problem" dropdown in the "Filters" pane still shows the contest problems.
   - there are no ERROR messages on the browser console.
- Note the "frozen" times in the browser window header, then on the PC2 Admin, START ("un-pause") the contest, then verify:
   - both clock fields start changing again, with Elapsed counting up and Remaining counting down, from the times they were "frozen" at when the contest was stopped.
   - there are no ERROR messages on the browser console.
- On the PC2 Admin, once again STOP the contest, then note the "Elapsed" and "Remaining" times which are now frozen on the browser header.
- Hit F5 in the browser window; verify that 
   - the Elapsed and Remaining Time clock values still have their same (frozen) values.
   - Clicking the "Select Problem" dropdown in the "Filters" pane still shows the contest problems.
- On the PC2 Admin, START the contest again; verify that 
   - the Elapsed and Remaining Time clocks resume counting correctly from their frozen values.
    - Clicking the "Select Problem" dropdown in the "Filters" pane still shows the contest problems.
- On the browser window, hit the "Logout" menu item; verify that you are returned to the PC2 Login screen.
- Login as another team; verify that the Elapsed and Remaining Time counters start at the expected values and are correctly changing  (the current "Elapsed" and "Remaining" times can be seen by hitting the "Refresh" button on the Admin "Times" tab).
- On the Admin, STOP the contest, then EXTEND the length of the contest by some amount (on the Admin "Times" screen, select Site1, click "Edit",   add some time to "Contest Length", and also add a corresponding amount of time to "Current Remaining Time".) Click "Update", then verify that  the "Remaining Time" on the team browser is properly updated.
- On the Admin, START the contest, then verify that the browser times begin properly changing again (and that "Remaining Time" correctly shows the updated value).
- MINIMIZE the team browser, wait a minute, then restore the browser.  Watch the "Elapsed Time" field as the browser is restored; it will start by displaying the elapsed time as it was AT THE MOMENT IT WAS MINIMIZED but should then (almost immediately) be updated to the correct elapsed time (which can be determined by hitting the "Refresh" button on the Admin "Times" window).  (This is happening because JavaScript timers don't run at the correct rate when the browser is minimized; 
a separate issue will be filed for this.)


projects/WTI-UI/src/app/modules/core/services/contest.service.ts
```diff
@@ -74,5 +117,100 @@ export class ContestService extends IContestService {
...
+			);	
+	}
+	
+	/** This method receives a WTI-UI ContestClock model containing new values which should be used to update the WTI-UI contest clock,
+	 *  including the onscreen displays.  It constructs a new ContestClock object containing the received data and installs that
+	 *  object as the current WTI-UI clock.  It then updates the separate "ContestTimer" object with the specified values, and
+	 *  if the received data indicates the clock should be running it starts the ContestTimer (which then genrates a "clock tick"
+	 *  once per second to update the clock displays).
+	 */
+	installNewContestClock(data: any) {
```
kkarakas: Using any type, unless there is a good reason (such as I/O), is frowned upon.

kkarakas: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#:~:text=%E2%9D%8C%20Don't%20use%20any,every%20usage%20of%20the%20variable.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
```diff
@@ -74,5 +117,100 @@ export class ContestService extends IContestService {
...
+			}
+			this.contestTimer.startTimer();
+		}
+	}
+	
+	/** Returns the number of seconds which have elapsed so far in the contest, which it obtains from the separate
+	 *  ContestTimer object.
+	 */
+	getElapsedSecs(): number {
+		let elapsedSecs = this.contestTimer.getElapsedSecs();
```
kkarakas: Is there a reason this is a "let", it can be const. Or it can be returned directly.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
```diff
@@ -74,5 +117,100 @@ export class ContestService extends IContestService {
...
+	getElapsedSecs(): number {
+		let elapsedSecs = this.contestTimer.getElapsedSecs();	
+		return elapsedSecs ;
+	}
+	
+	/** Returns the number of seconds remaining in the contest, which it obtains from the separate
+	 *  ContestTimer object.
+	 */
+	getRemainingSecs(): number {
+		let remainingSecs = this.contestTimer.getRemainingSecs();
```
kkarakas: Is there a reason this is a "let", it can be const. Or it can be returned directly.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
```diff
@@ -74,5 +117,100 @@ export class ContestService extends IContestService {
...
+		newContestClock.wallClockStartTime = data.wallClockStartTime ;
 
+		//save the new clock
+		if (DEBUG_MODE) {
+			console.log("ContestService (id", this.uniqueId, ").installNewContestClock(): replacing Contest Clock with:");
+			console.log(newContestClock);
+		}
+		this.contestClock = newContestClock;
+		
+		//pull the values out of the updated clock
```
kkarakas, line 149: Similar comment on usage of "let"

projects/WTI-UI/src/app/modules/login/components/login-page/login-page.component.ts
```diff
@@ -124,17 +129,49 @@ export class LoginPageComponent implements OnInit, OnDestroy {
...
+			console.log("  Invoking ContestService.getContestClock(), subscribing for callback result")
+		}
+		this._contestService.getContestClock() 
+			.subscribe(
+				(data: any) => {
+        			if (!data) { 
+						console.error ("LoginPageComponent.onSubmit() ContestClock subscription callback: unable to get ContestClock from PC2 API via ContestService!");
+					} else {						
+						//copy the data fields received from the PC2 Server (via the WTI-API) into the local ContestClock object
+						this.contestClock.isRunning = data.running ;
```
kkarakas: There is a lot of nested conditionals here, maybe we can use break to break out of conditions. This would improve readability.

projects/WTI-UI/src/app/modules/shared/components/problem-selector/problem-selector.component.ts
```diff
@@ -2,97 +2,162 @@ import { Component, forwardRef, OnInit, OnDestroy, Input } from '@angular/core';
...
-    }
-  }
+	/**
+	 * This method determines whether the contest problems should be displayed in a "ProblemSelectorComponent".
+	 * Problems are only allowed to be displayed when the contest has been started, which is determined by looking
+	 * at the "wallClockStartTime" value in the current ContestClock.  The method fetches the current clock from the WTI-API server,
+	 * and if wallClockStartTime is greater than zero it fetches the current problem list from the server and displays it in the
+	 * component; otherwise it sets the list of displayed problems to an empty list (array).
+	 */
+	private loadProblems(): void {
```
kkarakas, line 89: woah. Is it possible to simplify this method through having multiple methods breaks etc.

kkarakas (changes requested):
Did not run it on my machine yet, but these are my opinions. Looks like it would work well!

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Agreed; I'll fix it.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: I agree it should be const; I'll change it.  I prefer not to return expressions, for the reason that if the contents of the function needs to be modified/expanded later on, it could require making changes to the returned expression as well.  If someone misses this change while modifying the function it could lead to returning incorrect results. (I grant that since this is a simple one-line function it's not likely, but as a general principle I think it's safer to compute the return value within the function and then return the variable holding that return value.   Certainly, other people might disagree.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Same reply as the previous comment; I'll change it to `const`.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: And similar reply; I agree and I'll change them all to `const`.

projects/WTI-UI/src/app/modules/login/components/login-page/login-page.component.ts
clevengr: I don't quite follow this comment.  As far as I can tell there are no nested conditionals which would allow using `break`.  What there ARE (or at least, WERE) are a lot of "debug if's".  I will remove all those debug statements; hopefully the logic will be clearer.

projects/WTI-UI/src/app/modules/shared/components/problem-selector/problem-selector.component.ts
clevengr: As with the previous comment, I don't see any way to use `break`s to address this.  As before, I think the problem was the large number of "debug" `if`s in the code, making it hard to see the structure.  I've eliminated all those debug statements; hopefully that makes the code more readable.  I'll welcome any suggestion for how to use `break`s to simplify it further.

johnbrvc: I think the in-line lambda function is what clutters it up.  Folks love doing that in javascript too, and it makes it very hard to follow and read.  I guess it's the nature of the beast.

projects/WTI-UI/src/app/modules/login/components/login-page/login-page.component.ts
kkarakas: I meant return sorry. Maybe the method can be terminated early if it is not going to lead anywhere?

projects/WTI-UI/src/app/modules/shared/components/problem-selector/problem-selector.component.ts
kkarakas: I meant return sorry. Maybe the method can be terminated early if it is not going to lead anywhere?

projects/WTI-UI/src/app/modules/core/services/contestTimer.service.ts
```diff
@@ -0,0 +1,139 @@
...
+   * The method uses the JavaScript "setInterval()" method to detect passage of each second (1000 msec).
+   * Note that setInterval() is not guaranteed to continue operating at the correct rate if the browser is minimized;
+   * the method compensates for this by saving the current time at each update and then checking to be sure the
+   * current update is happening less than two seconds since the last update; if not, it invokes the 
+   * ContestService updateLocalContestClockFromServer() method to resync the local contest clock with the server.
+   */
+  startTimer() {
+	if (this.intervalId) {
+      console.error("ContestTimerService.startTimer():  call to startTimer() when timer is already running");
+      return;
```
kkarakas, line 85: Tabbing doesn't seem to match on my screen here and other places. Is this intentional?

kkarakas: Or is just on my screen?

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.scss
```diff
@@ -14,6 +14,16 @@ header {
...
+    padding-left: 8px;
+    padding-right: 8px;
+  }
+  
+  /* Set the container holding the contest clock value(s) to medium, white font */
+  .clock-container {
+	    color: white;
+    font-size: 1.0rem;
+    padding-left: 8px;
+    padding-right: 8px;
```
kkarakas, line 26: I have some concerns about how these pixel paddings would look in different window sizes. Did you play around with different window size and different pixel counts? Perhaps we could use relative units such as em/ rem, vw, just like you use it for font-size. But this is not necessarily problematic, there are cases where pixels are better.

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.ts
```diff
@@ -14,13 +25,31 @@ export class AppHeaderComponent {
...
   
+  //Return a boolean indicating whether or not to show the contest clock(s) in the header
+  get showClocks(): boolean { return this._authService.isLoggedIn; }
+  
   /* Return a string containing the "team id" -- that is, the PC2 team account number with
      the leading "team" removed */
   get teamId(): string { 
     let acctId = this._authService.username; 
     let teamId = acctId.substr(4);
     return teamId;
```
kkarakas, line 36: This could be changed to const while we are at it.

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.ts
```diff
@@ -14,13 +25,31 @@ export class AppHeaderComponent {
...
+	  if (DEBUG_MODE ) {
+		console.log ("Executing AppHeaderComponent constructor")
+	  }
+  }
+
 
-  constructor(private _authService: AuthService) { }
+  getElapsedSecs(): number {
+	let secs = this._contestService.getElapsedSecs() ;
+    return secs;
```
kkarakas: :)

kkarakas (changes requested):
More opinions!

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Fixed in next push.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Fixed in next push.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Fixed in next push.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Fixed in next push.

projects/WTI-UI/src/app/modules/login/components/login-page/login-page.component.ts
clevengr: I've removed the large number of "debug" statements, which makes the nested conditionals pretty easy to follow (as JohnB has noted elsewhere, the traditional JavaScript usage of "lamba functions" does complicate it a little bit).   As it now stands I don't really see where there's any place that the addition of "return" statements would be particularly helpful...

Let me know if you have a specific suggestion (short of rewriting the whole thing avoiding lambda expressions :) for improving it further.  If not, please mark the conversation "Resolved".

projects/WTI-UI/src/app/modules/shared/components/problem-selector/problem-selector.component.ts
clevengr: As noted, I've removed the large number of "debug" statements; hopefully this makes the code clearer.  Also, as @johnbrvc noted, the use of JavaScript "lambda functions" makes it a little hard to read.  (I didn't write that code originally, and I'm reluctant to change it as part of this PR.  If you think it still needs to be changed, I suggest submitting a separate Issue for cleaning that up...)

Again, let me know if you have a specific suggestion (short of rewriting the whole thing avoiding lambda expressions :) for improving it further. If not, please mark the conversation "Resolved".

projects/WTI-UI/src/app/modules/core/services/contestTimer.service.ts
clevengr: If you are talking about the way indentation appears when you look at the code in Eclipse, that's a function of your Eclipse settings -- both global settings and the setting for the particular editor you're viewing it with.  At one point in time we (the early members of the PC2 project) had agreed upon a common set of indentation settings; perhaps yours don't match?

(If it's NOT in Eclipse that you're talking about, then where is it?)

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.scss
clevengr: I did try it using various window sizes, and I didn't spot any problems.  Do you have a specific example of a platform and window size which shows a problem?

Regarding the "pixel paddings", I didn't create those; they came from a CSS file I got from @johnbrvc .   But I haven't observed any issues with it (I tested it in both Chrome and Edge, using various window sizes).  Again, do you have an example of a specific situation where a problem is demonstrated?

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.ts
clevengr: Agreed; fixed in next push.

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.ts
clevengr: Ageed; fixed in next push.

projects/WTI-UI/src/app/modules/core/services/contestTimer.service.ts
johnbrvc: It appears that there is an actual "TAB" character on the lines in question.  As @clevengr pointed out, if your Eclipse settings are "right" (by "right" I mean that they match ours), it will line up properly.   What IS interesting is, the code indenting appears to violate the indenting rules that we use in the java code.  The particular line in @kkarakas example has a TAB SPACE SPACE.  It seems to me, if TAB is 4 spaces, then the next level of indenting should be a total of 8 spaces (however you want to arrive at that: TAB TAB, 8x SPACE, etc.)  Not sure where the "+2" spaces comes from, unless that's standard for TS?  I attached a screenshot of my general editor properties.
image

clevengr: I'm not 100% positive, but I *think* the "default standard" for TypeScript IS "2 space indents".  I say this only because almost all the TS code we inherited from the students who did the original WTI project was indented by 2 spaces (on each indent).   I know that has caused some aggravation for me...  Not sure how we "fix" this.

johnbrvc: Ok, that's fine, but what they did initially, it seems, is they "know" that TAB is 4 spaces, which is 2 "TS" indents, one keypress.  Ugh.  So, I guess just set your TAB to 4 spaces, and you should be ok.  The problem is, adding new code, you have to be aware of this and increase levels by 2 spaces...

kkarakas (commented):
Code looks good to me. I ran the branch and tested contest clock. Works as advertised. However it does not support  remaining time well if the value is negative. The screenshot attached happens 13 seconds after the contest ends. Perhaps it should be 0 or it should continue showing negative time. 
Screenshot 2025-04-03 015536

johnbrvc (approved):
I am not pretending to be a typescript expert, but I have been through code quite a bit in the past 4 or 5 days and I have a general understanding of how it works, or better yet, why it DIDN'T work but now does.  

I have tested the latest version and it appears to work as designed.
I will leave it up the TS experts to determine if there are any coding issues with the TS.  TBH, I spent most of the time reviewing the generated Javascript (main.js) when testing.

clevengr:
This PR should not be merged yet; I am still working on fixing the issue that @kkarakas pointed out -- negative "Remaining Time" is not handled properly.  I hope to get that fixed today...

johnbrvc (approved):

clevengr:
> Code looks good to me. I ran the branch and tested contest clock. Works as advertised. However it does not support remaining time well if the value is negative. The screenshot attached happens 13 seconds after the contest ends. Perhaps it should be 0 or it should continue showing negative time. Screenshot 2025-04-03 015536

Good catch; fixed in latest push.  I made it so that the "Remaining Time" stops changing when it hits zero, and also that it displays in red when it is zero.  The following screenshot shows a 5-hour contest that has run 17 seconds past the end; the "Elapsed" time is still counting but the "Remaining Time" is frozen at zero, in red.

image

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
```diff
@@ -74,5 +96,90 @@ export class ContestService extends IContestService {
...
+	/** This method receives a WTI-UI ContestClock model containing new values which should be used to update the WTI-UI contest clock,
+	 *  including the onscreen displays.  It constructs a new ContestClock object containing the received data and installs that
+	 *  object as the current WTI-UI clock.  It then updates the separate "ContestTimer" object with the specified values, and
+	 *  if the received data indicates the clock should be running it starts the ContestTimer (which then genrates a "clock tick"
+	 *  once per second to update the clock displays).
+	 */
+	installNewContestClock(data: ContestClock) {
+		
+		//copy the data fields (received from the PC2 Server via the WTI-API) into a new ContestService ContestClock object
+		let newContestClock = new ContestClock();
```
kkarakas: This can be const.

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.scss
kkarakas: Doesn't seem to be an issue.

src/edu/csus/ecs/pc2/api/IContestClock.java
```diff
@@ -1,10 +1,13 @@
 // Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
kkarakas, line 1: copyright.

src/edu/csus/ecs/pc2/api/implementation/ContestTimeImplementation.java
```diff
@@ -1,6 +1,8 @@
 // Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
kkarakas: copyright.

kkarakas (commented):
These are my comments, I'll run the code tonight and see what happens.

projects/WTI-UI/src/app/modules/core/services/contest.service.ts
clevengr: Fixed in next push.

src/edu/csus/ecs/pc2/api/IContestClock.java
clevengr: Fixed in next push.

src/edu/csus/ecs/pc2/api/implementation/ContestTimeImplementation.java
clevengr: Fixed in next push.

projects/WTI-UI/src/app/modules/shared/components/app-header/app-header.component.scss
clevengr: Based on @kkarakas comment below ("Doesn't seem to be an issue.") I'm marking this Resolved.

kkarakas (approved):
LGTM! I ran the program, issues I pointed out have been solved.

PR #1040, opened 2025-02-03
I1032: Configure Contest Identifier

SamanwaySadhu:
### Description of what the PR does
Updates Contest Identifier object of InternalContest with CLICS Contest ID (or name if ID not present) as well as Contest ID object of Profile.

### Issue which the PR addresses
Fixes #1032 

### Environment in which the PR was developed (OS,IDE, Java version, etc.)
Windows 10, Eclipse 2021-12 R, JDK 8u381 (1.8), C++ 13.2.0

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)
- Run the Ant script build.xml in the pc2v9 project.
- Start a PC2 server.
- Start a PC2 Event Feeder (Generate one feeder account if not present).
- Click `View Event Feed`.
- The logs should now have the correct id from `contest.yaml` present

PR #1044, opened 2025-02-10
i_1042 Allow modification of contestID

### LINKED ISSUE #1042: Allow modification and display of Contest ID (Identifier)
There is no (easy/obvious) way to view the current Contest ID.   And there is no way to change the contest ID.
...

**Feature Description**: 

I think we need a way to allow modification and (easy) display of the current contest Id (internally called the Contest Identifier and stored in the Profile).

Right now, to see the contest ID, you can (and I only saw this by looking at the code - must be a Doug special) is to: Ctrl-Shift-Doubleclick the "Active ProfileName" LABEL.  Yes, the label.   The only way for people "not-in-the-know" to see what the contest ID is to look at one of the **Reports**, such as the "**Contest XML**" report.
```
<profiles contestid="Copy-5272909522774502478" currentprofile="true" description="Copy of Sumit" name="Copy" path="profiles\Pcc0fdab8-61a2-487e-965b-328e92f128ff" pc2id="Copy-5272909522774502478"/>
<profiles contestid="SumH" currentprofile="false" description="Default Contest" name="Default" path="profiles\P80bd31b6-55ca-4f9b-8484-f7cc23584201" pc2id="Default-2920137924903412731"/>

```
It would be nice if it showed the contest ID on the **Profiles** tab (First image below) and perhaps the **Settings** tab (Second image below) - without needing a special magic incantation.  In addition, when you "**Clone**" a profile, a new ID is automatically created by the clone process called: "`Copy-<big long unique ID`".  Maybe you should be able to specify the ID of the "new" cloned contest on the clone contest dialog (Third image below).  
...

The only ways to view the current contest ID is to generate a Report such as Contest XML, or, to look at the event feed.
There is no way to change the contest ID.
...

Update GUI tabs and dialog as discussed above and shown below.
...

**Additional context**: 

Image

Image

Image

...

johnbrvc:
Add label and text field to allow easy display and modification of contest id on the Profiles tab.

The issue mentions adding to the Settings tab on the Administrator as well.  I did not do this for 2 reasons:
1) It's silly to have 2 different ways to set the same setting
2) The Contest Id is not part of the ContestInformation settings, which is what the Settings tab is for.  Could we make a custom section for it?  Yes.  But I refer you back to 1).

Fixes #1042
See NOTE: at bottom - this is not ready to be merged into develop until after #1040 

Java Version    : 1.8.0_321

1) Start up any contest.
2) Go to the Administrator->Profiles tab
3) Note the Contest Id is now shown on the page, and you can modify it as well.

Note that is causes a Warning print in the log if you change the contest Id.  The warning says the contest Id has changed in the return packet.  This is true, and is not harmful.  The Packet code in PC2 shows this warning when this condition happens, presumably, in case it is not intentional.

NOTE: If the contest Id is change, no checking is performed on the validity of the Contest Id supplied.  The Id should be validated.  I will update this PR when PR #1040 is merged since there is a routine in there that checks for the validity of a contest Id.

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ProfilesPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ProfilesPane.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.BorderLayout;
@@ -50,10 +50,10 @@ import edu.csus.ecs.pc2.core.util.QuickLoad;
 
 /**
  * Profile administration pane.
- * 
+ *
  * Provides a front end to all profile functions, like rename, change, clone, etc.
- * 
- * 
+ *
+ *
  * @author pc2@ecs.csus.edu
  * @version $Id$
  */
@@ -62,7 +62,7 @@ import edu.csus.ecs.pc2.core.util.QuickLoad;
 public class ProfilesPane extends JPanePlugin {
 
     /**
-     * 
+     *
      */
     private static final long serialVersionUID = 9075523788534575300L;
 
@@ -94,6 +94,10 @@ public class ProfilesPane extends JPanePlugin {
 
     private JLabel profileDescriptionLabel = null;
 
+    private JTextField profileContestIdTextField = null;
+
+    private JLabel profileContestIdLabel = null;
+
     private JPanel topPanel = null;
 
     private MCLB profilesListBox = null;
@@ -105,7 +109,7 @@ public class ProfilesPane extends JPanePlugin {
     private SwitchProfileConfirmFrame switchFrame = null;
 
     private boolean usingExtraColumns = false;
-    
+
     private boolean showHidden = false;
 
     private JCheckBox showHiddenProfilesCheckbox = null;
@@ -115,7 +119,7 @@ public class ProfilesPane extends JPanePlugin {
 
     /**
      * This method initializes
-     * 
+     *
      */
     public ProfilesPane() {
         super();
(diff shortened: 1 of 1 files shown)


clevengr:
I think I disagree with your statement `It's silly to have 2 different ways to set the same setting`.  We have situations like that throughout PC2 -- the Server and the Admin can both control the contest clock, for example.   

In addition, as a user I would want to be able to find the way to change something in the place I thought "most logical".  I am not at all certain that I would think of looking in "Profiles" to change the "ContestId"; I'm pretty sure I would think of that as more of a "Setting" -- which would lead me to look under the "Settings" tab.

I do grant that "ContestId" isn't (currently) part of the "ContestInformation" class -- but I'm not sure that's sufficient justification for not making it available under the "Settings" tab.   Bottom line: I think this is a "user convenience/usability" issue, and I think I come down on the side of having it available in the (to me) most logical place (i.e, in the `Settings` tab).

However, I could live with putting this off to a separate, new, Issue...

clevengr (approved):
Code changes look reasonable, and I've tested this under Windows 11.  Seems good to me.

kkarakas (approved):
LGTM

SamanwaySadhu:
When the contest id is changed, I see this Warning `Warning: Packet contestId does not match for Packet[UPDATE_SETTING] #9 ADMINISTRATOR1 @ site 1 -> SERVER0 @ site 1 local:wf47 remote:wf46`. Otherwise works as described.

SamanwaySadhu:
I'm nitpicking but I feel it'll look better if all three text boxes are of equal length.
image

johnbrvc:
> When the contest id is changed, I see this Warning `Warning: Packet contestId does not match for Packet[UPDATE_SETTING] #9 ADMINISTRATOR1 @ site 1 -> SERVER0 @ site 1 local:wf47 remote:wf46`. Otherwise works as described.

This is to be expected.  `InternalController `checks to make sure that folks are connected to the right contest and prints a warning if the contestId's don't match.  Since we're changing the ID, the update packet will, of course, not have the right contestId until it's updated.  Ref: `InternalController.receiveObject`

johnbrvc:
> I'm nitpicking but I feel it'll look better if all three text boxes are of equal length. image

I agree.  There are many such cases throughout the code.  Doug was a proponent of doing it the way you see (centered).  I prefer the labels to be left justified, and the edit boxes to line up their left sides with each other, and be the same length, where possible.
This is one of those "someday" tasks.

SamanwaySadhu (approved):
LGTM!

PR #1045, opened 2025-02-10
i_1043 Fix GUI so you can see Scoring Properties

### LINKED ISSUE #1043: Judge's Settings Property Sheet not fully visible on ContestInformationPane
**Describe the issue**: 
Can not see or modify some judging properties on `ContestInformationPane `(**Administrator**->**Configuration**->**Settings** tab)

**To Reproduce**: 
View the **Settings** tab on an Administrator login.  Note the **Judge's Settings, Scoring Property** sheet is not completely visible.

**Expected behavior**:
The **Scoring Properties** sheet should be visible and all properties editable.

**Actual behavior**:
Can not see or edit several of the properties. 

**Environment**: 
Windows, and probably Linux as well.  This is a GUI display issue.

**Log Info**:  

**Screenshots**:

Image

**Additional context**:

johnbrvc:
Change the layout of the Judge's Settings to be `BoxLayout` to better control placement of the individual judge's panes. Adjust size of `ScoringPropertySheet` so it fits.

Fixes #1043 

Window 11
Linux Ubuntu 22.04
java version "1.8.0_321"

1) Build the distribution
2) Start the **pc2server**
3) Start the **pc2admin**
4) Look at the **Settings** tab, and then the **Scoring Properties** under **Judge's Settings**
5) Note that the property sheet fits nicely.
6) This has been tested on both Windows and Linux

### DIFF
--- a/src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
+++ b/src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.ui;
 
 import java.awt.BorderLayout;
@@ -14,6 +14,7 @@ import javax.swing.JTextField;
 import javax.swing.SwingUtilities;
 
 import com.ibm.webrunner.j2mclb.util.HeapSorter;
+import com.ibm.webrunner.j2mclb.util.StringComparator;
 /**
  * ScoringPropertiesPane is a inner pane for ContestInformationPane
  * Allows editing Scoring Properties for admin.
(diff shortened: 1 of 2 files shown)


src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
```diff
@@ -71,46 +69,49 @@ private MCLB getPropertyListBox() {
...
 
             HeapSorter sorter = new HeapSorter();
-            propertyListBox.setColumnSorter(1, sorter, 1);
+            sorter.setComparator(new StringComparator());
+            propertyListBox.setColumnSorter(0, sorter, 1);
 
-            propertyListBox.setColumnSize(0,lengthOfColumns);
-            propertyListBox.setColumnSize(1,lengthOfColumns);
+            propertyListBox.setColumnSize(0,PROPERTY_COL_WIDTH);
+            propertyListBox.setColumnSize(1,VALUE_COL_WIDTH);
```
clevengr, line 76: The effect of the above statement (along with the corresponding statement in method `refreshProperties()`) is to create a very wide "value" column -- even though the default data (and most reasonable data?) is very much narrower.  It seems like there should be a way to resize the width property based on the current content?  Or maybe make the `propertyListBox` resizable (following statement)?   Or maybe this is a manifestation of the fact that the Layout Manager for this component is set (in method `initialize`) to be a `BorderLayout` and then the component is added to the `CENTER` -- which I believe causes the `BorderLayout` to assign "all available space" to the component?   If so, maybe changing the layout manager to something like `FlowLayout` would cause it to shrink to "required space" instead of "all available space"?

In any case, this is certainly not any reason not to approve; just noting that it's a little bit of a "layout surprise" to the eye...

clevengr (approved):
I've looked through all the code, and run it on Windows11.  With the exception of the one minor layout issue that I noted, it all looks good to me.  (Note: I did not run it on Linux...)

src/edu/csus/ecs/pc2/ui/ScoringPropertiesPane.java
johnbrvc: The paths for Output HTML and Output Public HTML are long, in practice.  eg. `/home/icpc/pc2/wti/WebTeamInterface........ ` So, I figured it was reasonable to account for that by default.  I didn't think it was worth the time to dynamically resize the column (and control, and group box border) to account for that.    I actually changed the layout manager for the judges settings component (pane) to be BoxLayout so ti was more deterministic since the Judging Settings component doesnt resize anyway (neither do any of the other "Settings" components in there.  *sigh*

PR #1053, opened 2025-03-23
I764 throttle submissions

### LINKED ISSUE #764: For WTI submit run add governance/pacing to reduce flooding server with runs

Yes, a type of DOS can be done when using WTI.
A small script can be run to submit hundreds of runs (as fast as pssible) 

**Feature Description**: 

Prevent a team from making too many submissions within a certain period of time. Implement a governor algorithm.

_The scope of this change is ONLY for WTI, so it should be enforced in the pc2 java API (specifically in the ServerConnection class)._
NOTE: the above comment, written in the original issue, is probably no longer accurate; the scope should include not only WTI but also any other mechanism for submitting (e.g. the Fat Team Client and command-line or programmatic submissions via the CLICS API).

_This should not be enforced in InternalController because when run as shadow the server should allow many runs as fast as possible, esp if Shadowing is restarted._
NOTE: the above comment, written in the original issue, may not be accurate/relevant; see the discussion below regarding dealing with Shadow submissions.

Human intervention is the only other way to deal with this type of problem (turn of the ability to submit runs when
the condition is first observed).

**Additional context**: 
This issue first came to light in the Spring 2023 PacNW online practice contest, which contained an interactive problem called "Secret Remainder" which asked for a program to repeatedly "guess a number" in the range 110..1,000,000.  With their advisor's permission, a team ran the following script:
```python
from time import sleep
import requests
import base64

for M in range(110, 1_000_000):

  code = f"print(int(input())%({M}))"
  bytes = code.encode('ascii')
  submission = base64.b64encode(bytes).decode('ascii')
  print(f"submitting {M}")

  result = [requests.post](http://requests.post/)(
    "http://contest.icpc-vcss.org/api/teams/run",
    json = {
      'mainFile': {
        'byteData': submission,
        'fileName':'soln.py'
      },
      'extraFiles':[],
      'testFile':None,
      'additionalTestFiles':[],
      'probName':"F: Secret Remainder",
      "language": "Python 3",
      "isTest": False,
      "osName": "idk"
    },
    headers = {
      'team_id':redacted
    }
  )
  sleep(1)
```
Basically, a program which makes about a million submissions, one second apart, by invoking the PC2 WTI "submit" endpoint.

clevengr:
Adds a framework for specifying "throttling strategies" to limit run submissions.  

Specifically, defines a new interface `IThrottleStrategy`, together with multiple classes which "implement" the interface and provide different throttling strategies including `AcceptAll`, `RejectAll` (useful for testing), and `MaxSubmissionsPerMinute`.
Updates `InternalController` to apply the currently-instantiated strategy to all team submissions, performing a client-side rejection of those which exceed the throttle limit defined by the current strategy. Also updates `pc2.api.ServerConnection`, WTI-API `TeamsController`, and WTI-UI `new-run.component` to handle the new exception type which is thrown when the throttling strategy rejects a run.

Note that the current implementation "hard-codes" an instance of a strategy within `InternalController`; switching strategies currently requires updating `InternalController` (specifically, method `submitJudgeRun()`) and rebuilding the system. 
A better approach would be to support dynamic (run-time) selection of the strategy; a separate issue has been filed for that enhancement (see #1054).  Another enhancement would be the creation of a more complex strategy, such as a variable-window limit.

Fixes #764.

Windows 11, Java 1.8.381, Eclipse 2021-12 (4.22.0)

- Download the branch distribution.
- Start a contest (e.g. "sumithello").
- Start a PC2 Admin.
- Start a WTI server.
- Open a browser and login via WTI as a team.
- Use the WTI "Submit Problem" GUI to submit five runs in a row, as fast as possible -- all within the same minute (check the Runs grid on the PC2 Admin to verify the submission times are in the same minute).

The first four runs should be accepted (i.e. they should appear on the PC2 Admin's Runs grid).  The FIFTH run should be rejected, displaying a red dialog on the WTI screen saying "Submission threshhold exceeded".  This is because the currently-instantiated "ThrottleStrategy" is `MaxSubmissionsPerMinute(4)`.

To test other strategies (e.g. a different MaxSubmissionsPerMinute value, or the `AcceptAll` or `RejectAll` strategies),  edit method `InternalController.submitJudgeRun()` around line 618; comment-out the current strategy and uncomment the desired strategy then rebuild the system and start a new contest.  You'll also have to comment-out/uncomment the related `import` statements at the top of `InternalController`.

### Additional Notes
The PR code sets the current strategy to "MaxSubmissionPerMinute(4)". This was done to make it easy to test the PR via manual WTI GUI operations (it should be easy to submit 5 or more runs in a single minute, even manually).  However, "MaxSubmissionsPerMinute" of 4 is not a good value to put into production.  Before the PR is actually merged, `InternalController.submitJudgeRun()` should be edited to instantiate the `MaxSubmissionsPerMinute(DEFAULT_MAX_SUBMISSIONS_PER_MINUTE)` strategy.

### DIFF
--- a/src/edu/csus/ecs/pc2/api/ServerConnection.java
+++ b/src/edu/csus/ecs/pc2/api/ServerConnection.java
@@ -22,6 +22,7 @@ import edu.csus.ecs.pc2.core.InternalController;
 import edu.csus.ecs.pc2.core.ParseArguments;
 import edu.csus.ecs.pc2.core.PermissionGroup;
 import edu.csus.ecs.pc2.core.exception.IllegalContestState;
+import edu.csus.ecs.pc2.core.exception.SubmissionRejectedException;
 import edu.csus.ecs.pc2.core.model.Account;
 import edu.csus.ecs.pc2.core.model.ClientId;
 import edu.csus.ecs.pc2.core.model.ClientType;
@@ -600,7 +601,13 @@ public class ServerConnection {
         try {
             controller.submitJudgeRun(submittedProblem, submittedLanguage, serializedMainFile, serializedAdditionalFiles, 
                                         overrideSubmissionTimeMS, overrideRunId);
-        } catch (Exception e) {
+        } 
+        catch (SubmissionRejectedException e) {
+            //the submission was rejected, presumably because of the submitJudgeRun() method's ThrottleStrategy;
+            //forward the exception to the caller
+            throw e ;
+        }
+        catch (Exception e) {
             throw new Exception("Unable to submit run " + e);
         }
     }
--- a/src/edu/csus/ecs/pc2/api/ServerConnection.java
+++ b/src/edu/csus/ecs/pc2/api/ServerConnection.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2020 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.api;
 
 import java.io.File;
--- a/src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
+++ b/src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
@@ -1,6 +1,4 @@
-/**
- * 
- */
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core;
 
 import edu.csus.ecs.pc2.core.model.Run;
--- a/src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
+++ b/src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
@@ -1,0 +1,21 @@
+/**
+ * 
+ */
+package edu.csus.ecs.pc2.core;
+
+import edu.csus.ecs.pc2.core.model.Run;
+
+/**
+ * This interface defines the methods which a "throttler" (governor) of submission rates must implement.
+ * 
+ * @author John Clevenger
+ *
+ */
+public interface IThrottleStrategy {
+    /**
+     * This method returns true if the implementing Strategy indicates the specified Submission ("Run" in PC2 terms)
+     * should be accepted (that is, forwarded to the PC2 server); false if the Submission exceeds the 
+     * implementation-defined strategy for submission limits.
+     */
+    boolean accept(Run run);
+}
--- a/src/edu/csus/ecs/pc2/core/InternalController.java
+++ b/src/edu/csus/ecs/pc2/core/InternalController.java
@@ -78,8 +78,8 @@ import edu.csus.ecs.pc2.core.security.FileSecurity;
 import edu.csus.ecs.pc2.core.security.FileSecurityException;
 import edu.csus.ecs.pc2.core.security.Permission;
 //import edu.csus.ecs.pc2.core.strategies.AcceptAllStrategy;
-//import edu.csus.ecs.pc2.core.strategies.MaxSubmissionsPerMinuteStrategy;
-import edu.csus.ecs.pc2.core.strategies.RejectAllStrategy;
+import edu.csus.ecs.pc2.core.strategies.MaxSubmissionsPerMinuteStrategy;
+//import edu.csus.ecs.pc2.core.strategies.RejectAllStrategy;
 import edu.csus.ecs.pc2.core.transport.ConnectionHandlerID;
 import edu.csus.ecs.pc2.core.transport.IBtoA;
 import edu.csus.ecs.pc2.core.transport.ITransportManager;
@@ -621,7 +621,8 @@ public class InternalController implements IInternalController, ITwoToOne, IBtoA
             // chosen from among a list of available strategies and specified by, e.g. a YAML file or an interactive
             // GUI (such as the PC2 Admin)
 //            IThrottleStrategy strategy = new AcceptAllStrategy();
-            IThrottleStrategy strategy = new RejectAllStrategy();
+//            IThrottleStrategy strategy = new RejectAllStrategy();
+            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,4);
 //            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,MaxSubmissionsPerMinuteStrategy.DEFAULT_MAX_SUBMISSIONS_PER_MINUTE);
             
             accept = strategy.accept(run);
--- a/src/edu/csus/ecs/pc2/core/InternalController.java
+++ b/src/edu/csus/ecs/pc2/core/InternalController.java
@@ -31,6 +31,7 @@ import edu.csus.ecs.pc2.core.archive.PacketArchiver;
 import edu.csus.ecs.pc2.core.exception.ContestSecurityException;
 import edu.csus.ecs.pc2.core.exception.ProfileException;
 import edu.csus.ecs.pc2.core.exception.ServerProcessException;
+import edu.csus.ecs.pc2.core.exception.SubmissionRejectedException;
 import edu.csus.ecs.pc2.core.log.EvaluationLog;
 import edu.csus.ecs.pc2.core.log.Log;
 import edu.csus.ecs.pc2.core.log.StaticLog;
@@ -76,6 +77,9 @@ import edu.csus.ecs.pc2.core.report.ContestSummaryReports;
 import edu.csus.ecs.pc2.core.security.FileSecurity;
 import edu.csus.ecs.pc2.core.security.FileSecurityException;
 import edu.csus.ecs.pc2.core.security.Permission;
+//import edu.csus.ecs.pc2.core.strategies.AcceptAllStrategy;
+//import edu.csus.ecs.pc2.core.strategies.MaxSubmissionsPerMinuteStrategy;
+import edu.csus.ecs.pc2.core.strategies.RejectAllStrategy;
 import edu.csus.ecs.pc2.core.transport.ConnectionHandlerID;
 import edu.csus.ecs.pc2.core.transport.IBtoA;
 import edu.csus.ecs.pc2.core.transport.ITransportManager;
@@ -605,8 +609,31 @@ public class InternalController implements IInternalController, ITwoToOne, IBtoA
         Run run = new Run(contest.getClientId(), language, problem);
         RunFiles runFiles = new RunFiles(run, mainFile, otherFiles);
 
-        Packet packet = PacketFactory.createSubmittedRun(contest.getClientId(), serverClientId, run, runFiles, overrideSubmissionTimeMS, overrideRunId, overrideStopOnFailure);
-        sendToLocalServer(packet);
+        //determine whether to apply the "current throttling strategy" to this submission 
+        // (i.e., whether to accept the run for submission to the PC2 Server)
+        boolean accept = true;
+        Account submitterAccount = contest.getAccount(contest.getClientId());
+        
+        if (submitterAccount.isTeam()) {
+            //Determine the throttling strategy to be applied to team submissions.
+            //The following shows several alternative strategy selections, with only one being enabled.
+            //A preferable extension would be to allow external (e.g. run-time) selection of the desired strategy,
+            // chosen from among a list of available strategies and specified by, e.g. a YAML file or an interactive
+            // GUI (such as the PC2 Admin)
+//            IThrottleStrategy strategy = new AcceptAllStrategy();
+            IThrottleStrategy strategy = new RejectAllStrategy();
+//            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,MaxSubmissionsPerMinuteStrategy.DEFAULT_MAX_SUBMISSIONS_PER_MINUTE);
+            
+            accept = strategy.accept(run);
+        }
+        
+        Packet packet ;
+        if (accept) {
+            packet = PacketFactory.createSubmittedRun(contest.getClientId(), serverClientId, run, runFiles, overrideSubmissionTimeMS, overrideRunId, overrideStopOnFailure);
+            sendToLocalServer(packet);
+        } else {
+            throw new SubmissionRejectedException("Submission threshhold exceeded");
+        }
     }
 
     @Override
--- a/src/edu/csus/ecs/pc2/core/InternalController.java
+++ b/src/edu/csus/ecs/pc2/core/InternalController.java
@@ -620,9 +620,9 @@ public class InternalController implements IInternalController, ITwoToOne, IBtoA
             // GUI (such as the PC2 Admin)
 //            IThrottleStrategy strategy = new AcceptAllStrategy();
 //            IThrottleStrategy strategy = new RejectAllStrategy();
-            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,4);
+            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,6);
 //            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,MaxSubmissionsPerMinuteStrategy.DEFAULT_MAX_SUBMISSIONS_PER_MINUTE);
-            
+//            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest); //uses DEFAULT_MAX_SUBMISSIONS_PER_MINUTE; same as prev line
             accept = strategy.accept(run);
         }
         
--- a/src/edu/csus/ecs/pc2/core/InternalController.java
+++ b/src/edu/csus/ecs/pc2/core/InternalController.java
@@ -605,9 +605,7 @@ public class InternalController implements IInternalController, ITwoToOne, IBtoA
     public void submitJudgeRun(Problem problem, Language language, SerializedFile mainFile, SerializedFile[] otherFiles,
                                 long overrideSubmissionTimeMS, long overrideRunId, boolean overrideStopOnFailure) throws Exception {
 
-        ClientId serverClientId = new ClientId(contest.getSiteNumber(), Type.SERVER, 0);
         Run run = new Run(contest.getClientId(), language, problem);
-        RunFiles runFiles = new RunFiles(run, mainFile, otherFiles);
 
         //determine whether to apply the "current throttling strategy" to this submission 
         // (i.e., whether to accept the run for submission to the PC2 Server)
@@ -630,6 +628,8 @@ public class InternalController implements IInternalController, ITwoToOne, IBtoA
         
         Packet packet ;
         if (accept) {
+            ClientId serverClientId = new ClientId(contest.getSiteNumber(), Type.SERVER, 0);
+            RunFiles runFiles = new RunFiles(run, mainFile, otherFiles);            
             packet = PacketFactory.createSubmittedRun(contest.getClientId(), serverClientId, run, runFiles, overrideSubmissionTimeMS, overrideRunId, overrideStopOnFailure);
             sendToLocalServer(packet);
         } else {
--- a/src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
+++ b/src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
@@ -1,0 +1,54 @@
+/**
+ * 
+ */
+package edu.csus.ecs.pc2.core.exception;
+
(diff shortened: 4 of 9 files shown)


johnbrvc:
I think additional issues should be created for the different sub-systems that submit runs so that they can properly handle the new exception that is thrown.    These include WTI (which is evidently done), PC2 API (which is evidently done as well), **pc2team** (fat client), **pc2submit** (command line tool) - we should probably wait and do the new one that uses the CLICS api, and _Shadow_ `RemoteEventFeedMonitor`.  I think we have to have PR's for those 3 additional issues (pc2team, pc2submit, Shadow) before merging this.

src/edu/csus/ecs/pc2/api/ServerConnection.java
johnbrvc, line 1: Update copyright notice

src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
johnbrvc, line 1: Add copyright notice.

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -605,8 +609,32 @@ public void submitJudgeRun(Problem problem, Language language, SerializedFile ma
...
+        
+        if (submitterAccount.isTeam()) {
+            //Determine the throttling strategy to be applied to team submissions.
+            //The following shows several alternative strategy selections, with only one being enabled.
+            //A preferable extension would be to allow external (e.g. run-time) selection of the desired strategy,
+            // chosen from among a list of available strategies and specified by, e.g. a YAML file or an interactive
+            // GUI (such as the PC2 Admin)
+//            IThrottleStrategy strategy = new AcceptAllStrategy();
+//            IThrottleStrategy strategy = new RejectAllStrategy();
+            IThrottleStrategy strategy = new MaxSubmissionsPerMinuteStrategy(contest,4);
```
johnbrvc: Make the 4 a constant so we can easily change it as suggested in the PR when we merge.

src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
johnbrvc, line 1: Add copyright notice

src/edu/csus/ecs/pc2/core/strategies/AcceptAllStrategy.java
johnbrvc, line 1: Add Copyright notice.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
johnbrvc, line 1: Add copyright notice.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
```diff
@@ -0,0 +1,111 @@
...
+        ClientId newSubmitter = newRun.getSubmitter();
+
+        //a list of runs submitted by the same submitter as who submitted "newRun", initially empty
+        ArrayList<Run> teamRuns = new ArrayList<Run>();
+
+        //get all the runs in the contest
+        Run[] allRuns = contest.getRuns();
+        
+        //check each contest run to see if it came from the same submitter
+        for (Run curRun : allRuns ) {
```
johnbrvc: Is there a reason the 2 loops can't be combined into one?  Precompute the cut-off time, `contestElapsedSecs` (which you do), but only do one loop and just count instead of making an collection of runs up and then rescanning them.   My guess is, this is going to be the strategy used at NAC (at least).  4 per minutes seems reasonable.  Maybe make another strategy that checks if the problem is the same too?

src/edu/csus/ecs/pc2/core/strategies/RejectAllStrategy.java
johnbrvc, line 1: Add copyright notice.

johnbrvc:
Code review complete.  Looks good.  We just have to decide on what strategy to implement for NAC.  We also have to decide on how we are going to merge this.  Must it be simultaneously merged with the changes for the other subsystems that make submissions?

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: > Update copyright notice

Done; next push.

src/edu/csus/ecs/pc2/core/IThrottleStrategy.java
clevengr: > Add copyright notice.

Done; next push.

src/edu/csus/ecs/pc2/core/InternalController.java
clevengr: > Make the 4 a constant so we can easily change it as suggested in the PR when we merge.

I can do that.  The down side of doing that is that if someone decides to select a different strategy by commenting out this line and uncommenting one of the other strategy constructors, Eclipse will (depending on the local developer's configuration) suddenly generate either a warning or an error on the definition of the (now) "unused" constant.  So which is better -- opening the possibility of generating confusing errors in the IDE, or providing a single place where the value can be changed?  

Note that there is _already_ a definition of `DEFAULT_MAX_SUBMISSIONS_PER_MINUTE`, in class `MaxSubmissionsPerMinuteStrategy`.  So what would (should) the relationship be between that existing definition and the proposed similar definition in THIS class?

src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
clevengr: > Add copyright notice

Done; next push.

src/edu/csus/ecs/pc2/core/strategies/AcceptAllStrategy.java
clevengr: > Add Copyright notice.

Done; next push.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
clevengr: > Add copyright notice.

Done; next push.

src/edu/csus/ecs/pc2/core/strategies/RejectAllStrategy.java
clevengr: > Add copyright notice.

Done; next push.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
clevengr: > Is there a reason the 2 loops can't be combined into one? Precompute the cut-off time, `contestElapsedSecs` (which you do), but only do one loop and just count instead of making an collection of runs up and then rescanning them. My guess is, this is going to be the strategy used at NAC (at least). 4 per minutes seems reasonable. Maybe make another strategy that checks if the problem is the same too?

This comment comprises two different things:  the one-loop-vs-two-loops issue, and the question of whether to develop a new strategy for NAC (and what that would look like).

For the first: agreed.  Next push will combine the loops.

For the second:  
- I'm undecided if 4 per minute is "reasonable".  That's one every 15 seconds.  I think somebody could use a GUI faster than that.  So, say it takes 6 seconds per GUI interaction; that would be 10 per minute.  I think that's reasonable (generous, but reasonable).
- Regarding "another strategy that checks if the problem is the same":  I'm undecided whether this is necessarily good or not.  I think what we're trying to do is limit "scripted submissions at a high rate".  Wouldn't we WANT to include "all problems" in such a limit?

clevengr:
> I think additional issues should be created for the different sub-systems that submit runs so that they can properly handle the new exception that is thrown. These include WTI (which is evidently done), PC2 API (which is evidently done as well), **pc2team** (fat client), **pc2submit** (command line tool) - we should probably wait and do the new one that uses the CLICS api, and _Shadow_ `RemoteEventFeedMonitor`. I think we have to have PR's for those 3 additional issues (pc2team, pc2submit, Shadow) before merging this.

This PR is about supporting throttling -- primarily, protection against scripted flooding -- for NAC and WF Baku.  You said `I think we have to have PR's for those 3 additional issues (pc2team, pc2submit, Shadow) before merging this.`  **pc2team** (i.e. the Fat Team Client) only supports GUI-based submissions; as far as I can see there's no path by which the Fat Team Client could be used in a scripting attempt.  (In addition, we're probably not even going to use that tool for NAC/Baku, although it's true that the plan is to put it on the machine.)  These factors lead me to think requiring a **PR** for Fat Team Client isn't necessary.

Also, we have no plans to use PC2 for Shadowing at NAC/Baku, so I don't see why we need a **PR** for fixing Shadowing.

I do agree we should file _Issues_ for these two, and perhaps we might indeed need a PR for the _pc2submit_ case.  But I don't see that requiring a PR for Fat Team Client or Shadowing is necessary.

kkarakas:
LGTM

kkarakas (approved):

src/edu/csus/ecs/pc2/core/InternalController.java
```diff
@@ -605,8 +609,32 @@ public void submitJudgeRun(Problem problem, Language language, SerializedFile ma
         Run run = new Run(contest.getClientId(), language, problem);
         RunFiles runFiles = new RunFiles(run, mainFile, otherFiles);
```
johnbrvc: The creation of the RunFiles and serverClientId objects can be moved to the point where PacketFactory.CreateSubmission() is called since it is only used if the submission is actually made.  This can save overhead.  It seems necessary to create the Run object since I can envision a IThrottleStrategy using submitter, language and/or property to make a decision, so creating the Run object is worth it.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+        long contestElapsedSecs = contestTime.getElapsedSecs();
+        
+        //get all the runs in the contest
+        Run[] allRuns = contest.getRuns();
+        
+        //count the number of runs previously submitted in the last minute by the same client
+        int numSubmittedInLastMinute = 0;
+        
+        //check each contest run 
+        for (Run curRun : allRuns ) {
```
johnbrvc, line 80: I'm not in love with this loop.  Since this will be called for every submission, it might make sense to scan the list of runs backwards and stop as soon as it finds the limit is exceeded (or the time is past 1 minute).

johnbrvc (commented):
Couple of comments we should address below.

johnbrvc:
My testing is complete and the code seems to work fine.  I used python pc2submit, using the default of 4 subs/minute:

```
C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Submission received: id = s2279, time = 15:49:16

C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Submission received: id = s2280, time = 15:49:17

C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Submission received: id = s2281, time = 15:49:20

C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Submission received: id = s2282, time = 15:49:22
```
The next 2 submissions should fail since I submitted them right after the one above:
```
C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Traceback (most recent call last):
  File "C:\MyNetProjects\ICPC\pc2v9\bin\pc2submit", line 772, in <module>
    do_api_submit()
  File "C:\MyNetProjects\ICPC\pc2v9\bin\pc2submit", line 496, in do_api_submit
    raise RuntimeError(f'Submission failed (code {response.status_code} - {response.text})')
RuntimeError: Submission failed (code 500 - Unable to submit run: Submission threshhold exceeded)

C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40397 minutes!
Traceback (most recent call last):
  File "C:\MyNetProjects\ICPC\pc2v9\bin\pc2submit", line 772, in <module>
    do_api_submit()
  File "C:\MyNetProjects\ICPC\pc2v9\bin\pc2submit", line 496, in do_api_submit
    raise RuntimeError(f'Submission failed (code {response.status_code} - {response.text})')
RuntimeError: Submission failed (code 500 - Unable to submit run: Submission threshhold exceeded)
```
I waited about minute and tried to make a couple of submissions which now works since the previous submissions timed out:
```
C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40398 minutes!
Submission received: id = s2283, time = 15:50:21

C:\MyNetProjects\ICPC\pc2v9\bin>py pc2submit -y -u https://localhost:50443 -p c ads-wa-case2.cpp
Warning: `ads-wa-case2.cpp' has not been modified for 40398 minutes!
Submission received: id = s2284, time = 15:50:26
```

I am happy to approve this is we address the comments I had about the code.

src/edu/csus/ecs/pc2/core/InternalController.java
clevengr: Agreed; fixed in next push.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
clevengr: I looked into this.  The problem I see with replacing the current code with a loop that "runs backwards" is that 
`allRuns` is an array whose values were obtained by the call `Run[] allRuns = contest.getRuns();`. `getRuns()` in turn calls `runList.getList()`.  The runs in class `RunList` are stored in a hashtable; method `getList()` returns an array of Runs by executing
```java
return (Run[]) runHash.values().toArray(new Run[size()]);
```
This means that there is no guarantee that the elements in the returned array are in any way sorted by submission time.  This in turn means that in order to "run backwards" through the runs list it would first be necessary to _sort_ the returned array based on submission time.  I'm not sure that invoking an array sorter _for every submission_ is any better than simply looping through all the runs every time.

I suppose an alternative would be to examine all the runs in the array, extracting only those that came from the same submitter, and _then_ sort that smaller list on submission time -- or perhaps insert them into a priority queue based on submission time -- but still you'd have to process every run, and now you've added sorting or queuing time on top of that.  I'm not clear that there's any way to improve the process, short of totally changing the `RunList` structure (which I'd be loathe to do at this point...)

Thoughts?

clevengr:
@johnbrvc I pushed a fix for one of your two requests.  See my comment regarding the other one and let me know your thoughts.

src/edu/csus/ecs/pc2/core/InternalController.java
johnbrvc: Eventually, I assume these values will be configurable somewhere (as previously mentioned).  I think it's fine to make a constant NOW, and then whoever changes the code will be responsible for removing / commenting out old code.  Alternatively, to avoid the "unused" thing, maybe, instead of commenting out the code, "if(false)" the code?
This is not a show stopper, and has nothing to do with the actual implementation the PR is for (as Doug might say, "out of scope")?

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
johnbrvc: I'm marking this as resolved.   Specifics to how strategies actually work are out of scope.  We can add issues as we need new strategies.

johnbrvc:
> > I think additional issues should be created for the different sub-systems that submit runs so that they can properly handle the new exception that is thrown. These include WTI (which is evidently done), PC2 API (which is evidently done as well), **pc2team** (fat client), **pc2submit** (command line tool) - we should probably wait and do the new one that uses the CLICS api, and _Shadow_ `RemoteEventFeedMonitor`. I think we have to have PR's for those 3 additional issues (pc2team, pc2submit, Shadow) before merging this.
> 
> This PR is about supporting throttling -- primarily, protection against scripted flooding -- for NAC and WF Baku. You said `I think we have to have PR's for those 3 additional issues (pc2team, pc2submit, Shadow) before merging this.` **pc2team** (i.e. the Fat Team Client) only supports GUI-based submissions; as far as I can see there's no path by which the Fat Team Client could be used in a scripting attempt. (In addition, we're probably not even going to use that tool for NAC/Baku, although it's true that the plan is to put it on the machine.) These factors lead me to think requiring a **PR** for Fat Team Client isn't necessary.
> 
> Also, we have no plans to use PC2 for Shadowing at NAC/Baku, so I don't see why we need a **PR** for fixing Shadowing.
> 
> I do agree we should file _Issues_ for these two, and perhaps we might indeed need a PR for the _pc2submit_ case. But I don't see that requiring a PR for Fat Team Client or Shadowing is necessary.

Agreed.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
johnbrvc: Ooooh... right.  Hash tables.   You know, it occurred to me at one point when looking at the code that maybe the `Runs `are not in order due to how they're stored, but it was a fleeting thought - I saw "array" and assumed "in order".  
No sense in sorting or anything like that - in fact - that adds MORE overhead.  The full scan is fine, but, I still think that as soon as you find "X" runs in the last minute, the loop should break out - we know at that point that the run won't be accepted, so why bother scanning further?

clevengr: Agreed; fixed in next push.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
```diff
@@ -91,12 +89,18 @@ private int getNumberOfRunsSubmittedInTheLastMinute(IInternalContest contest, Ru
                 if (runSubmisionTimeSecs >= contestElapsedSecs-60) {
                     
                     //the run was submitted in the last minute; count it
-                    numSubmittedInLastMinute++ ;
+                    countSoFar++ ;
+                    
+                    //if the submitter has already exceeded the limit, no need to check any more runs
+                    if (countSoFar > maxPerMinute) {
```
johnbrvc: Shouldn't this be if countSoFar >= maxPerMinute ?

johnbrvc (changes requested):
Made a comment that probably requires a change regarding checking the limit.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
clevengr: Yes; pushed a fix for this.   This was a side-effect of converting the method from `int getNumberOfSubmissions()` to `boolean getIfLimitExceeded()`; previously it was counting the last-found one; now it should be.  Fixed.

I also bumped the current limit-per-minute to 6.  If you approve these additional commits, this is ready to be merged.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
```diff
@@ -0,0 +1,106 @@
...
+        this.contest = inContest ;
+        this.maxPerMinute = maxAllowed ;
+    }
+
+    @Override
+    /**
+     * Rejects the specified run if the submitting team has already submitted the maximum allowed number of runs in the past minute. 
+     */
+    public boolean accept(Run run) {
+        return numSubmittedInTheLastMinuteExceedsMaximum(contest, run);
```
johnbrvc: I think this should be:
`return !numSubmittedInTheLastMinuteExceedsMaximum(contest, run);`
Basically, I can't make *any* submissions ATM.

johnbrvc (approved):
Reviewed and tested.

src/edu/csus/ecs/pc2/core/strategies/MaxSubmissionsPerMinuteStrategy.java
clevengr: Fixed in most recent push.

PR #1074, opened 2025-04-21
I1073 Control which notifications on the event feed can use collections

### LINKED ISSUE #1073: Provide option to control the use of Event Feed collections
The **CDS** processes the "`accounts`" notification differently depending on whether the notification is a collection or a series of single notifications.  This causes a credential problem unless the accounts match the CDS' pre-configured account list  precisely.
...

**Feature Description**:
It has been brought to our attention that some applications that consume the _CLICS 2023-06 Event Feed_ handle _collections_ differently.  _Collections_ are a notification that contains an array of data objects for the notification type.  In order to avoid issues with some applications, it would be best to be able to disable the use of this feature either on an individual notification basis or for all notifications.
...

Make sure that accounts provided by the `accounts `notification collection precisely match the configured list of accounts on the CDS.  This can be problematic since PC2 only supports accounts with login Id's of _type_#, where _type_ is one of **administrator**, **judge**, **team**, **feeder**, **scoreboard**, and other clients, such as a CDS, may have account names such as "**admin**".
...

In the `pc2v9.ini` file, a "`[clics]`" section is already supported.  Add additional properties under this section to support this new feature such as:

```
disable-collections=all
disable-collections=accounts
disable-collections=accounts,teams,problems

```
PC2 already has a `boolean useCollections` to control whether collections are used for everything (all), but it is a compile time option only at the current time.
...

**Additional context**:   
...

johnbrvc:
Allow specification of when to use notification collections on the event feed in the `pc2v9.ini `file.
Change default setting for the `disable-collections` property in the `pc2v9.ini` file due to the way the ICPC tools CDS works.  We do not want to send `accounts` as collections since the CDS will interpret a collection as a complete list of accounts and remove any accounts configured by the CDS.  This is not desirable since it would loose the CDS' 'admin' account privileges and other CDS accounts.

CI: Give the EventFeed feeder log (list of event feed notifications sent) a unique name based on the client doing the feeding.   This will prevent event feed log corruption if someone decides to run 2 event feeders are run on different ports, for example. Ex.
Old Name: `Eventfeed_2023_06_SumH.log` 
New Name after CI: `Eventfeed_2023_06_SumH_feeder1.log`
CI: change endpoint from "**team-members**" to "**persons**", even though we don't use this at the moment.  We want it right for the future.

Fixes #1073
Windows 11
java version "1.8.0_321"

1. Load and bring up the clics_sumithello sample contest.
2. Generate a feeder account (feeder1).
3. Start the event feeder client using the newly generated feeder1 account.
4. Start the event feed feeding the CLICS 2023-06 API.
5. Using curl, or another utility, access the event feed: https://administrator1:administrator1@localhost:50443/contests/SumH/event-feed
6. Note that the accounts notifications are now individual instead of grouped into a collection (like the other notifications: teams, problems, etc).  
[SampleEventFeed.txt](https://github.com/user-attachments/files/19835577/SampleEventFeed.txt)

7. You can try setting the disable-collections property in the pc2v9.ini file to "all", or add additional notification types to the list to test it out.
8. You can leave out (comment out) the disable-collections line in the pc2v9.ini file to allow collections on all notification types.

### DIFF
--- a/src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
+++ b/src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
@@ -18,6 +18,8 @@ import edu.csus.ecs.pc2.core.list.AccountComparator;
 import edu.csus.ecs.pc2.core.list.ClarificationComparator;
 import edu.csus.ecs.pc2.core.list.GroupComparator;
 import edu.csus.ecs.pc2.core.list.RunComparator;
+import edu.csus.ecs.pc2.core.log.Log;
+import edu.csus.ecs.pc2.core.log.StaticLog;
 import edu.csus.ecs.pc2.core.model.Account;
 import edu.csus.ecs.pc2.core.model.Clarification;
 import edu.csus.ecs.pc2.core.model.ClarificationAnswer;
@@ -61,11 +63,17 @@ public class EventFeedJSON extends JSON202306Utilities {
         super();
         this.jsonTool = jsonTool;
 
-        try {
-            String [] disList = CommaSeparatedValueParser.parseLine(IniFile.getValue(CLICS_DISABLE_COLLECTIONS_KEY));
-            this.disableNotificationCollections(disList);
-        } catch (Exception e) {
-            // Use default values for collections
+        String disCSVList = IniFile.getValue(CLICS_DISABLE_COLLECTIONS_KEY);
+        if(disCSVList != null && !disCSVList.isEmpty()) {
+            try {
+                String [] disList = CommaSeparatedValueParser.parseLine(disCSVList);
+                this.disableNotificationCollections(disList);
+            } catch (Exception e) {
+                Log log = StaticLog.getLog();
+                if(log != null) {
+                    log.warning("Error parsing INI key: " + CLICS_DISABLE_COLLECTIONS_KEY + " (" + disCSVList + "): " + e);
+                }
+            }
         }
     }
 
@@ -1089,7 +1097,7 @@ public class EventFeedJSON extends JSON202306Utilities {
 
     /**
      * Tests if notification collections should be used for a specific notification type.
-     * This first tests is notification collections are being used in general, then if so,
+     * This first tests if notification collections are being used in general, then if so,
      * checks if the specific type is not disabled.
      *
      * @param notificationType
@@ -1100,12 +1108,12 @@ public class EventFeedJSON extends JSON202306Utilities {
     }
 
     /**
-     * Adds the comma-separated list of notification types (EventFeedType's) to the HashSet of
+     * Adds the array of notification types (EventFeedType strings) to the HashSet of
      * disabled notification types.  Any notification type in this list will not use collections
      * on the event feed.  The special type ALL_NOTIFICATION_TYPES will disable collections for all
-     * notification and will clear the HashSet of disabled types and simply set the useCollections flag to false.
+     * notifications and will clear the HashSet of disabled types and simply set the useCollections flag to false.
      *
-     * @param list Comma-separated list of notification types (EventFeedType's) to not use
+     * @param list array of notification types (EventFeedType strings) to not use
      *      collections on.
      */
     public void disableNotificationCollections(String [] list) {
@@ -1122,13 +1130,14 @@ public class EventFeedJSON extends JSON202306Utilities {
     }
 
     /**
-     * Remove the comma-separated list of notification types (EventFeedType's) from the Hashset
-     * of disabled notification types.  If collections are enabled, then the types included in the
-     * supplied list will be allowed to generate collections.  The special type ALL_NOTIFICATION_TYPES will disable collections for all
+     * Remove the array of notification types (EventFeedType strings) from the Hashset
+     * of disabled notification types.  The types included in the
+     * supplied list will be allowed to generate collections, assuming collections are
+     * being used (useCollections).  The special type ALL_NOTIFICATION_TYPES will enable collections for all
      * notification and will simply set the useCollections flag to true and clear the HashSet of
      * disabled types.
      *
-     * @param list Comma-separated list of notification types (EventFeedType's) to allow collections on.
+     * @param list array of notification types (EventFeedType strings) to allow collections on.
      */
     public void enableNotificationCollections(String [] list) {
         if(list != null) {
--- a/src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
+++ b/src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
@@ -12,6 +12,7 @@ import javax.ws.rs.core.SecurityContext;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import edu.csus.ecs.pc2.core.IniFile;
 import edu.csus.ecs.pc2.core.exception.IllegalContestState;
 import edu.csus.ecs.pc2.core.list.AccountComparator;
 import edu.csus.ecs.pc2.core.list.ClarificationComparator;
@@ -33,6 +34,7 @@ import edu.csus.ecs.pc2.core.model.Problem;
 import edu.csus.ecs.pc2.core.model.Run;
 import edu.csus.ecs.pc2.core.model.RunTestCase;
 import edu.csus.ecs.pc2.core.security.Permission;
+import edu.csus.ecs.pc2.core.util.CommaSeparatedValueParser;
 import edu.csus.ecs.pc2.core.util.IJSONTool;
 
 /**
@@ -46,7 +48,11 @@ public class EventFeedJSON extends JSON202306Utilities {
     public static final String EVENT_ID_PREFIX = "pc2-";
     public static final int EVENT_ID_PREFIX_LENGTH = EVENT_ID_PREFIX.length();
 
+    public static final String CLICS_DISABLE_COLLECTIONS_KEY = "clics.disable-collections";
+    public static final String ALL_NOTIFICATION_TYPES = "all";
+
     private boolean useCollections = true;
+    private HashSet<String> disableCollections = new HashSet<String>();
 
     /**
      *
@@ -54,6 +60,13 @@ public class EventFeedJSON extends JSON202306Utilities {
     public EventFeedJSON(IJSONTool jsonTool) {
         super();
         this.jsonTool = jsonTool;
+
+        try {
+            String [] disList = CommaSeparatedValueParser.parseLine(IniFile.getValue(CLICS_DISABLE_COLLECTIONS_KEY));
+            this.disableNotificationCollections(disList);
+        } catch (Exception e) {
+            // Use default values for collections
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


SamanwaySadhu (approved):
LGTM!

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -46,14 +48,25 @@ public class EventFeedJSON extends JSON202306Utilities {
...
      */
     public EventFeedJSON(IJSONTool jsonTool) {
         super();
         this.jsonTool = jsonTool;
+
+        try {
+            String [] disList = CommaSeparatedValueParser.parseLine(IniFile.getValue(CLICS_DISABLE_COLLECTIONS_KEY));
+            this.disableNotificationCollections(disList);
+        } catch (Exception e) {
+            // Use default values for collections
```
clevengr: It seems like there should at least be a log message generated when this happens.  Otherwise, any error in parsing the IniFile effectively becomes "silently ignore the error".

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+     *
+     * @param useCollections - true to try to use collections
+     */
     public void setUseCollections(boolean useCollections) {
         this.useCollections = useCollections;
     }
+
+    /**
+     * Tests if notification collections should be used for a specific notification type.
+     * This first tests is notification collections are being used in general, then if so,
```
clevengr: I think the word "is" was intended to be "if"?

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+     *
+     * @param notificationType
+     * @return
+     */
+    public boolean isUseNotificationCollection(EventFeedType notificationType) {
+        return(isUseCollections() && !disableCollections.contains(notificationType.toString()));
+    }
+
+    /**
+     * Adds the comma-separated list of notification types (EventFeedType's) to the HashSet of
```
clevengr: I think it's confusing to refer here to "comma-separated list" -- it makes the reader think that what's being passed into this method is a comma-separated list, when in fact it's not -- the "comma-separators" were removed by the call to `CommaSeparatedValueParser.parseLine()` (around line 65).  What's being passed into this method is a String array of notification types which are to be disabled.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+                    disableCollections.clear();
+                    break;
+                }
+                disableCollections.add(notificationType);
+            }
+        }
+    }
+
+    /**
+     * Remove the comma-separated list of notification types (EventFeedType's) from the Hashset
```
clevengr: Same as the previous comment:  I think it's confusing to refer here to "Removing a comma-separated list..." -- there's no comma-separated list being processed by this method.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+     * of disabled notification types.  If collections are enabled, then the types included in the
+     * supplied list will be allowed to generate collections.  The special type ALL_NOTIFICATION_TYPES will disable collections for all
+     * notification and will simply set the useCollections flag to true and clear the HashSet of
+     * disabled types.
+     *
+     * @param list Comma-separated list of notification types (EventFeedType's) to allow collections on.
```
clevengr: As previously mentioned, referring to a "comma-separated list" here is confusing, because that's not what the method is receiving.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+        return(isUseCollections() && !disableCollections.contains(notificationType.toString()));
+    }
+
+    /**
+     * Adds the comma-separated list of notification types (EventFeedType's) to the HashSet of
+     * disabled notification types.  Any notification type in this list will not use collections
+     * on the event feed.  The special type ALL_NOTIFICATION_TYPES will disable collections for all
+     * notification and will clear the HashSet of disabled types and simply set the useCollections flag to false.
+     *
+     * @param list Comma-separated list of notification types (EventFeedType's) to not use
```
clevengr: As mentioned above, the actual parameter here is not a comma-separated list; I think this is confusing.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+                }
+                disableCollections.add(notificationType);
+            }
+        }
+    }
+
+    /**
+     * Remove the comma-separated list of notification types (EventFeedType's) from the Hashset
+     * of disabled notification types.  If collections are enabled, then the types included in the
+     * supplied list will be allowed to generate collections.  The special type ALL_NOTIFICATION_TYPES will disable collections for all
```
clevengr: The above comment reads wrong to me for a couple of reasons.  

First, it says `if collections are enabled then the types included in the supplied list will be allowed to generate collections`.  If I'm reading the following code correctly, then when `ALL_NOTIFICAITON_TYPES` is specified, the effect is that _all_ notifications are allowed to use collections -- not just "the types included in the supplied list".  (Perhaps there is some confusion between the definitions "collections are enabled" and "ALL_NOTIFICATION_TYPES" -- but in any case I can't reconcile the comment with what it appears the code is actually doing.  But since there's actually no test in the method code for the condition "collections are enabled", I'm assuming that's what the test `if(notificationType.equals(ALL_NOTIFICATION_TYPES))` is testing.  But again, I can't reconcile the difference between what the comment says and what the code appears to be doing.)

Second, the comment says `The special type ALL_NOTIFICATION_TYPES will _disable_ [emphasis mine] collections for all notification...`.  It appears to me that what the code is actually doing in this case is _enabling_ all notifications, not disabling them.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
```diff
@@ -1056,13 +1069,80 @@ public void setFilter(Filter filter) {
...
+     * notification and will simply set the useCollections flag to true and clear the HashSet of
+     * disabled types.
+     *
+     * @param list Comma-separated list of notification types (EventFeedType's) to allow collections on.
+     */
+    public void enableNotificationCollections(String [] list) {
```
clevengr, line 1142: It does not appear that this method is invoked anywhere in the workspace (pc2v9 project).  Is that intentional?

clevengr (changes requested):
Reviewed changes to `src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java`.  Request one change (addition of logging of an error condition), plus recommend several changes to documentation.

clevengr (commented):
Reviewed all the changed files.  Requested one change in `EventFeedJSON`, plus several recomended (minor) fixes.  All the other files look good.  Proceeding to testing.

clevengr:
Completed execution test; appears to work as advertised.  I'm ready to approve pending response to my comment(s) in `EventFeedJSON`.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Fixed in next push.  This was more work than it was worth, I think....

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Yes.  Fixed in next push.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Yes, since I added a "disableNotificationCollections", it seemed fitting to add an "enableNotificationCollections" for future use.

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: Maybe the confusion is, "If collections are enabled, then..." The intention here is: when making use of these flags and the Hashset, if the "useCollections" member is true, meaning, that collections are enabled, eg "isUseCollections()", then we are effectively removing these items in the supplied list from the hashset, thereby allowing those things in the array to generate collections.  So, it "undoes" the disableNotificationCollections for a list of notification types, allowing them to generate collections, iff useCollections is true.

Setting *any* element of the supplied list[] to ALL_NOTIFICATION_TYPES (all) basically enables everything to generate collections, regardless of any other things on the list.  "all" overrides everything in both cases (disable and enable).

src/edu/csus/ecs/pc2/clics/API202306/EventFeedJSON.java
johnbrvc: I added this log message:
`250425 113341.319|WARNING|qtp1050320876-39|log|Error parsing INI key: clics.disable-collections ("accounts): java.lang.Exception: String missing closing double quote '"accounts'`

clevengr: I didn't really intend that you actually figured out how to tell the user what the specific problem was -- I only suggested that a log message indicating there was problem would be better than "silently ignoring" the problem.  I can see why it was a lot of work doing what you did...

clevengr (approved):
Looks good; I approve this PR.

PR #1098, opened 2025-05-11
i1097: keep clar Submit button disabled until text AND problem selected

### LINKED ISSUE #1097: WTI displays poor error message when submitting a clar without selecting a problem
**Describe the issue**: 

  If a WTI user selects "Submit Clarification", enters some text, and clicks "Submit" without selecting a problem, what the WTI-UI code does is POST a clarification to the WTI Server.  The Server responds with a "Bad Request" message (which it should).  However, the message displayed to the user isn't helpful:

Image

The underlying issue is that the WTI-UI code should not have enabled the "Submit" button until both text has been entered AND a problem has been selected.

**To Reproduce**: 
- Start a contest.
- Start the WTI server.
- Open a browser and login to the WTI as a team.
- Select "Clarifications", then "Submit Clarification".
- Enter some text in the text box, but do NOT select a problem.
- Click "Submit".

**Expected behavior**: 
  The system should either refuse to allow "Submit" to be enabled unless both clar text has been entered AND a problem has been selected.

**Actual behavior**: 
  See Description.

**Environment**: 
 Observed under Windows 11 with Chrome Version 134.0.6998.178 (Official Build) (64-bit), but it probably happens on every platform.

**Additional context**: 
  Currently, `WTI-UI/new-clarification.component.ts` has no `Validators.required` entry for the "problem" field in `buildForm()` (it DOES have such an entry for the text field).  It's likely that simply adding this entry will keep the Submit button disabled until both clar text has been entered AND a problem has been selected, although this hasn't been confirmed.

clevengr:

  Adds `Validators.required` to the `problem` entry in the WTI `new-clarification.component`'s `buildForm()` method.  The effect of this is to keep the `Submit` button on that form disabled until both clarification text has been entered AND a problem has been selected.

  Further details:  `newClarificationForm` is defined as a `FormGroup` in `new-clarification.component.ts`.  `new-clarification.component.html` selects this formGroup, assigns method `submitNewClarification()` as the action handler for the `Submit` button, and marks the button as disabled until the form state becomes `valid`.  The `content` field in the form (the text box) was already annotated as `content: [undefined, [Validators.required]]`.  All this PR does is add that same Validation requirement to the Problem drop-down.  (In other words, the PR now keeps the `Submit` button disabled until BOTH some text has been entered AND a problem has been selected.

Fixes #1097 

Note:  the PR does not actually "fix" the error message which was being generated; what it does is eliminate the underlying cause of the error message.  The Server will still return `BAD REQUEST` if (somehow) the user figures out a way to bypass form validation (e.g. by hacking the browser's JavaScript at run time...)

Windows 11, Eclipse 2021-12, java version "1.8.0_381"

- Start a PC2 server, loading a contest with at least one team and one problem (e.g. `sumithello`)
- Start the contest.
- Start a WTI Server.
- Open a browser; if you've previously used this browser to connect to PC2 then _clear the browser's data cache_.
- Connect the browser to the WTI Server
- Login as a team.
- Select "Clarifications", then "New Clarification".
- Enter some text in the text box, but do NOT select a problem.
- Verify that the "Submit" button is disabled.
- Select a problem.
- Verify that the Submit button becomes enabled.
- Submit the clar; verify that what comes back is a green "Clar Submitted" message.

### DIFF
--- a/projects/WTI-UI/src/app/modules/clarifications/components/new-clarification/new-clarification.component.ts
+++ b/projects/WTI-UI/src/app/modules/clarifications/components/new-clarification/new-clarification.component.ts
@@ -60,7 +60,7 @@ export class NewClarificationComponent implements OnInit, OnDestroy {
 
   private buildForm(): void {
     this.newClarificationForm = this._formBuilder.group({
-      problem: [undefined, []],
+      problem: [undefined, [Validators.required]],
       content: [undefined, [Validators.required]]
     });
   }


johnbrvc (approved):
Tested. Seems to do what the PR says it should.

PR #1101, opened 2025-05-13
i_841 Deal with kotlin lower case source file names

### LINKED ISSUE #841: Kotlin source files that do not start with upper case letters get compile errors
**Describe the issue**:
Kotlin file names do not have to be uppercase, but, the Kotlin compiler morphs the name of the generated class files to start with upper case letters.  ex.  `convex_finn.kt  `produces  `Convex_finnKt.class` as an output.  The PC2 compile result validator is looking for `convex_finn_Kt.class`.  As a result, it does not find it, and fails, generating a **CE.**

**To Reproduce**: 
As described above, create a Kotlin submission that has a lower case letter as the first letter. 
Submit it.
**CE** will be generated.

**Expected behavior**: 
Kotlin compiles of files that start with a lower-case letter, and have no errors, should not produce **CE.** 

**Actual behavior**:
A CE is generated if a Kotlin source file starts with a lower case letter.

**Environment**: 
Ubuntu 22.04

**Log Info**:  
**Screenshots**: 

**Additional context**: 
Perhaps add a checkbox (boolean) for the problem execution result that indicates that a case-insensitive compare of the compiler resulting file should be performed.

johnbrvc:
When kotlin compiles a file that starts with a lower case letter, it produces a class file that starts with an upper case letter.  As a result, the PC2 routine to check if the compile worked failed because, for example, `testprogKt.class` != `TestprogKt.class`.   This PR checks the compile result filename for this condition, and then uses the correct name when executing the result.

Fixes #841 

Ubuntu 24.04
Build:
java version "1.8.0_321"
Execute:
openjdk version "21.0.4" 2024-07-16
OpenJDK Runtime Environment (build 21.0.4+7-Ubuntu-1ubuntu224.04)
OpenJDK 64-Bit Server VM (build 21.0.4+7-Ubuntu-1ubuntu224.04, mixed mode, sharing)

Make a Kotlin submission using a Kotlin source file that starts with a lower case letter.  You can use this file, and rename the extension to ".kt": [columbo_finn.txt](https://github.com/user-attachments/files/20195149/columbo_finn.txt)
The submission should be accepted and not generate a compile error.  Previously a compile error would result.
It will almost certainly generate a **WA** result for whatever problem you submit it as, but that is fine - it means the program successfully executed, but produced the wrong answer.  If you get an **RTE**, it means it probably didn't execute properly.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core.execute;
 
 import java.io.BufferedInputStream;
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -2826,6 +2826,11 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
             executionData.setCompileStderr(new SerializedFile(prefixExecuteDirname(COMPILER_STDERR_FILENAME)));
 
             program = new File(prefixExecuteDirname(programName));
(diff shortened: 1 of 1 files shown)


clevengr (commented):
Reviewed code changes; they look reasonable to me.  I'll have to install Kotlin on my machine to run an execution test...

clevengr (approved):
I think this PR is safe enough, and we need it for NAC, so I'm marking it approved.

Issue #1104, opened 2025-05-24
WTI-UI should not allow submission of Runs or Clars if the contest is not running (including that it has ended)..


clevengr:
**Describe the issue**: 

It is currently possible for a team to submit a Run, or a Clarification, when the contest isn't running, including after the contest has ended.  The result of doing that is a very unhelpful message:  

![Image](https://github.com/user-attachments/assets/8c5be174-042c-4317-b258-9fccfb66f32e)

(Note that this is very similar to - _but not identical to_ - a problem which was fixed in PR #1098.  The issue which that PR fixed is that a team could submit a clar without first selecting a problem.  The issue being described here produces the same (unhelpful) error message as what was described in that other PR -- but it's happening because the contest is not running, possibly because it has ended.)

A similar issue exists with submitting a new Run:  the same (unhelpful) error message is displayed when the Run submission occurs after the contest has ended.

Note that teams _shouldn't_ be submitting after the contest -- but they frequently do, because of "last-minute attempts" that just miss the shutoff.  Also, teams might try to submit when the contest clock is _paused_... this will likely also produce the same (unhelpful) error message.  

The WTI-UI should not allow attempting to submit something if the contest is not running -- including after the contest has ended. Teams doing these things need to get a "Contest is not running" (or "Contest is over") message, not a generic "Bad request" message.

**To Reproduce**:
- Start a contest.
- Start the WTI Server.
- Pause the contest clock.
- Use the WTI-UI to submit a run, or a clar.
This likely produces the problem.
- Set the contest time to "almost ended", then wait for the time to expire.
- Submit a run, or a clar.
This will produce the (unhelpful) error message.

**Expected behavior**:   
The system should display a non-cryptic error message, such as "Contest is not running".

**Actual behavior**: 
The system displays the cryptic error message shown above.

**Environment**: 
Windows 11, Chrome -- but it likely happens in all environments.

**Additional context**:  

WTI-UI class `new-run.component.ts` currently contains a method `isReadyToSubmit()` which is called in `new-run.component.html` to determine whether the `Submit` button should be activated.  `isReadyToSubmit()` should also check whether the contest is currently running.  (It  could potentially do this by invoking `ContestService.isContestRunning()` -- however, that service is not currently imported by the class.)

Also, class `new-run.component` contains the following:
```typescript
  private buildForm(): void {
    this.newSubmissionForm = this._formBuilder.group({
      problem: [undefined, [Validators.required]],
[...]

PR #1105, opened 2025-05-24
i1104: change run-submission error message.

clevengr:

Changes the error message displayed when a run is submitted after the contest has ended.  

Note: this also changes the error message displayed if a legitimate "Bad Request" occurs _during_ the contest.  This is a temporary fix applied for running NAC 2025; the issue needs to be fully addressed later.

Partial fix for Issue #1104.     Does not actually address the documented issue; rather, this push changes the error message displayed so that it's less confusing for the user.   Still need to develop code to actually fix the underlying problem described in the Issue.

Windows 11, java version "11.0.16.1" 2022-08-18 LTS

Follow the instructions in Issue #1104 for "Reproducing the issue"; verify that the error message returned from submitting a run after the contest has ended is "Not accepting submissions at this time".

### DIFF
--- a/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
+++ b/projects/WTI-UI/src/app/modules/runs/components/new-run/new-run.component.ts
@@ -141,7 +141,7 @@ export class NewRunComponent implements OnInit, OnDestroy {
 	        this._teamService.runsUpdated.next();
 	    },
 	    error: (error: any) => {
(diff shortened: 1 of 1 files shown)


kkarakas (approved):
LGTM. Good enough for a temporary solution.

johnbrvc (approved):
I agree with @kkarakas that this fix is sufficient for now and certainly better than what was there.  <sub><sup>someday</sup></sub>, of course, it should be addressed as described in the issue.

Issue #1103, opened 2025-05-23
The displayed order of submissions for WTI is wrong when they are submitted at the same minute.


kkarakas:
**Describe the issue**:As shown in the screenshot, when submissions are made within the same minute, they are ordered in increasing order. However, they should normally be ordered in decreasing order. This behaviour is observed when the user sends clarification.

**To Reproduce**:
Submit submissions/clarifications within the same minute. 

**Expected behavior**:
Correct order 

**Actual behavior**:
Wrong order

**Environment**: please complete the following information:
 - Windows 11:
 - Recent PC2:   
 - Chrome

**Screenshots**: 

![Image](https://github.com/user-attachments/assets/c6ec93fb-3f0a-49b8-ae45-3ee2642ca86f)

**Additional context**:

PR #1106, opened 2025-05-24
i1103: display order of runs and clarifications are changed to descending order

kkarakas:
The runs and clarifications are now in descending order by time. If the time is equal it is sorted by the id (descending).  Before if the runs or clarifications were sent at the same minute, they would be sorted in ascending order of id. 

#1103 fixes.
Windows11 Eclipse, Java 1.8

Log in as a team.
Send couple of clarifications within the same minute,
They should be ordered in descending order.
Log in as a judge and respond to these clarifications. 
They should be ordered right. 

Do the same action for runs.


johnbrvc (changes requested):
I would like to see both the Issue #1103 and the PR descriptions cleaned up.  I know these were entered "in the heat of the moment" during a live contest (NAC2025), but now that we have the time, please remove unnecessary stuff from the issue and make the issue and PR writeups more readable.

It should also be mentioned, as a CI, that you sort for both runs AND clarifications.  

Tested and appears to work - will approve once the descriptions are cleaned up.

johnbrvc (approved):
Changes made.  Approved.

PR #1115, opened 2025-06-09
i_1114 Put correct version of pc2sandbox_interactive.sh in repo

### LINKED ISSUE #1114: The pc2sandbox_interactive.sh script is out-of-date from what we used at recent contests
**Describe the issue**:
The version of the sandbox interactive problem execution script that is in git does not contain the latest version.   Somewhere along the way a change got lost that affects the behavior of the script if the interactive validator finishes before the submission.  Without the correct version of this script, some differences between a primary and shadow may occur if the "judge supplied" interactive validator fails for some reason.

**To Reproduce**: 
Compare the production version of the pc2sandbox_interactive.sh with what is in git.

**Expected behavior**: 
The version in git should be the version that we actually ran for recent contests (NAC2025 online practice, NAC2025 systests, NAC2025 Dress and NAC2025).

**Actual behavior**: 
The versions are different:

```
C:\ICPC\pc2v9\scripts>diff -b int.sh pc2sandbox_interactive.sh
607,610d606
<                       # Added this test 04/30/2025 - apparently it was missing.  We will only kill the submission if this
<                       # flag is 1, as per the comment at the top of this script.
<                       if test ${KILL_WA_VALIDATOR} -eq 0
<                       then
627,629d622
<                       else
<                               REPORT_DEBUG Killing child submission pid $submissionpid since it is still running "(KILL_WA_VALIDATOR=1)"
<                       fi
659c652
<                                       GenXML "wrong answer" \`head -n 1 $feedbackfile\`
\---
\>                                       GenXML "No - Wrong answer" \`head -n 1 $feedbackfile\`
661c654
<                                       GenXML "wrong answer" "No feedback file"
\---
\>                                       GenXML "No - Wrong answer" "No feedback file"

```

**Environment**: 
Ubuntu 24.04 (Linux systems only)

**Log Info**:  

**Screenshots**: 

**Additional context**:

johnbrvc:
Update repo with correct version of the interactive sandbox script.  A test was missing to kill a submission if the interactive validator finishes first.  Without this fix, we would sometimes notice JE (judging errors) or other judgment mismatches when comparing with a primary or shadow.  This fixes the so-called "DOMjudge" mode of interactive problem validation.

In addition, the judgment string was changed to the CLICS compatible "wrong answer" instead of "No - Wrong answer".

Fixes #1114 

Ubuntu 24.04 (Linux systems only)

The git diff included with this PR shows the missing code, however, it is better to look at the diff shown in the Issue #1114 since that diff ignores the different indenting levels.  The git diff shows the indenting differences, making it harder to read.

The version of `pc2sandbox_interactive.sh` included in this PR was used for all contests related to the NAC2025 (online, systests, dress and actual contest).

### DIFF
--- a/scripts/pc2sandbox_interactive.sh
+++ b/scripts/pc2sandbox_interactive.sh
@@ -604,22 +604,29 @@ do
 		REPORT_DEBUG Validator finishes with exit code $wstat
 		if test "$contestant_done" -eq 0
 		then
-			REPORT_DEBUG Waiting for submission pid $submissionpid to finish...
-			# Wait for child to finish - it has to, one way or the other (TLE or just finish)
-			wait -n "$submissionpid"
-			COMMAND_EXIT_CODE=$?
-
-			GetSandboxStats
-
-			if test $COMMAND_EXIT_CODE -eq 127
+			# Added this test 04/30/2025 - apparently it was missing.  We will only kill the submission if this
+			# flag is 1, as per the comment at the top of this script.
+			if test ${KILL_WA_VALIDATOR} -eq 0
 			then
-				REPORT_DEBUG No more children found.  Setting submission exit code to 0
-				COMMAND_EXIT_CODE=0
+				REPORT_DEBUG Waiting for submission pid $submissionpid to finish...
+				# Wait for child to finish - it has to, one way or the other (TLE or just finish)
+				wait -n "$submissionpid"
+				COMMAND_EXIT_CODE=$?
+	
+				GetSandboxStats
+	
+				if test $COMMAND_EXIT_CODE -eq 127
+				then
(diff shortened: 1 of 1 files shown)


clevengr (approved):
I reviewed the code changes.  While I don't pretend to be completely familiar with all the details of this script, the changes look reasonable to me.  Since it's based on what was used at NAC25, seems reasonable to approve.

johnbrvc:
> I reviewed the code changes. While I don't pretend to be completely familiar with all the details of this script, the changes look reasonable to me. Since it's based on what was used at NAC25, seems reasonable to approve.

Changes made as discussed above, that is, use Integer.parseInt() to cause an exception of bad scores are present in the scoreboard.

PR #1130, opened 2025-06-22
Fix missing state information in scoreboard JSON report

### LINKED ISSUE #1109: The scoreboard.json produced by the Results Export Files report is missing data
**Describe the issue**: 
If the **Results Export Files report** is generated, the `scoreboard.json `produced does not contain values for the contest state.

**To Reproduce**: 
Generate a **Results Export Files** report.

**Expected behavior**: 
The `scoreboard.json `should have the contest state in it.
`{"time":"2025-05-25T20:02:29.062406577Z","contest_time":"1:30:10","state":{"started":"2025-05-25T13:31:59.004-04","frozen":"2025-05-25T14:31:59.004-04","ended":"2025-05-25T15:01:59.004-04","thawed":null,"finalized":"2025-05-25T16:00:00.099-04","end_of_updates":null,"paused":null,"resumed":"2025-05-25T13:31:59.004-04","contest_time":"01:30:10.162"}`

**Actual behavior**: 
The `scoreboard.json `does not have contest state in it.
`{"event_id":"b349d405-23e1-4067-a971-f8243bc88e24","time":null,"contest_time":"","state":{"finalized":"","started":"","thawed":"","frozen":"","ended":"","end_of_updates":""}`

**Environment**: 
All.

**Log Info**:  

**Additional context**:

johnbrvc:
The incorrect **CLICS** routine was being called to generate the JSON scoreboard report.  We should be using the newest **CLICS API** Scoreboard objects (2023-06). This fix involved adding Jackson deserialization capability to several of the CLICS **API202306** classes (an empty constructor and getter methods). The **FileComparisonUtilities** had to be changed to use the correct CLICS API class names.
CI: Removed unused argument "**controller**" from `API202306.CLICSScoreboard`
CI: Fix datatype error for the "**time**" property in` API202306.CLICSScore`
CI: Attempt to populate the primary and shadow results folders to speed up EOC procedures (no longer should be necessary to track down the results folder using an explorer window).

This bug was found during NAC when the shadow (DOMjudge) noticed we didn't include state information in the scoreboard JSON.  The CLICS API Endpoint returned the correct data, but the reports did not.

This is an easy PR to review.  Many of the changes are adding _getter_ methods to some classes as well as default constructors.

Fixes #1109 

Windows 11
java version "1.8.0_321"

1. Load a completed contest, or at the very least, a contest that has several correct submissions.
2. From an administrator client generate the: "**Results Export File**s" report.
3. Verify that the `scoreboard.json` produced is correct (contains a correct state object instead of all empty strings)
4. From an administrator client generate the: "**Scoreboard JSON**" report and save it someplace.
5. Open the saved file and notice that the state object in the JSON contains valid information as opposed to empty strings.

To test the CI regarding populating the _Primary_ and _Shadow_ folder edit fields:
Load a completed contest, such as **NAC2025**  or a contest that has a `results` folder in the CDP base folder (same level as the `config `folder).
From an administrator client, on the **Run Contest** tab, choose the **Results Compare** tab.
Notice that the folders for _Primary_ and _PC2_ are filled in.

It should be noted that the nomenclature chosen for the results folders should read _Primary_ and _Shadow_ as opposed to _Primary_ and _PC2_.  PC2 can be the primary and some other CCS can be the shadow.  These should be fixed at some point someday.

### DIFF
--- a/src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
+++ b/src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import edu.csus.ecs.pc2.core.StringUtilities;
 import edu.csus.ecs.pc2.core.Utilities;
 import edu.csus.ecs.pc2.core.standings.TeamStanding;
 
@@ -43,13 +42,14 @@ public class CLICSScore {
      * Fill in the properties for a team's score
      *
      * @param teamStanding The team's scoring information
+     * @throws NumberFormatException if bad scores are in the standings
      */
     public CLICSScore(TeamStanding teamStanding) {
         num_solved = Utilities.nullSafeToInt(teamStanding.getSolved(), 0);
-        total_time = Utilities.nullSafeToInt(teamStanding.getPoints(), 0);
+        total_time = Integer.parseInt(teamStanding.getPoints());
         if(num_solved > 0) {
             // Problem solution time is in minutes.
-            time = StringUtilities.getIntegerValue(teamStanding.getLastSolved(), 0);
+            time = Integer.parseInt(teamStanding.getLastSolved());
         }
     }
 
--- a/src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
+++ b/src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
@@ -1,9 +1,11 @@
 // Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.clics.API202306;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import edu.csus.ecs.pc2.core.StringUtilities;
 import edu.csus.ecs.pc2.core.Utilities;
 import edu.csus.ecs.pc2.core.standings.TeamStanding;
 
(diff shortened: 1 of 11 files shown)


src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
```diff
@@ -39,8 +49,20 @@ public CLICSScore(TeamStanding teamStanding) {
         total_time = Utilities.nullSafeToInt(teamStanding.getPoints(), 0);
         if(num_solved > 0) {
             // Problem solution time is in minutes.
-            time = teamStanding.getLastSolved();
+            time = StringUtilities.getIntegerValue(teamStanding.getLastSolved(), 0);
```
clevengr: I worry that "0" is not a good default for `getIntegerValue()` -- that would mean defaulting to "the team solved this at time zero".  (I realize this "should never happen" -- but what if it does, Tammy?)  It seems like if the "default value" were ever actually used, it's an indication of a substantial problem in the code.  It seems like it would be better to at LEAST log a warning, maybe a severe, error; maybe even throw an exception at that point...

clevengr (commented):
Reviewed `src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java`.  Left one comment/concern.

src/edu/csus/ecs/pc2/clics/API202306/CLICSScore.java
johnbrvc: I was (when I <del>wrote</del> copied this) more concerned about the `total_time `line 3 lines back.  It too defaults to 0.   `getIntegerValue` will throw an exception if the value is not a number.  If sets it to the default (0) if the string is `null `or empty.  I suppose we can change it to `Integer.parseInt(teamStanding.getLastSolved())` and have it cause an exception for all non-integer values; at least we'll have a log of what happened.  Probably the same for `total_time `?  If that is satisfactory, I'll make the change.

clevengr: Well, it appears that making such a change would at least avoid choosing a "last solved time" of zero if there's an error in the system, so yes that does seem better.  Although if `Integer.parseInt()` throws an exception, the value of `time` will still have its (Java-default) value of zero.  But at least we'd have an exception to look at, so... yes.

clevengr (approved):
I reviewed all 11 changed files.  I commented on one place I would suggest be updated; other than that, all the changes seem reasonable.   I built the system, loaded the NAC2025 contest into it, and verified that the `scoreboard.json` files generated by both the `ReportsExportFiles` and the `Scoreboard JSON` file reports contain valid "state" info rather than empty strings.   Everything seems to work; I approve the PR.

JoeTerlizzi (approved):
Tested with NAC2025 contest.  scoreboard.json contains valid states for state object.  Approved.

Issue #604, opened 2022-11-13
Must add CLICSValidator judgement to judgement list


lane55:
**Describe the issue**: 

Analysis required to determine the impact when CLICSValidator judgements do not
match judgements in judgement list

The loaded judgements likely do not include all the judgements
that the CLICSValidator can return.   This may cause non-matching
judgements with a primary CCS

The following entries had to be added to reject.ini
Wrong Answer|WA
No - Wrong answer|WA
Incomplete output|WA
Undetermined|W

**To Reproduce**: 

**Expected behavior**: 

**Actual behavior**: 

**Environment**: 

**Log Info**:  

**Screenshots**:  

**Additional context**: 

The full reject.ini was:
correct|AC 
compiler error|CE 
run error|RTE 
timelimit|TLE 
wrong answer|WA 
memory limit|MLE 
no output|NO 
output limit|OLE 
Wrong Answer|WA
No - Wrong answer|WA
Incomplete output|WA
Undetermined|WA

lane55:
Perhaps the CLICS validator should also return a judgement acronym to match against
judgement acronyms in model's list of judgements

clevengr:
This issue has the "Shadow" label attached, but it's likely not just a Shadow issue -- it could be a problem when PC2 is the Primary CCS as well.

PR #1134, opened 2025-06-28
i_604 CLICSValidator produces result string returned to user

johnbrvc:
This was a fix to accommodate WF and NAC rules of only returning "_Wrong Answer_" on things such as _Incomplete output_, Excessive _output_ and _Output format error_.  Before this patch, the actual details of the validator were returned to the user causing confusion (IE "_Incomplete Output_" instead of "_Wrong Answer_").  A more permanent and configurable fix is needed for this.  This patch was successfully used at NAC2025.

Partial fix for #604 
This is more of a temporary fix until we rework the way the internal CLICS validator maps error returns.

Windows 11
Ubuntu 24.04
java version "1.8.0_321"
This patch was already tested at NAC2025.
**Note that only one file changed and there are 100's of differences.  Just about all of them are related to extra spaces on lines.  The only lines of interest are: 1 and 101-106 where 3 lines were commented out and 3 lines added to replace them.
To not show white space differences in the git web interface:
On the **Files Changed** tab, select the gear icon next to **Jump To**
Click the **Hide whitespace** check box.
Press the **Apply and reload** button
**
image

1. Start a contest such as sumithello
2. Configure one of the problems to use the built-in CLICS validator
3. Start an autojudge that judges the problems in the contest
4. Start the WTI
5. Make a submission that generates excessive output using the WTI
6. Notice the judgment producedand returned to team client.  It should say _Wrong answer_ as opposed to _Excessive output_.

## Additional note to testers
If you test using the PC2 code prior to this PR fix being applied and you use the "sumithello" contest and submit a program which produces excessive output for the "hello" problem, you will get a "Wrong Answer" response (not "Excessive Output").  This is due to an issue with the way the CLICS Validator handles the "hello.ans" file associated with the "sumithello" sample contest, the result of which is that submitting a program that produces "excessive output" will actually return a "Wrong Answer" result (a separate issue will be filed for that).  A recommended test, if you are trying to verify the existence of the issue prior to verifying that the PR fixes the issue, is to submit a program that generates "incomplete output" (for example, a modified version of the samps/src/hello.java program which writes its output to stderr instead of stdout).  With the previous PC2 code, this should produce an "Incomplete Output" result, whereas with the code with the PR included it should produce "Wrong Answer".

### DIFF
--- a/src/edu/csus/ecs/pc2/validator/clicsValidator/ClicsValidator.java
+++ b/src/edu/csus/ecs/pc2/validator/clicsValidator/ClicsValidator.java
@@ -1,6 +1,6 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 /**
- * 
+ *
  */
 package edu.csus.ecs.pc2.validator.clicsValidator;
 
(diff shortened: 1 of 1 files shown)


clevengr (approved):
I reviewed the (3-lines) of code changes; they look like a match to what we did at NAC25.  I ran a runtime test and verified that whereas the code prior to the PR fix returns "Incomplete Output" for a submission that generates no output, the code with the PR fix returns "Wrong Answer" for the same submission.    I approve the PR.

PR #1141, opened 2025-07-05
i_1122 Create a catalog file during execution

### LINKED ISSUE #1122: Provide the ability for judges to get a detailed report of judge's submissions
This feature request is related to a shortcoming of the "Submit Samples" tab.
...

**Feature Description**:  
The "Submit Samples" tab on the administrator (and judge) thick clients allows the execution of some or all of the judge's submissions.  It provides a GUI to show which submission match the expected results, which ones don't, how much CPU time TLE's took, etc.  The problem is, there is no way to view this information AFTER the tab is closed without re-running all the submissions.

A way is needed to allow judges to review the results of submitting the samples when the contest system is no longer running.  Either a report or web site of some sort would be useful....

...

1) Rerun all the judge's submissions every time a judge wants to see the results and possibly take screen shots of the GUI output of the "Submit Samples" tab.
2) Look at the log files generated by the judges and create some sort of summary report (this is rather difficult if there are multiple judges running on different machines).

Currently, the WJI (web judge interface) provides web pages to view contestant submissions (actually, ANY submissions, for that matter).  If submissions could be tagged in some way to indicate they are from the judge's submission in the CDP, then a new web page could be created that shows the same detailed information as it does for contestant submissions, in addition to showing if any of the results are unexpected. 
...

**Additional context**: 
The judge's submissions need to be executed in the actual contest environment BEFORE the contest starts.  This allows the judges to choose accurate time limits and track down any compiler/environmental compatibility issues.  That is why this is an important feature.
...

johnbrvc:
The `src_catalog.json` file contains details about the contestant's source code, such as the original full path of the submission, whether the file is stored externally, etc. This is useful for generating specific web pages for the judge's submissions.  For example, one can look at the full path of the submission source and see if it's from the CDP (for example); if so, then we know they're judge's sample submissions.  

Fixes #1122 

Windows 11
java version "1.8.0_321"

1. Load a contest that has sample submissions from the judges in the CDP (ex. **NAC2025**)
2. Make sure the `pc2v9.ini` file has the `[admin].sampleSubmitPane` key set to true.  This will make the **Submit Samples** tab show up on the administrator's **Run Contest** tab.
3. Start the server.
4. Start an administrator.
5. Start the autojudges.
6. Start the contest.
7. Using the **Submit Samples** tab, filter by a single problem or two (so you don't have to wait for the system to judge all the judge's submissions.
8. Submit the judge's submissions and wait for them to be judged.
9. Look in the execute folder for one or more of the judgments (The execute folder will be named something like: `ex_6_A_atotientquotient_1_cpp_judge1`) for a file with the name `src_catalog.json`
10. It should contain detailed information about source file(s) for the submission.

### DIFF
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -2770,7 +2770,7 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
             // create submission catalog file
             BufferedWriter catlog = new BufferedWriter(new FileWriter(prefixExecuteDirname(SOURCE_CATALOG_FILENAME)));
             if(runFiles != null) {
-                runFiles.createCatalogJSON(catlog);
+                runFiles.createCatalogJSON(catlog, log);
                 catlog.flush();
             }
             BufferedOutputStream stdoutlog = new BufferedOutputStream(new FileOutputStream(prefixExecuteDirname(COMPILER_STDOUT_FILENAME), false));
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -4,11 +4,13 @@ package edu.csus.ecs.pc2.core.execute;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.FileReader;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.nio.file.FileAlreadyExistsException;
@@ -161,6 +163,11 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
      * Validator stderr filename.
      */
     public static final String VALIDATOR_STDERR_FILENAME = "vstderr.pc2";
+
+    /**
+     * Information about the submission source file
+     */
+    public static final String SOURCE_CATALOG_FILENAME = "src_catalog.json";
 
     /**
      * The default limit (in seconds) for validation of a single run (test case) of a submission.
@@ -2760,6 +2767,12 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
             String cmdline = substituteAllStrings(run, language.getCompileCommandLine());
             log.log(Log.DEBUG, "after  substitution: " + cmdline);
 
+            // create submission catalog file
+            BufferedWriter catlog = new BufferedWriter(new FileWriter(prefixExecuteDirname(SOURCE_CATALOG_FILENAME)));
+            if(runFiles != null) {
+                runFiles.createCatalogJSON(catlog);
+                catlog.flush();
+            }
             BufferedOutputStream stdoutlog = new BufferedOutputStream(new FileOutputStream(prefixExecuteDirname(COMPILER_STDOUT_FILENAME), false));
             BufferedOutputStream stderrlog = new BufferedOutputStream(new FileOutputStream(prefixExecuteDirname(COMPILER_STDERR_FILENAME), false));
 
--- a/src/edu/csus/ecs/pc2/core/execute/Executable.java
+++ b/src/edu/csus/ecs/pc2/core/execute/Executable.java
@@ -2768,10 +2768,10 @@ public class Executable extends Plugin implements IExecutable, IExecutableNotify
             log.log(Log.DEBUG, "after  substitution: " + cmdline);
 
             // create submission catalog file
-            BufferedWriter catlog = new BufferedWriter(new FileWriter(prefixExecuteDirname(SOURCE_CATALOG_FILENAME)));
             if(runFiles != null) {
+                BufferedWriter catlog = new BufferedWriter(new FileWriter(prefixExecuteDirname(SOURCE_CATALOG_FILENAME)));
                 runFiles.createCatalogJSON(catlog, log);
-                catlog.flush();
+                catlog.close();
             }
             BufferedOutputStream stdoutlog = new BufferedOutputStream(new FileOutputStream(prefixExecuteDirname(COMPILER_STDOUT_FILENAME), false));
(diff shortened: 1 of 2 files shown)


src/edu/csus/ecs/pc2/core/execute/Executable.java
```diff
@@ -2760,6 +2767,12 @@ protected boolean compileProgram() {
             String cmdline = substituteAllStrings(run, language.getCompileCommandLine());
             log.log(Log.DEBUG, "after  substitution: " + cmdline);
 
+            // create submission catalog file
+            BufferedWriter catlog = new BufferedWriter(new FileWriter(prefixExecuteDirname(SOURCE_CATALOG_FILENAME)));
+            if(runFiles != null) {
+                runFiles.createCatalogJSON(catlog, log);
+                catlog.flush();
+            }
```
clevengr, line 2775: Shouldn't there be a `catlog.close()` here?

src/edu/csus/ecs/pc2/core/model/RunFiles.java
```diff
@@ -178,16 +182,88 @@ public String toString() {
...
+     */
+    public void createCatalogJSON(BufferedWriter catlog, Log log) {
+        try {
+            boolean needComma = false;
+            catlog.write("[");
+            if(mainFile != null) {
+                catlog.newLine();
+                catlog.write(" {\"mainfile\":");
+                catlog.write(serializedFileToJSON(mainFile));
+                catlog.write(" }");
```
clevengr: I don't see the need for a space in front of the `}` in the above line...

src/edu/csus/ecs/pc2/core/model/RunFiles.java
```diff
@@ -178,16 +182,88 @@ public String toString() {
...
+                needComma = true;
+            }
+            if(otherFiles != null && otherFiles.length > 0) {
+                boolean needComma2 = false;
```
clevengr: I don't see any reference/use of `needComma2` in the code.  Am I missing something?

src/edu/csus/ecs/pc2/core/model/RunFiles.java
```diff
@@ -178,16 +182,88 @@ public String toString() {
...
+     *     and create a JSON string.  Ex.
+     *     { "absolutePath":"C:/home/icpc/pc2/current/config/atotientquotient/submissions/time_limit_exceeded/evouga-brute.cpp",
+     *       "name":"evouga-brute.cpp",
+     *       "fileType":2,
+     *       "externalFile":false
+     *     }
+     * @param f the serialized file
+     * @return JSON string representation of the detailed info
+     */
+    private String serializedFileToJSON(SerializedFile f) {
```
clevengr: The name of this method seems a bit misleading; it implies that it's creating a JSON string for the serialized file (whereas what it's actually doing is creating a JSON string for the "detailed info" of the serialized file).

clevengr (commented):
I reviewed the code and left minor comments; nothing critical.  I pulled the code from the author's fork, built the system locally, and ran the tests described in the PR.  I see what appears to be correct `src_catalog.json` files in the selected (filtered) problems.  I also see submissions for exactly the submission files contained in the selected problems' `submissions` folders.

However, I do not see "expected results" in the "Status" field on the Submit Samples pane, and I likewise don't see expected judgements on the Runs screen.  Rather, except for the couple that got CE (which are explainable due to the different platform), all the submissions ended up in the "NEW" state.  Somethin is wrong somewhere, possibly (probably?) with my configuration.

clevengr (approved):
After struggling greatly with my local configuration, I finally got it correct and was able to properly execute the PR.  I get mismatches in the actual judge submission "Expected results" vs. "Status" -- compiler errors for one (which isn't surprising at all), plus several TLEs where "AC" was expected.  But again, those are explainable.  

And none of that is actually what the PR is about; it's about producing the proper `src_catalog.json` files in the execute directory.  I verified that this works, so I'm approving the PR.

src/edu/csus/ecs/pc2/core/execute/Executable.java
JoeTerlizzi: Use try-with-resources to ensure close is called even if an exception occurs.

The try-with-resources statement in Java is a language construct introduced in Java 7 to simplify resource management, especially for resources that need to be closed after use (e.g., file streams, database connections, network sockets). It ensures that each resource declared within the try statement's parentheses is closed automatically when the try block exits, regardless of whether an exception occurs.

src/edu/csus/ecs/pc2/core/model/RunFiles.java
JoeTerlizzi: Agreed.  Remove this line.

JoeTerlizzi (approved):
I ran with NAC2025 contest; Resubmitted solutions to be autojudged; src_catalog.json had the correct abs path for source code.

I agree with @clevengr RE his two comments and left my own feedback there.  Assuming these are addressed...approved.

src/edu/csus/ecs/pc2/core/execute/Executable.java
johnbrvc: Changed `catlog.flush()` to `catlog.close()`.
I don't think we need the extravagance of the _try-with-resources_.  If it fails to create the file, the we don't have to clean anything up, and it will exception itself out up the chain (which is fine).  If the write fails (`createCatalogJSON()`), that catches it and deals with it gracefully.

src/edu/csus/ecs/pc2/core/model/RunFiles.java
johnbrvc: I just had it there for readability of the json by a human.  I can live without it.  Removed.

src/edu/csus/ecs/pc2/core/model/RunFiles.java
johnbrvc: This was a bug in my loop.  The loop should have used `needcomma2 `not `needcomma`.   Basically, if there's more than one "other file", then we have to put commas between those.  The other files are in a separate json object, so they have their own separators (`needComma2`).

I fixed the loop.

src/edu/csus/ecs/pc2/core/model/RunFiles.java
johnbrvc: Changed to `serializedFileInfoToJSON()`.

Issue #1006, opened 2024-09-16
Add ability to support "Scoring" contests


johnbrvc:
**Is your feature request related to a problem?** 
This is not a problem but development for future contests.

**Feature Description**: 
Currently, PC2 only supports "pass-fail" type contests, that is, contests where a submission either completely works on all judge data (accepted), or fails on at least one judge test case.   "Scoring" contests give different number of points per submission depending on how many of just test cases are solved.  It is desired to add this "scoring" type contest ability to PC2.

**Have you considered other ways to accomplish the same thing?** 
It is not currently possible to support "scoring" contests.

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** 
Several subsystems within PC2 have to be modified. 

1. Reading of the problem data
2. Judging a submission
3. Scoreboard
4. (There are other's as well, I am sure).

**Additional context**: 
May be necessary to implement this for the NAC challenge.

PR #1143, opened 2025-07-10
i1006: add "pointscoring" sample contest (v.1)

clevengr:
Adds to the PC2 `samps/contests` folder a new contest named "pointscoring".  The initial version of this contest has three (identical) problems in it, all based on the PC2 samps "Sumit" problem.  Currently, "problem a" in the sample contest is configured to use only Problem Package Specification defaults for scoring -- including that it does not contain any sub-folders (other than just `sample` and `secret`) beneath the `data` folder.  

Problems 'b' and 'c' are currently just duplicates of problem 'a'; the intent is eventually to update those problems (and add more if necessary) to use a variety of other "point scoring" configuration settings/flags.

Partial work towards #1006.

Windows 10, Eclipse Version: 2020-12 (4.18.0), Java version "1.8.0_271"

- start a PC2 Server using `--load pointscoring`
- start a PC2 Admin.
- Verify that the loaded contest is configured with three problems, named "Sumit a", "Sumit b", and "Sumit c".
- Edit the "Sumit a" problem; examine the "Test Data Files" tab and ensure that there are three data file sets -- one from "sample" and two from "secret".


samps/contests/pointscoring/config/contest.yaml
```diff
@@ -0,0 +1,112 @@
+#
+#
+--- 
+
+name: Point Scoring Contest  
+short-name: pointscoring
+
```
johnbrvc, line 7: `contest.yaml `should have a "`scoreboard_type:`" property, either "`pass-fail`" or "`score`".  This is how the CCS knows how to score problems.  [Reference is spec for contests](https://ccs-specs.icpc.io/2023-06/contest_api#contests)  Note, however, if you set it to "score" at the moment, you will receive an Invalid configuration saying that "score" is Invalid only "pass-fail" is allowed (`ContestSnakeYamlLoader.java`, around line 616.  Maybe just add it as `pass-fail` now, with a comment saying we have to change it to score once `ContestSnakeYamlLoader` is fixed.

samps/contests/pointscoring/config/a/problem.yaml
```diff
@@ -0,0 +1,12 @@
+# Point Scoring problem configuration, problem 'a', version 1.0 
+--- 
+
+name: a
+type: scoring
```
johnbrvc, line 5: While it is harmless to have the `type: scoring` in the `problem.yaml`, and, it part of the program package specification, it is of no use for a CCS since the `contest.yaml `file specifies if the contest is "score" or "pass-fail" using the `scoreboard_type` property.  Tou can't have a mixed contest (pass-fail and score).  If the `contest.yaml` says it is **score**, then all problems are assumed to be **score**.

johnbrvc (changes requested):
Other than adding the scoreboard_type: property to contest.yaml, and noting the meaningless type: in the problem.yaml, this looks fine.
Once the scoreboard_type is added, I can approve this.

samps/contests/pointscoring/config/contest.yaml
clevengr: Fixed in latest push.  Set scoreboard_type to "pass-fail" in `contest.yaml`, with a comment that it needs to be changed to "score" once ContestSnakeYAMLLoader is updated.

samps/contests/pointscoring/config/a/problem.yaml
clevengr: I guess there is a "slight" justification for having `type: scoring` in a `problem.yaml` file:  problems get put in archives separate from the contest that originally used them; if someone later pulls the problem out of an archive then it's useful to have a flag that makes it clear that "this problem is expected to use point scoring".    (That doesn't mean it's particularly useful in a PC2 sample contest, but I think I'll leave it there just for completeness...)

johnbrvc (approved):

PR #1146, opened 2025-07-11
i_1006 Remember scoreboard type in ContestInformation

johnbrvc:
Allow "**score**" scoreboard type in addition to the standard "**pass-fail**".   Save the scoreboard type in the contest's `ContestInformation`.
Update CLICS contest API endpoint to use the scoreboard type defined in `ContestInformation` instead of hard-coding **pass-fail**.
CI: `ContestInformation` was missing compares for many members in the `isSameAs() `method, so if any of these were different, `isSameAs()` would not detect that.
CI: Add string constants for contest scoreboard type.

Partial fix for #1006 

Windows 11

Start fresh (no `profile `or` profiles.properties`)
Start a server loading the `samps/pointscoring` contest
It should load correctly (note that is uses `scoreboard_type: score`)


src/edu/csus/ecs/pc2/clics/API202306/CLICSProblem.java
```diff
@@ -82,15 +93,45 @@ public CLICSProblem(IInternalContest model, Problem problem, int ordinal) {
         }
         test_data_count = problem.getNumberTestCases();
         time_limit = problem.getTimeOutInSeconds();
+        if(model.getContestInformation().isScoreboardTypeScore()) {
+            isPointScoring = true;
+            ProblemDataFiles problemDataFiles = model.getProblemDataFile(problem);
+            if(problemDataFiles != null) {
+                TestDataGroup [] testDataGroups = problemDataFiles.getJudgesDataGroups();
+                // Really, there's only 1 top level TestDataGruop
```
clevengr: Nit: typo in comment (TestDataGruop -> TestDataGroup).

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
```diff
@@ -0,0 +1,110 @@
...
+        Account accountB = cachedAccountList.getAccount(teamB.getClientId());
+        nameB = accountB.getDisplayName();
+        bClientHash = teamB.getClientId().hashCode();
+        nameComparison = accountNameComparator.compare(nameA.toLowerCase(), nameB.toLowerCase());
+
+        //
+        // Primary Sort = score (high to low)
+        // Secondary Sort = earliest submittal of last submission (low to high)
+        // Forth Sort = teamName (low to high)
+        // Fifth Sort = clientId (low to high)
```
clevengr: I don't understand why these two comments refer to "Forth [sic] Sort" and "Fifth Sort".  The two previous lines refer to "Primary [i.e. First] Sort" and "Secondary Sort".  So the next one should be "Third Sort" and the last one "Fourth Sort"... right?

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
```diff
@@ -0,0 +1,110 @@
...
+            status = 0; // elements equal, this shouldn't happen, Tammy...
+        } else {
+            if ((bScore > aScore)
+                    || ((bScore == aScore) && (bLastSolvedTime < aLastSolvedTime))
+                    || ((bScore == aScore) && (bLastSolvedTime == aLastSolvedTime) && (nameComparison > 0))
+                    || ((bScore == aScore) && (bLastSolvedTime == aLastSolvedTime)
+                            && (nameComparison == 0) && (bClientHash < aClientHash))) {
+                status = 1; // a greater then b
+            } else {
+                status = -1; // a less then b
```
clevengr: These return values seem backwards to me.  A comparator is supposed to return negative if the first comparand (A) is less than the second comparand (B), and positive if the first comparand (A) is greater than the second comparand (B).  The `if` clause (at line 90) starts with the condition `if ( (bScore > aScore) || ... `.  So if bScore is indeed greater than aScore, the rest of the conditions in the `if` will be ignored and the code will execute the `then` part of the `if` -- meaning it will return `1`.  But a return of `1` is supposed to indicate that `aScore > bScore` (as it says in the comment on line 95).   I think the status assignments are backwards (or something).

Also, minor nit:  the comments  `//a greater then b` and `a less then b` should use "than" rather than "then".

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -245,6 +247,72 @@ private ProblemSummaryInfo calcProblemScoreData(TreeMap<Run,Run> treeMap, IInter
         return problemSummaryInfo;
     }
 
+    /**
+     * Get the Point Scoring and Statistics information for one problem.
+     *
+     * @return pc2.ex.ProblemScoreData
```
clevengr: This `@return` refers to a class in a non-existent package (`pc2.ex`).

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -245,6 +247,72 @@ private ProblemSummaryInfo calcProblemScoreData(TreeMap<Run,Run> treeMap, IInter
...
+     * @param treeMap
+     *            java.util.TreeMap
```
clevengr: It would be nice if the `@param` explained what the `treeMap` is expected to contain...

src/edu/csus/ecs/pc2/core/util/XMLMemento.java
```diff
@@ -1,11 +1,11 @@
-// Copyright (C) 1989-2019 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-20259 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
```
clevengr: Typo in updated copyright date (20259 -> 2025)

clevengr (changes requested):
I have reviewed all but one of the (20) changed files in this PR (the changes to `DefaultScoringAlgorithm` are extensive and I need more time to look at those).

  I noted once place where I think there is a logic error (in `DefaultPointScoringStandingsRecordComparator.java`).  Other than that I made only minor comments, which I'd like to see considered and addressed, but wouldn't stop me from Approving the PR.  
  
  Given that there's no practical way to TEST this PR until the entire Point-Scoring pipeline implementation is done, I would simply approve the PR once the comments I've made are addressed.

src/edu/csus/ecs/pc2/clics/API202306/CLICSProblem.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
johnbrvc: Correct.  When I copied this from DefaultStandingsRecordComparator, there was a third criteria (# solved) which we dont use for PS; I forgot to update the comment and fix the spelling error.  

Fixed in next upsh.

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
johnbrvc: The sorting algorithm sorts things from "low to high" by default (ascending).  The comparators return a value as to whether the first thing should be considered(!) less-than, equal to or greater than the 2nd thing.  That is, does the first thing appear AFTER the second thing in the sorted result.   So, for Point scoring, the guy with the biggest score should be the first thing in the  result.  When comparing 2 things (ascore and bscore), we have to determine if we want ascore to come after bscore.  (IE if ascore is smaller than bscore, then a comes after b, and should be consider that "a" is bigger than "b" in terms of how the sorting is done, so we return 1, indicating that a should come after b in the list. )

If you look at DefaultStandingsRecordComparator.java, (which this was copied from) that uses a1,a2,a3,a4,a5 and b1,b2,b3,b4,b5.  I changed these to what they really are: aScore, bScore, aLastSolvedTime, bLastSolvedTime, etc. to make reading the compound "if" statement easier.

So, what the compound "if" is testing is whether "A" should come after "B" in the score (return 1), they're equal (return 0), or "A" should come before "B" in the sort (return -1).   if bScore > aScore, then "A" should come AFTER "B" (return 1).  If aScore and bScore are equal, we check the LastSolvedTime.  "A" should come after "B" if  "A"'s LastSolvedTime is > "B"'s.  SImilarly for name.

What we could do, and make it a whole lot more complicated and bug ridden, is to use separate comparators for each property and chain them, and make "score" be ".reversed()".... Nah...

I fixed "then" -> "than", and left the compare as-is.  I also changed the comments to read:
```
                status = 1; // a to be considered greater than b
            } else {
                status = -1; // a to be considered less than b
            }

```
Next push.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: Yikes.  That must have been WAY before my time.  (I fixed calcProblemScoreData's comment as well while I was at it.)
Next push.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: Haha.  I added that when I fixed the comment.
Next push.

src/edu/csus/ecs/pc2/core/util/XMLMemento.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
clevengr: Ok, I *think* I understand the logic.  However, the standard JavaDoc explanation of the function of a comparator, which is copied into the top of the code (probably as legacy from the original version) not only isn't helpful in understanding the logic -- it is almost counter-intuitive.   It might be helpful if some (abbreviated) version of the explanation of this logic was present -- perhaps a full-line or two comment right at the point where `status = 1` and `status = -1` is assigned.  Otherwise, we're (someone) is going to have this same conversation again in the future -- and someone might decide (based on the existing JavaDoc) like I did, that the logic is incorrect (backwards).

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -245,6 +248,73 @@ private ProblemSummaryInfo calcProblemScoreData(TreeMap<Run,Run> treeMap, IInter
...
+                        unJudgedRun = true;
+                    }
+                }
+            }
+        }
+        // TODO put another if around this if there was a setting to include all
+        // no's before yes
+        if (!solved) {
+            score = 0.0;
+        }
```
clevengr, line 308: This appears to be making a hard-coded decision that if the problem is not solved then the score is 0.0.  It seems like there could be situations where the Grader could have assigned "some points" (for example, due to having passed "some tests" but not all) even if the problem wasn't "entirely solved".  Shouldn't the score assignment here be coming from some input somewhere (like, the JudgementRecord) rather than being hard-coded to be zero?

src/edu/csus/ecs/pc2/core/scoring/DefaultPointScoringStandingsRecordComparator.java
johnbrvc: Added comment, mostly from my description above, but trimmed down a bit.

Just pushed.

clevengr (approved):
I have re-reviewed all changed files, and also completed a cursory review of `DefaultScoringAlgorithm`.  I'm not sure about all the changes in `DefaultScoringAlgorithm`; they look reasonable but there are a lot of details in the changes.  That said, and given that there's no way to do a run-time test of the PR, I'm approving the PR so we can move forward.

johnbrvc:
This PR has been merged already to facilitate further development.  Both the **i1006_point_scoring** and **i1006_update_prs** branches have this code.  This PR will be kept open, however, for future review.

PR #1148, opened 2025-07-17
i_1147 Remove html2text dependency in pc2submit

### LINKED ISSUE #1147: pc2submit should not depend on html2text python package
**Describe the issue**: 
The WF image does not include the python html2text package, and pc2submit made use of it.  The dependency
on this package should be removed, especially since it is not necessary. 

**To Reproduce**: 
Try running pc2submit on the contest image.

**Expected behavior**: 
The command should not produce an exception about a missing package: html2text

**Actual behavior**: 
An exception about a missing package is produced: html2text

Ubuntu 24.04 contest image.

**Log Info**:  
```
$ ./pc2submit
Traceback (most recent call last):
  File "/home/john/ICPC/icpc-image-building/files/./pc2submit", line 34, in <module>
    import html2text
ModuleNotFoundError: No module named 'html2text'

```

**Screenshots**:  

**Additional context**: 
This package was only used in the rare case where a "FORBIDDEN" HTML status was returned.  These cases have been fixed by more descriptive return status codes in PC2, so the default string of "Forbidden" is fine now.

johnbrvc:
The python pc2submit utility no longer needs html2text and it has been removed since it is not included in the WF contest image.

Fixes #1147 

Ubuntu 24.04
Windows 11

Attempt to run pc2submit on the WF contest image.
It should no longer produce an exception about a missing html2text package.

### DIFF
--- a/bin/pc2submit
+++ b/bin/pc2submit
@@ -31,7 +31,6 @@ import netrc
 import getpass
 import io
 import zipfile
-import html2text
 
 from pathlib import Path
 from typing import NoReturn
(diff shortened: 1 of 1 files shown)


kkarakas (approved):
Read through the code, did not run it. LGTM.

PR #1161, opened 2025-08-17
i_606 Limit source code submission size

### LINKED ISSUE #606: Add source file limit

Yes, team's are allowed to submit HUGE source files

**Feature Description**: 

Add a global source file limit (in K)

Return a CE ??  if a team submits a file that is too large

**Additional context**:

johnbrvc:
Add `ContestInformation `member for maximum source code size (defaults to 256KiB - the WF limit, 0 means unlimited).
Change **Settings** GUI pane to allow setting max source size.
Change snake yaml loader to read value from `system.pc2.yaml`.
Enhanced `SubmissionRejectedException` to have a "rejection type" (too many submissions or source too big).
Use `SubmissionRejectionException `rejection type to set appropriate HTML status in WTI `TeamsController `(**TOO_MANY_REQUESTS** or **REQUEST_ENTITY_TOO_LARGE**)
Change WTI typescript **new-run component** to check for the HTML status and use the error message provided by PC2 if its one of the 2 above.
Add code in all places where a submission is accepted to calculate the size of the submission and compare against the max accepted size from `ContestInformation`. (CLICS API, thick team client, WTI API client)

Fixes #606 

Windows 11
Ubuntu 24.04
java version "1.8.0_321"

This procedure appears to be quite long, and it took about 90 minutes to write it up, so the least you can do is follow it through, thank you.  It does not take that long to follow this procedure, and it is mostly repetitive.  It goes through the three different methods of making a submission into the pc2 system, and confirming that the source limiting works.

Prerequisites:
1) You'll need Python 3.10 installed on your PC.   You can test this from a command prompt by typing: `py --version`
Installing an configuring Python for your system is beyond the scope of this PR.  
   1.1 Added note:  during testing on Windows 11 I found that Python 3.11 worked just fine as well.  However, initially I was getting Python compilation errors on the `import requests` statement.  I found that this was because the `requests` module is not installed in Python (PyPy) by default.  I used `python -m pip install requests` to upgrade my Python.  This in turn led to a `pip not found` error, so I had to first install pip. (JLC)
2) To build the PC2 distribution AND the _Web Team Client_ (which you'll need for testing) in the Eclipse _Project Explorer_ (left pane), find "`build.xml"`, right click and choose **Run As**->**Ant Build...**  Select only **jarSkipTests**, press **Apply** then **Run**.  This will build PC2 AND the _Web Team Client_.  If you're inclined to waste time, feel free to build the entire package using the `package.xml` - I don't have time to waste, so I don't do that.

1) Starting from scratch, load the **sumithello** contest.
2) Start the **administrator1** client.
3) On the **Configure Contest**->**Accounts** tab, press **Generate** and create 1 **FEEDER** account (you'll need this for testing _CLICS_ command line submits.
4) Edit the newly created **feeder1** account and add the following permissions: "**Shadow Proxy Team**", "**Submit a run**"
5) On the **Configure Contest**->**Settings** tab, scroll down to the "**Team Settings**" section, and set the max source size to 8  (this sets it to 8KiB for testing).
6) In the **Remote CCS Settings**, set **Enable CCS Test Mode**
7) Press the **Update** button.  Make sure that both the maximum source size is set to 8 and the Enable CCS Test Mode button is checked.
8) Rename this attached file [hello.cpp.txt](https://github.com/user-attachments/files/21834001/hello.cpp.txt) to `hello8k.cpp` in some directory of your choosing.  Remember where you extracted it to, as you'll need this below.
9) Start an event feeder client and log in as feeder1 (ef1) using password feeder1
10) Press the **Start** button the main **Web Services** tab.  This will start the CLICS 2023-06 API hander.
11) Open a web browser and navigate to: [https://administrator1:administrator1@localhost:50443/contests/SumH/problems](https://administrator1:administrator1@localhost:50443/contests/SumH/problems)  You should see a JSON description of the 2 configured problems, **hello** and **sumit**.  If not, **STOP** you have done something wrong, no point in proceeding.  Check your work.
12) In _Windows Explorer_, navigate to your repo's `pc2v9/projects` folder.  There you will see the `WebTeamInterface-1.2-.zip` file.  Make sure the date of that file reflects the date/time you built it.  If not, go back to the prerequisites above.
13) Right click on the `WebTeamInterface-1.2-.zip `file and Extract All to a folder of your choosing.  This is where you'll run the WTI from.
14) From a command prompt, change (cd) to place you extracted the ZIP file to, eg.  <some path>/WebTeamInterface-1.2
15) Type: bin\pc2wti  This will start the WTI.  Make sure there are no errors.  You can minimize that command prompt window to eliminate desktop clutter.
16) From the administrator client, select **Configure Contest**->**Times**, and press the **Start All** button to start the contest.
17) In your web browser, enter: [http://localhost:8080](http://localhost:8080)  The WTI login screen should come up, log in using **team1** with password **team1**
18) Press the orange **Submit Problem** button, select problem "**Hello**" and Language "**GNU C++**".
19) Press the **Main File** big button, and navigate to where you copied the `hello8k.cpp` file above.
20) Press the orange **Submit** button.  You should receive a red error message that looks like this:
image

21) On the administrator client, go back to the **Settings** tab and change the maximum source size to 0 (this is unlimited size), and press **Update**.
22) Try the same submission again from the WTI.  This time it should work.
23) On the administrator client, go back to the **Settings** tab and change the maximum source size to 256 (this is 256KiB), and press **Update**.
24) Try the same submission again from the WTI.  It should work this time as well.
25) On the administrator client, go back to the **Settings** tab and change the maximum source size back to 8, and press **Update**.
26) Open a new command prompt and change to your repo's pc2v9\bin folder.
27) Now you'll submit the same hello8k.cpp file using the CLICS command line utility by typing: `py pc2submit -u https://team1:team1@localhost:50443 -p hello -y <your-path-to>/hello8k.cpp`
28) You should get an error indicating the submission is too large:
```
C:\ICPC\pc2v9\bin>py pc2submit -u https://team1:team1@localhost:50443 -p hello -y c:/tmp/hello8k.cpp
Submission failed (code 413 - Submission too large. Maximum allowed size is 8192 bytes.)
```
Depending on how long this procedure takes, you may see a warning that the source file is older than some time:
`Warning: c:/tmp/hello8k.cpp' has not been modified for 43 minutes!'
`You can ignore this - it is to let you know you may be submitting something that is old.  It just means you are slow.
29) You can then fiddle around with the maximum source size setting in the administrator and make additional command line submissions to test if 0 and 256 or other values work correctly.
30) Next, we'll test the thick **pc2team** client.  First, make sure the maximum source size setting is back to 8.
31) Start the **pc2team** client and login using **team1** with password **team1**
32) On the **Submit Run** tab, select the **Hello** problem, and Language **GNU C++**.  For the **Main File**, select the `hello8k.cpp `file.
33) Press the **Submit** button.
34) You should receive a pop-up dialog:
image

35) Now, you can change the maximum source size setting to various values like 0 and 256 to test that the source limit works.


src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -400,59 +401,72 @@ public void submitJudgeRun(IProblem problem, ILanguage language, String mainFile
...
                 throw new Exception("File '" + additionalFileNames[i] + "' no such file (not found)");
             }
         }
 
+        long sourceSizeByteLimit = internalContest.getContestInformation().getMaxSourceSizeInBytes();
+        if(sourceSizeByteLimit > 0 && sourceSize > sourceSizeByteLimit) {
+            String sizeMsg = "Source file(s) are too large (" + sourceSize + " bytes) - maximum is " + sourceSizeByteLimit + " bytes.";
+//            I wondered why no logging is ever done is this module, so I won't enable this -- JTB
+//            LogUtilities.ensureStaticLog();
+//            StaticLog.getLog().log(Level.WARNING, sourceSizeByteLimit);
```
clevengr: I realize this is commented out, but:  if someone were to uncomment it, it appears to me that it would be logging a very confusing message:  a single (long) integer.  It should probably have a message describing the situation better.

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -518,49 +532,63 @@ public void submitJudgeRun(IProblem problem, ILanguage language, IFile mainFile,
...
+                sourceSize += nextFile.getByteData().length;
             }
         }
 
+        long sourceSizeByteLimit = internalContest.getContestInformation().getMaxSourceSizeInBytes();
+        if(sourceSizeByteLimit > 0 && sourceSize > sourceSizeByteLimit) {
+            String sizeMsg = "Source file(s) are too large (" + sourceSize + " bytes) - maximum is " + sourceSizeByteLimit + " bytes.";
+//            I wondered why no logging is ever done is this module, so I won't enable this -- JTB
+//            LogUtilities.ensureStaticLog();
+//            StaticLog.getLog().log(Level.WARNING, sourceSizeByteLimit);
```
clevengr: Same comment about this comment: I think the second log() parameter should be a string message...

src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
```diff
@@ -7,32 +7,52 @@
...
     /**
      * Constructs a SubmissionRejectedException containing text message.
-     * 
-     * @param message A text message associated with the Exception, typically the reason for the Exception.
+     *
+     * @param message A text message associated with the Exception using the default reason
      */
     public SubmissionRejectedException(String message) {
         super(message);
+        rejectReason = SubmissionRejectionReason.THROTTLE_EXCEEDED;
```
clevengr: It seems weird to me (slightly wrong, in fact) that the "Default Reason" (as assigned by this constructor) is `THROTTLE_EXCEEDED`.  Code which called the previous no-argument constructor would get an object that has a default "rejectReason" of `UNKNOWN`.  That seems like a more logical default, and it seems illogical to have two different default values for an equally-unspecified attribute value.  (In other words, seems like defaults should be consistent, and I think `UNKNOWN` is the most logical choice.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -717,6 +718,21 @@ public IInternalContest fromYaml(IInternalContest contest, String[] yamlLines, S
...
 
+        Object maxSourceSize = fetchObjectValue(content, MAX_SOURCE_SIZE_K_KEY);
+        if (maxSourceSize != null) {
+
+            if (maxSourceSize instanceof Integer) {
+                int maxSizeInK = ((Integer) maxSourceSize).intValue();
+                if (maxSizeInK > 0) {
+                    setMaxSourceSize(contest, maxSizeInK * Constants.BYTES_PER_KIBIBYTE);
+                } else {
+                    throw new YamlLoadException("Invalid max-source-size-K value '" + maxSourceSize + " size must be > 0 ", null, contestFileName);
```
clevengr: The message passed to the YamlLoadException constructor has an opening single-quote in front of the integer value of maxSourceSize, but there's no corresponding close single-quote.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -717,6 +718,21 @@ public IInternalContest fromYaml(IInternalContest contest, String[] yamlLines, S
...
+                }
+            } else {
+                throw new YamlLoadException("Invalid maxSourceSize value '" + maxSourceSize + " size must an integer", null, contestFileName);
```
clevengr: Same as previous comment:  missing closing single-quote.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
```diff
@@ -1051,6 +1067,14 @@ private void setMaxOutputSize(IInternalContest contest, int maxOutputBytes) {
         contest.updateContestInformation(contestInformation);
     }
 
+    private void setMaxSourceSize(IInternalContest contest, int maxSourceBytes) {
```
clevengr: It worries me that here, maxSourceBytes is being handled as an int (so, max value is 2^32), whereas in other locations it is being handled as a long.   While it _probably_ won't make a difference, the whole business about limiting source size is to deal with people who just _might_ submit something larger than 2^32 -- so it seems like the same data size (long) should be used throughout.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
```diff
@@ -660,6 +667,10 @@ private Component getTeamSettingsPane() {
             teamSettingsPane.add(getMaxOutputSizeLabel(), null);
             teamSettingsPane.add(getMaxOutputSizeInKTextField(), null);
             teamSettingsPane.add(getRigidArea1());
+            teamSettingsPane.add(getMaxSourceSizeLabel(), null);
+            teamSettingsPane.add(getMaxSourceSizeInKTextField(), null);
+            teamSettingsPane.add(getMaxSourceSizeWhatsThisButton(), null);
+            teamSettingsPane.add(getRigidArea1());
```
clevengr: The call to `getRigidArea1()` is returning the exact same object as the previous call to that same method -- meaning the same component is specified as appearing in two places.  I think this can cause layout problems.  Normally the approach is to define a separate method `getRigidArea<N>()` for each rigid area to be placed in the layout.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
```diff
@@ -772,6 +783,17 @@ private JLabel getMaxOutputSizeLabel() {
         return labelMaxOutputSize ;
     }
 
+    private JLabel getMaxSourceSizeLabel() {
+       if (labelMaxSourceSize == null) {
+
+           labelMaxSourceSize = new JLabel();
+           labelMaxSourceSize.setHorizontalAlignment(SwingConstants.RIGHT);
+           labelMaxSourceSize.setBorder(new EmptyBorder(0,10,5,5));
+           labelMaxSourceSize.setText("Default max source size (in KB): ");
```
clevengr: The label "Default max source size" seems inaccurate.  The _default_ is hard-wired into the code (I think it is 256K?)  It seems like this field is just specifying "Max source size (in KB)" as a value intended to _override_ the default, not _set_ the default.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
```diff
@@ -1682,6 +1708,62 @@ public void keyReleased(java.awt.event.KeyEvent e) {
...
+                    public void mousePressed(MouseEvent e) {
+                        JOptionPane.showMessageDialog(null, maxSourceSizeWhatsThisMessage, "About Maximum Source Size", JOptionPane.INFORMATION_MESSAGE, null);
+                    }
+                });
+                maxSourceSizeWhatsThisButton.setBorder(new EmptyBorder(0, 15, 0, 0));
+            }
+            return maxSourceSizeWhatsThisButton;
+        }
+
+    // the string which will be displayed when the "What's This" icon in the Judge Settings panel is clicked
```
clevengr: Is maxSourceSize really a _Judge_ setting?

clevengr (commented):
I have reviewed all of the source code changes.  I've made a number of comments, but there's nothing that appears to be a show-stopper.  I will try to perform an execution test ASAP.

src/edu/csus/ecs/pc2/api/ServerConnection.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/core/exception/SubmissionRejectedException.java
johnbrvc: This was done to avoid changing `InternalController` to pass the new parameter.  This way it is "backwards" compatible. I'll fix the comment in the constructor with the string indicating this is for historical use by the throttling code.  Changing `InternalController `to use a new constructor for `SubmissionRejectedException `would require a great deal more testing since `InternalController `is kind of an important module.  I can create an issue indicating that InternalController should be updated to use the new constuctor.

Comments updated and calling sequence changed to use overloaded constructor.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: Yeah, this was copied from the code just above for Max Output Size which has the same issue (which I just fixed as well as CI).  

Fixed in next push.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: Fixed in next push.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
johnbrvc: The object mapper returns an `Integer `object.  This value is from the `system.pc2.yaml` file and is in KiB.  I changed it to convert the `Integer `to a `long`, then multiply by number of bytes in a KiB.
As a side note, setMaxOutputSize (just above) has the same issue, which I did NOT change (out of scope).

Fixed in next push.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
johnbrvc: I'm not sure this is a problem since it's in the same sub-pane and not being resized or anything, but, I'll created a new one anyway.

Fixed in next push.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
johnbrvc: Again, this was modeled after the existing max output size, and it too has a default constant **DEFAULT_MAX_OUTPUT_SIZE_K** (512).  But, I'll change the wording for source size.  Fixing the max output size  message is out of scope.

Fixed in next push.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
johnbrvc: Copy/paste error.

Fixed in next push.

johnbrvc:
> I have reviewed all of the source code changes. I've made a number of comments, but there's nothing that appears to be a show-stopper. I will try to perform an execution test ASAP.

I made most of the suggested changes.  And here I thought there would be complaints with my pidgin type-script changes.

I also added an extensive test procedure to this PR description.  Enjoy.

src/edu/csus/ecs/pc2/imports/ccs/ContestSnakeYAMLLoader.java
clevengr: I agree that fixing maxOutputSize is out of scope for this PR.  I'm actually not quite as concerned about that one, since I think limiting output to 2^32 is probably ok.

src/edu/csus/ecs/pc2/ui/ContestInformationPane.java
clevengr: I'm pretty sure I've seen issues resulting from reusing the same component... but it might be that those issues were confined to displays in the WindowBuilder in Eclipse (I don't recall).  Anyway, thanks for fixing it.

clevengr:
> I made most of the suggested changes. And here I thought there would be complaints with my pidgin type-script changes.

The TypeScript changes seemed reasonable to me... what were you thinking might be a concern?

clevengr (approved):
I reviewed the code changes, which look good.  I ran the complete testing procedure listed in the PR (and added an extra note about testing when dealing with Python under Windows).  All tests passed, and everything looks good as far as I can tell.  I approve this PR.

JoeTerlizzi (approved):
I have followed the testing procedure and verified the expected results.  FWIW I used python 3.13.7.

Approved

PR #1167, opened 2025-08-24
i_1166 Better heuristics for guessing problem name

### LINKED ISSUE #1166: Submitting judge's sample solutions may use wrong problem
**Describe the issue**: 
If there are 2 problems where one's short-name is a substring of the other, eg.  "**bettor**" and "**bettor2**", the "**Submit Samples**" tab will only submit the shorter short-name (eg. **bettor**).  

**To Reproduce**: 
1. Start a contest and create 2 problems where one's short-name is a substring in the other one.
2. Make sure there are sample submissions in the `problem/submissions/accepted` folder of each.
3. Be sure to have the following key in your` pc2v9.ini `file so the **Submit Samples** pane will appear:
```
[admin]
sampleSubmitPane=true

```
4. Using the Administrator's **Submit Samples** pane, make a sample submission to the problem that has the longer short-name.  You can choose the problem you want by using the **Filter** button and enabling the Filter.
Note that the submission is shown as the problem with the shorter short-name.

**Expected behavior**: 
The correct problem should be submitted.

**Actual behavior**: 
The problem with the shorter short-name is submitted even if the longer short-name is chosen.

**Environment**: 
All.

**Log Info**: 
**Screenshots**:  
**Additional context**: 
This problem occurs because the `QuickSubmitter.guessProblem()` method just uses "`indexOf()`" to find a matching problem.  It should use other heuristics to better guess the problem name, such as the position of the problem name in the full pathname, eg.  `/ICPC/contests/wf49baku/contests/dress/config/bettor2/submissions/accepted/try.cpp`
or, it could search for `File.separator + 'problem-short-name' + File.separator`.  If either of those fail, then it can, as a last resort, just use the `indexOf `like it currently does.

johnbrvc:
Fixed `QuickSubmitter.guessProblem()`, which is ONLY used pre-contest when submitting judge's samples, so it better chooses the correct problem.  The fix involved adding 2 additional ways to guess the correct problem.  First, it will try looking at the expected pathname of the judge's submission, and it will isolate the problem short-name (id).  If that fails, it will try to look for `'/' + problem-short-name + '/'` in the pathname.  If that fails, it will just do what it did before, namely, just see if the problem short-name is anywhere in the pathname.

Also include an updated `pc2v9.ini `that will show the **Submit Samples** pane on the admin and judges.  (This is a good default anyway.)

Fixes #1166 

java version "1.8.0_321"
and
Ubuntu 24.04
Java 21.0.4

Follow the same steps in the issue #1166, but note that the correct problem is chosen.

### DIFF
--- a/pc2v9.ini
+++ b/pc2v9.ini
@@ -25,6 +25,13 @@ server=localhost:50002
 wtiport=8080
 wtiwsName=/websocket
 
+[admin]
+sampleSubmitPane=true
+
+[judge]
+sampleSubmitPane=true
+
+
 [clics]
 apiVersionsSupported=2023-06,2020-03
 # Comma separated list of notification types (EventFeedTypes) to not use collections on.
--- a/src/edu/csus/ecs/pc2/ui/team/QuickSubmitter.java
+++ b/src/edu/csus/ecs/pc2/ui/team/QuickSubmitter.java
@@ -186,8 +186,47 @@ public class QuickSubmitter implements UIPlugin {
      * @return
      */
     private Problem guessProblem(IInternalContest contest2, String absolutePath) {
+
         Problem[] problems = contest.getProblems();
+        // The problem short name will come before the submissions/ folder
+        String submissionsBase = File.separator + IContestLoader.SUBMISSIONS_DIRNAME + File.separator;
+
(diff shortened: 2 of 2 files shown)


kkarakas (approved):
LGTM!

PR #1173, opened 2025-09-19
I1170 handle zero-length WTI submission files

### LINKED ISSUE #1170: Zero-length file in WTI submission returns illogical error message
**Describe the issue**: 
  If a WTI user submits a main program file which is empty (zero-length), the WTI produces an error message like this:

Image

This message makes no sense; there's no way for the user to understand what has actually gone wrong.

**To Reproduce**: 
- Start a PC2 Server and load a contest (e.g. with `--load sumithello`). 
- Start the contest clock running.
- Start the WTI Server. 
- Open a browser and login to the WTI as a team.
- Select "Submit Run".
- Choose a problem and a language.
- Create a file of zero length and select that as the "Main File".
- Click "Submit"

**Expected behavior**: 
The response should contain a reasonable error message, such as "Submitting zero-length files is not allowed". 

**Actual behavior**: 
The response is the above illogical message.

**Environment**: 
All.

**Additional context**: 

The WTI-UI "Submit Run" button builds a `core.models.Submission` (in method `new-run.component.onSubmitProblem()`) and ultimately sends it to PC2 API method `ServerConnnection.submitJudgeRun(...)`.  That method in turn invokes `InternalController.submitJudgeRun(...)`, which already does a number of sanity checks on the requested submission and throws `SubmissionRejectedException` if the submission isn't valid.  It's likely that the best fix for this issue is to define a new type of `SubmissionRejectedException` (such as `SubmissionRejectionReason.ZeroLengthFile`) and throw that if any file of zero length is detected.

Also, the default submission rejection message should probably be changed; `Not accepting submissions at this time` is not a reasonable default message. A more generic default message such as `Error submitting program` would probably be better.

clevengr:
  Provides support for enabling or disabling acceptance of zero-length files in WTI submissions.  Specifically:
- Adds a new configuration variable `allowZeroLengthSubmissionFiles` to the `ContestInformation` class (the default value for this is `false`), along with accessors and corresponding appropriate updates to method `isSameAs(ContestInformation)`.
- Adds a new "Allow zero-length submission files" checkbox to the `ContestInformationPane`.  Checking or unchecking this box respectively allows or disallows WTI submissions to contain zero-length files.  Pressing `Update` saves the new value in the current `ContestInformation`.
- Updates `IContestLoader` and `ContestSnakeYAMLLoader` by adding support for a new YAML variable `allow-zero-length-submission-files`.  The default value for this variable (if it is undefined in a YAML file being loaded) is `false`; specifying `allow-zero-length-submission-files: true` in YAML will override the default and update the current `ContestInformation` to allow zero-length files in WTI submissions.
- Updates `SubmissionRejectedException` to contain a new `SubmissionRejectedReason` of `ZERO_LENGTH_FILE`.
- Updates PC2 API `ServerConnection` to throw `SubmissionRejectedException`, with a reason of `ZERO_LENGTH_FILE`, when the WTI client submits a zero-length file.
- Updates `WTI-UI.new-run.component` to handle the various `SubmissionRejectionReason`s better. 
- Updates the `catch` clause in `clics.API202306.SubmissionService` to handle all `SubmissionRejectedReason`s.

Also contains the following "Continuous Improvement (CI)" updates:
- Updates `SubmissionRejectedException` by eliminating constructors which are inappropriate (e.g. constructors with no arguments; every `SubmissionRejectedException` is now required to specify a _reason_).
- Updates `InternalController` to invoke a constructor which specifies a `SubmissionRejectionReason` (instead of a constructor which allowed no reason, which was eliminated).
- Updates `TeamsController` to properly recognize when the list of "additional files" is empty, and to handle `SubmissionRejectedExceptions` properly.
- Replaces the default error message `Not accepting submissions at this time` in `WTI-UI.new-run.component` (see the Issue description) with the content of the actual error message supplied by the PC2 API, or none then with a more generic message "Error in submission".

Fixes #1170

Windows 11, Eclipse Version: 2020-12 (4.18.0), Java 1.8.0_271, Chrome Version 140.0.7339.81

#### Main file test:

- Start a PC2 Server and load a contest (e.g. with --load sumithello).
- Start a PC2 Admin.
- On the `Configure Contest > Settings` tab, verify that the "Allow zero-length submission files" checkbox (in the `Team Settings` section) is UNCHECKED.
- Start the contest clock running (e.g. using the Admin "Times" tab).
- Go into the "Projects" folder and unzip the WebTeamInterface-1.2 zip file.
- Go into the unzipped WTI folder and start the WTI Server (e.g. using `./bin/pc2wti`).
- Open a browser and login to the WTI as a team.
- On the "Runs" page, select "Submit Problem".
- Choose any  problem and any language.
- Create a file of zero length and select that as the "Main File".
- Click "Submit"
- Verify that a red error message like "Zero-length file not allowed" is displayed.

#### Additional Files test:

- Select "Submit Problem" on the WTI Runs page.
- Choose any problem and language.
- Select any NON-ZERO-LENGTH file as the "Main File".
- Click "Additional Files" and select a ZERO-LENGTH file.
- Click "Submit"
- Verify that an error message like "Zero-length file not allowed" is displayed in red.

#### Test that Throttling still works as expected:

- Submit 7 (or more) runs within a single minute.  (It's possible to do this via the WTI GUI if you're quick.)
- Verify that on the 7th submission you get a red error message like "Submission threshold exceeded".

#### Test that rejection of files which are too large still works as expected:

- Select "Submit Problem" on the WTI Runs page.
- Choose a problem and a language.
- Select a text file which is larger than the source file size limit as the "Main File".
	(The default is 256KiB, plus some overhead to allow for encoding; any text file larger than 264K should suffice.)
- Click "Submit"
- Verify that a red error message like "Source file(s) are too large" is displayed.

#### Test Allow-Zero-Length-Files from GUI

- On the PC2 Admin "Settings" tab, CHECK (enable) the "Allow zero-length submission files" checkbox, then click "Update".
- Re-execute the "Main File test" and "Additional Files tests" above; verify that both tests now DO allow submission of a
zero-length file (i.e. that a green "Run successfully submitted" message is received in response to each of those tests).

#### Test Allow-Zero-Length-Files from YAML

- Completely shut down the WTI Browser, the WTI Server, and the PC2 Server.
- Delete the "profiles" and "logs" folders, and the "profiles.properties" file, under the PR's PC2 distribution (i.e. the files which were created by the above tests).
- Add an entry "allow-zero-length-submission-files: true" to the "contest.yaml" or "system.pc2.yaml" file for whatever contest you are using for testing. (For example, if you've been using `--load sumithello`, go into `samps/contests/sumithello/config` and edit the `contest.yaml` file, adding that entry.)
- Restart the PC2 Server (be sure it starts from scratch; i.e. it doesn't give a Warning about "using an existing profile" -- check the console for such a message).
- Restart the PC2 Admin.
- Verify that the "Allow zero-length submission files" checkbox on the Admin Settings pane is CHECKED (enabled).
- Start the contest running.
- Restart the PC2 WTI Server.
- Start a new browser session (or clear the browser's cache).
- Connect the browser to the PC2 WTI Server and login as a team.
- Re-execute the "Main File" and "Additional Files" tests (above); verify that zero-length files ARE accepted.

#### Test disabling of Allow-Zero-Length-Files after being enabled from YAML

- On the PC2 Admin, UNCHECK the "Allow zero-length submission files" checkbox, then click `Update`.
- In the WTI browser, re-execute the "Main File" and "Additional Files" tests (above); verify that zero-length files now REJECTED.

### DIFF
--- a/src/edu/csus/ecs/pc2/api/ServerConnection.java
+++ b/src/edu/csus/ecs/pc2/api/ServerConnection.java
@@ -27,6 +27,7 @@ import edu.csus.ecs.pc2.core.exception.SubmissionRejectedException;
 import edu.csus.ecs.pc2.core.model.Account;
 import edu.csus.ecs.pc2.core.model.ClientId;
 import edu.csus.ecs.pc2.core.model.ClientType;
+import edu.csus.ecs.pc2.core.model.ContestInformation;
 import edu.csus.ecs.pc2.core.model.ContestTime;
 import edu.csus.ecs.pc2.core.model.IFile;
 import edu.csus.ecs.pc2.core.model.IInternalContest;
@@ -431,37 +432,26 @@ public class ServerConnection {
         checkWhetherLoggedIn();
 
         checkIsAllowed (Permission.Type.SUBMIT_RUN, "User not allowed to submit run");
-
-        File mainFile = new File(mainFileName);
-        if (! mainFile.isFile()){
-            throw new Exception("File '" + mainFileName + "': no such file (not found)");
-        }
         
-        long sourceSize = mainFile.length();
-        
-        // make sure we don't accept a zero-length main file if that's not allowed
-        if (sourceSize <= 0) {
-            throw new SubmissionRejectedException("Zero-length file not allowed", SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);            
-        }
+        //throw an exception if the main file is not legal
+        checkLegalFile(mainFileName);
 
         // we are keeping track of the total submission size in sourceSize
+        File mainFile = new File(mainFileName);
+        long sourceSize = mainFile.length();
+        
+        //check each "additional file" (if any) for validity and if valid, add file size to running total size
         SerializedFile[] list = new SerializedFile[additionalFileNames.length];
         for (int i = 0; i < additionalFileNames.length; i++) {
-            File addFile = new File(additionalFileNames[i]);
-            //make sure the named item really is a file...
-            if (addFile.isFile()) {
-                // make sure we don't accept a zero-length additional file if that's not allowed
-                if (addFile.length() <= 0) {
-                    throw new SubmissionRejectedException("Zero-length file (in 'additional files') not allowed", 
-                            SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);
-                } else {
-                    //file is ok; add to list and update total source size
-                    list[i] = new SerializedFile(additionalFileNames[i]);
-                    sourceSize += addFile.length();
-                } 
-            } else {
-                throw new Exception("File '" + additionalFileNames[i] + "': no such file (not found)");
-            }
+            
+            //throw an exception if the current additional file is not legal
+            checkLegalFile(additionalFileNames[i]);
+            
+            //file is ok; add to list and update total source size
+            list[i] = new SerializedFile(additionalFileNames[i]);
+            File addlFile = new File(additionalFileNames[i]);
+            sourceSize += addlFile.length();
+
         }
 
         long sourceSizeByteLimit = internalContest.getContestInformation().getMaxSourceSizeInBytes();
@@ -472,6 +462,7 @@ public class ServerConnection {
 //            StaticLog.getLog().log(Level.WARNING, sizeMsg);
             throw new SubmissionRejectedException(sizeMsg, SubmissionRejectedException.SubmissionRejectionReason.SOURCE_TOO_BIG);
         }
+        
         ProblemImplementation problemImplementation = (ProblemImplementation) problem;
         Problem submittedProblem = internalContest.getProblem(problemImplementation.getElementId());
 
@@ -578,18 +569,9 @@ public class ServerConnection {
 
         checkIsAllowed (Permission.Type.SUBMIT_RUN, "User not allowed to submit run");
 
-        //reject zero-length file if such is not allowed
-        if (mainFile!=null && mainFile.getByteData().length<=0 && 
-                !internalContest.getContestInformation().isAllowZeroLengthSubmissionFiles()) {
-            throw new SubmissionRejectedException("Zero-length file not allowed", 
-                        SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);
-        }
+        //throw an exception if the main file isn't legal
+        checkValidIFile(mainFile);
         
-        //validate mainFile param
-        if (!validIFile(mainFile)) {
-            throw new Exception("Invalid mainFile parameter");
-        }
-
         // we keep track of the total submission size of all the source files
         long sourceSize = mainFile.getByteData().length;
         
@@ -597,24 +579,10 @@ public class ServerConnection {
         if (additionalFiles != null && additionalFiles.length > 0) {
             for (IFile nextFile : additionalFiles) {
                 
-                //reject zero-length additional file if such is not allowed
-                if (nextFile!=null && nextFile.getByteData().length<=0 && 
-                        !internalContest.getContestInformation().isAllowZeroLengthSubmissionFiles()) {
-                    throw new SubmissionRejectedException("Zero-length file (in 'additional files') not allowed", 
-                                SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);
-                }
-                
-                //check IFile validity
-                if (!validIFile(nextFile)) {
-                    // check whether the "invalid IFile" reason is a zero-length file if that's not allowed
-                    if (nextFile.getByteData().length <= 0) {
-                        throw new SubmissionRejectedException("Zero-length file (in 'additional files') not allowed", 
-                                                                SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);
-                    } else {
-                        throw new Exception("Invalid IFile in additionalFiles array");
-                    }
-                }
-                
+                //throw an exception if the current "additional file" isn't legal
+                checkValidIFile(nextFile);
+                        
+                //nextFile is valid; add its size to the running total size
                 sourceSize += nextFile.getByteData().length;
             }
         }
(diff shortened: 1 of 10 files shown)


src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -434,19 +434,33 @@ public void submitJudgeRun(IProblem problem, ILanguage language, String mainFile
...
         File mainFile = new File(mainFileName);
         if (! mainFile.isFile()){
-            throw new Exception("File '"+mainFileName+"' no such file (not found)");
+            throw new Exception("File '" + mainFileName + "': no such file (not found)");
+        }
+        
+        long sourceSize = mainFile.length();
+        
+        // make sure we don't accept a zero-length main file if that's not allowed
+        if (sourceSize <= 0) {
```
johnbrvc: By including this code, we are saying that we are never going to allow zero length files from the WTI; that is, there is no way to disable this.  Which is fine, I am just pointing it out.

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -564,20 +578,37 @@ public void submitJudgeRun(IProblem problem, ILanguage language, IFile mainFile,
 
         checkIsAllowed (Permission.Type.SUBMIT_RUN, "User not allowed to submit run");
 
+        //reject zero-length file if such is not allowed
```
johnbrvc: The statement that follows this new block of code:
`if (!validIFile(mainfile)){`
apparently already checks for all these conditions (including zero length files); why not modify that routine to throw the exception, or at least, return an exception string.  Basically, why check the same stuff twice?

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -564,20 +578,37 @@ public void submitJudgeRun(IProblem problem, ILanguage language, IFile mainFile,
...
         }
 
         // we keep track of the total submission size of all the source files
         long sourceSize = mainFile.getByteData().length;
-
+        
         // validate additionalSourceFiles param
         if (additionalFiles != null && additionalFiles.length > 0) {
             for (IFile nextFile : additionalFiles) {
+
```
johnbrvc, line 581: Same comment as above.   the ValidIFile() routine checks all these conditions already - why duplicate?

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -706,19 +736,42 @@ public void submitTestRun(IProblem problem, ILanguage language, IFile mainFile,
...
+                                SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);
+                }
+                
                 if (!validIFile(nextFile)) {
                     throw new Exception("Invalid IFile in additionalSourceFiles array");
                 }
             }
         }
 
+        //reject zero-length testDataFile if such is not allowed
```
johnbrvc: Same comment again.  Why duplicate this code over and over; just put it in `validIFile `? Or, at the very least, rework `validIFile()` to accommodate the desired action.

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -706,19 +736,42 @@ public void submitTestRun(IProblem problem, ILanguage language, IFile mainFile,
...
+        if (testDataFile!=null && testDataFile.getByteData().length<=0 &&
```
johnbrvc: Same. ` validIFile()`

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -727,6 +780,15 @@ public void submitTestRun(IProblem problem, ILanguage language, IFile mainFile,
         // validate additionalTestDataFiles param
         if (additionalTestDataFiles != null && additionalTestDataFiles.length > 0) {
             for (IFile nextFile : additionalTestDataFiles) {
+
+                //reject zero-length additionalTestDataFile if such is not allowed
```
johnbrvc: Same. validIFile()

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -1610,21 +1672,26 @@ public static void main(String[] args) {
...
+     * @return true if the IFile is not null and has a file name that is not null and not empty;
+     *         false otherwise.
+     *         
+     * @see SubmissionRejectedException
      */
     private boolean validIFile(IFile file) {
         if (file==null || file.getFileName()==null || file.getFileName().equals("")) {
             return false;
         }
-        if (file.getByteData().length<=0) {
```
johnbrvc, line 1625: Why was this removed?  It seems unnecessary to duplicate this code over and over.  Just "fix" validIFile() to return a String perhaps?  If it's non-null, then an error occurred?  I would rather have all the checking for validity be in the "validIFFile()" routine, after all, it's CALLED that for a reason.  Or, just throw the exception from validIFile() if the file is empty?

src/edu/csus/ecs/pc2/api/ServerConnection.java
```diff
@@ -706,19 +736,42 @@ public void submitTestRun(IProblem problem, ILanguage language, IFile mainFile,
 
         checkIsAllowed (Permission.Type.SUBMIT_RUN, "User not allowed to submit test run");
 
+        //reject zero-length file if such is not allowed
```
johnbrvc: Same comment as above regarding `validIFile()`

johnbrvc (changes requested):
The only major problem I have is with the `validIFile()` stuff.  I don't like the idea of duplicating the code that checks for zero length files all over the place.  It seems it should be in the `validIFile()` method, or, at the very least another method that explicitly checks the the length of the file and throws the exception if it's bad.  The name `validIFile() `seems fitting enough though, IMHO.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: I agree this should be fixed; I missed the presence of this test in the alternative version of submitJudgeRun(...)  (that is, the version that accepts string filenames instead of IFiles).  I'll fix it.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Originally, `validIFile()` DID check for zero-length files; however, the CURRENT version of that method does not (this is documented in the method, and in the commit message for that change -- 7fdbcc8ccce6fc56e52f1356f9fddd1c7226dfe0).  The reason that `validIFile()`  was changed so that it no longer checks file length is that there's no way within the method to determine whether an "invalid file" (due to being zero length when that condition is disallowed by the new `ContestInformation` setting) is an invalid "main file" or an invalid "additional file".  (There are different exception messages returned for those two cases -- but `validIFile()` has no way to determine which condition applies.)  That's why I originally did it with separate external checks rather than inside `validIFile()` -- because it required different exception messages for the two different conditions.

I think the simplest solution is (as you suggested) to put the "file length checking" back into `validIFile()`, and to return the same exception message regardless of whether the illegal zero-length file is the main file or an "additional file".  It takes a bit of information away from the user, but it makes the code much cleaner.   (Although I will also first try to find a solution that preserves both goals...)

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: I assume that this comment refers to the block of code FOLLOWING the comment (i.e., lines 600-605), not the block of code (lines 596-598) just above the comment?

Assuming so, then it appears to me that your assertion is wrong; `validIFile()` (in it's current version) does NOT check for "all those conditions" -- specifically, it does not currently check the file length.

(If on the off chance that you actually _**were**_ referring to lines 596-598, then I don't understand your comment; those lines are checking the length of the _array of files_, not the length of any one file in that array.)

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Same answer (see above for why it was done that way initially).  I'll rework it to address the concern of duplicated code spread around different places.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Same answer as above.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Same answer.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: The answer to "why was this removed?" is (I think) explained above.  Basically, `validIFile()` has no way to determine which kind of file (main vs. additional) is illegally zero-length, so it can't return different exception messages for those two conditions.  Hopefully I'll find an acceptable fix for this.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Same answer as above.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in next push:  replaced the incorrect line with a call to new method `checkFilePassesZeroLengthRestrictions(File) throws SubmissionRejectedException`, which looks like:
```java
    /**
     * Checks whether the specified File is zero length; throws {@link SubmissionRejectedException} if so AND if
     * zero-length files have been configured as being disallowed.
     * 
     * @param aFile the File whose length is to be checked.
     * @throws SubmissionRejectedException if the file is zero length AND zero-length files are not allowed in submissions.
     */
    private void checkFilePassesZeroLengthRestrictions(File aFile) throws SubmissionRejectedException {
        
        long fileSize = aFile.length();

        if (fileSize <= 0 && !internalContest.getContestInformation().isAllowZeroLengthSubmissionFiles()) {
            throw new SubmissionRejectedException("Zero-length file not allowed", SubmissionRejectedException.SubmissionRejectionReason.ZERO_LENGTH_FILE);            
        }   
    }

```

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Assuming I understood the comment correctly, fixed in latest push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push.

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push (basically, the test implemented in the removed code is now implemented in a method which is called at that point.)

src/edu/csus/ecs/pc2/api/ServerConnection.java
clevengr: Fixed in latest push.

johnbrvc (commented):
Did a code review and looks OK.  Now I have to test it.  That's the hard part.

clevengr:
> Did a code review and looks OK. Now I have to test it. That's the hard part.

Huh; I would have thought the code review would be the hard part (since I refactored a number of things).  I thought I provided a pretty complete set of tests; I was able to execute all the tests in about 10 minutes...

johnbrvc (approved):
Ran through all the tests.  Wow.  That was A LOT of stuff to test.  All tests ran as expected.

SamanwaySadhu (approved):
Ran every test except the Large File Size. Runs as described.

Issue #689, opened 2023-02-15
Improve/automate/redesign how divisions are handled


lane55:
**Feature Description**: please provide a clear and concise description of the proposed feature:  

Currently divisions are identified by a D# hard coded at the end of a group name.

Change the system to not require manual changing of group names, problem.yaml, xsl etc with
a particular emphasis on automating (reducing manual work).

...

**Have you considered other ways to accomplish the same thing?** Please provide a clear and concise description of any alternative solutions or features you've considered.  
...

**Do you have any specific suggestions for how your feature would be ***implemented*** in PC^2?** If so, please describe.  
...

**Additional context**: please provide any additional information that will help us in considering your feature request.  
...

clevengr:
Among the (many) things that need to be considered in the process of implementing a solution for this issue are:

- Figuring out how to deal with separate PDF writeups for contests with different _divisions_ (such as the PacNW contest).  See Issue #1184  for relevant information.
-

PR #1186, opened 2025-11-12
I1184 update wti scoreboard problem letter pdf links

### LINKED ISSUE #1184: Update WTI Scoreboard problem letter PDF links to support CDP shortname-based URLs
The current implementation of WTI Scoreboard Problem PDF links (that is, clicking on a problem letter opens the PDF for the corresponding problem writeup) are essentially "hard-coded" to reference PDFs with problem letter-based names -- A.pdf, B.pdf, etc.  This is inflexible and cumbersome; it requires the Contest Administrator to create an extra set of "letter-named" pdfs for the problem.  Also, the letter values in the HREF text are "hard-coded" to be A, B, C, etc., meaning you can't have a problem set that uses different letter sequences.

**Feature Description**: 
Update the HREFs associated with the problem headers so that the hyperlink references a file based on the "CLICS shortname" associated with the problem.

- Update the Default Scoring Algorithm so that it inserts a shortname-based URL in the problem definitions which get put into the XML output (which is then converted to JSON and transmitted to the WTI server).
- Update the Admin Guide section which explains how/where to post PDFs that are linked from the WTI scoreboard.

clevengr:
Updates the URL attached to Problem Letters in the WTI Scoreboard to follow the convention that for a given problem with letter <letter> the PDF for the problem is expected to be in the WTI "WebContent/WTI-UI/problems" folder in a file named like "<letter>-<problemShortName>.pdf in a folder (beneath "/problems") named like "<letter>-<problemShortname>".

Fixes #1184 

Windows 11, Java 1.8, Eclipse 2020-12

- Start the PC2 Server and a PC2 Admin.
- Use the PC2 Admin to define four problems, "Hello1", "Hello2", "Sumit1", and "Sumit2" with letters A, B, C, and D respectively.
  - Alternatively you can define such a contest, with such problems, in YAML files before starting the PC2 Server.
- Create four _different_ PDFs, one for each problem and following the convention outlined above -- that is, four PDFs named respectively "A-Hello1.pdf", "B-Hello2.pdf", "C-Sumit1.pdf", and "D-Sumit2.pdf".  (The PDFs don't actually have to represent a "real problem"; they just need to be named as specified and need to have content which is recognizably different.)
- In the WTI Server folder, create four folders underneath `WebContent/WTI-UI/problems` following the convention outlined above -- that is, create a "problems" folder and then under that folder create subfolders "A-Hello1", "B-Hello2", "C-Sumit1", and "D-Sumit2".
- Copy each PDF file into its corresponding folder underneath `WebContent/WTI-UI/problems`.
- Start the WTI Server.
- Start a browser and connect/login to the contest.
- Select the "Scoreboard" tab on the browser.
- Click each of the problem letter headers at the top of the table.  

  Each header link should open a new browser tab which displays the correct PDF when a problem letter is clicked in the WTI-UI Scoreboard.  

Note:  this has NOT yet been tested with teams assigned to different DIVISIONS and with the problems assigned to unique divisions.  This has been added as a separate element of Issue #689.


src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -1064,6 +1094,25 @@ private void generateSummaryTotalsForProblem(Problem[] problems, Hashtable<Eleme
...
+     * @param input the string to be processed.
+     * @return an equivalent string with the first character capitalized if it is a letter.
+     */
+    private String capitalizeFirstLetter(String input) {
+        if (input == null || input.isEmpty()) {
+            return input; // Return as-is if null or empty
+        }
+        
+        char firstChar = input.charAt(0);
+        if (!Character.isLetter(firstChar)) {
```
johnbrvc, line 1109: I do not believe this test is necessary as toUpperCase() will return non-letters unchanged..

johnbrvc (changes requested):
Other than the one minor comment I made, I see no issues with the code  although I am no Angular/typescript expert so the Angular stuff I just looked through.  

I tested it with SumitHello, and it does not seem to work.  That is, the scoreboard letters at the top of each column do not open a PDF (even though I made sure they exist).  I put a `System.err.println()` in DSA where it computes the URL, and that appears to be right:

```
Version 9 () Java ver 1.8.0_321 build  Windows 11 10.0 (amd64)
Login: team2 (aka team2)
 receiveObject(S) end   got Packet[LOGIN_SUCCESS] #13 SERVER0 @ site 1 -> TEAM2 @ site 1
ContestController: debug 22 Start
ContestController: Runs for 4e791260-5069-4031-af82-b468ecee410d Not using division useDivisionFilter false, total runs  = 3
***** url is: problems/A-Hello/A-Hello.pdf
***** url is: problems/B-Sumit/B-Sumit.pdf
```
Also, there are no instructions indicating that you have to create: `WebContent/WTI-UI/problems/A-Hello/A-Hello.pdf or WebContent/WTI-UI/problems/B-Sumit/B-Sumit.pdf`

I looked at the headers in the browser debugger and all I see for the letter headers are:

`<th _ngcontent-ng-c1632187819="" ng-reflect-ng-style="[object Object]" class="ng-star-inserted" style="text-align: center; padding: 2px; background-color: rgb(255, 255, 0); color: black; cursor: pointer;"> B </th>`

I only see the "B" there, no link.  (Should there be a link?)  

Also, I think the `[A-Z].pdf `files in the problems folder should be removed from the project since they are no longer relevant.  Specifically, instructions have to be provided in the PR on where to create the PDF's and what to call them.

I also do not understand the description of the PR.  it says:

>  is expected to be in the WTI "WebContent/WTI-UI/problems" folder in a file named like "-.pdf in a folder (beneath "/problems") named like "-".

While it does suggest the files have to be called "something", what is shown in the PR is not right.  Also, there should be a step in the test procedure saying to create said files.
What does `"-.pdf" `mean?  according to the `System.err.println`, it should be `problems/A-Hello/A-Hello.pdf`

projects/WTI-UI/src/app/modules/scoreboard/components/scoreboard-page/scoreboard-page.component.html
```diff
@@ -13,12 +13,14 @@
                 		'color': item.textColor,
                 		'cursor': 'pointer' 
                 	}">
+				<!--
```
johnbrvc: Could this be the reason that the URL's are not working for problems?  The HTML code is commented out?  Again - I don't know how Angular does things, maybe it adds it somehow later, but clearly, this reference to the URL IS commented out for some reason.

johnbrvc: In the interest of speeding up approval of this PR, I tried uncommenting the HREF generation HTML in scoreboard-[page.component.html ](https://github.com/pc2ccs/pc2v9/pull/1186/files#diff-36e690e8f87d685803820cc08ef390b3be73354724518083499a425a9e789be7) so it looks like this:

```
                	<a [href]="isContestStarted() ? item.url : null" 
                			(click)="onLinkClick($event, item.url)" 
                			target="_blank" 
                			class="problem-writeup-link" >
    					{{ item.label }}
  			        </a>

```

I then rebuilt and tried it, and that seems to fix the problem - meaning, the problem letter links are clickable and they contain the correct link.   Can @clevengr confirm that this IS the issue with this PR, and that the fix I tried above is the correct fix?  Perhaps equally as important, how come it was commented out in the first place?

clevengr: Your analysis is correct:  the symptom you were seeing (problem letters don't contain hyperlinks) is indeed because the hyperlink code was commented out.  Uncommenting that block, as you discovered, makes everything work as expected.

Regarding "how come it was commented out in the first place": this is answered in the commit `1458ea0d7c32503c7c7657771f9c413a297c1c44` message of 2025-11-14 22:47:32, which says (in the extended message):

```
i1184: remove hyperlink from problem letters; necessary because...

we haven't yet figured out how to make the hyperlinks point to different
problem writeups when there are two Divisions.
```
This commit was entered from my hotel room at the Wild Palms in San Jose the night before the PacNW Regional Contest because, well, we hadn't figured out how to make the hyperlinks work with divisions.  I also pointed this out in the "Steps for testing the PR", where I said:

```
I have NOT yet tested it with teams assigned to different DIVISIONS and with the problems assigned to unique divisions.
```
 So basically the "fix" is to remove that commit.  And then we (eventually) need to figure out how this code, probably along with a whole bunch of other code, is going to deal with 'divisions'.

You are also correct that I failed to include, in the "Steps for testing the PR", the need to create PDFs at the appropriate location prior to running the test. I've added that to the Testing Steps.

clevengr:
I have reverted the "commenting-out" of the PDF links, updated the "Steps to test the PR", and run those tests myself; everything seems to work as originally intended.  Requesting re-review and approval.

johnbrvc (approved):
Works as expected.

.gitignore
```diff
@@ -22,5 +22,4 @@ cacerts.pc2
 
 projects/*.gz
 projects/*.zip
-projects/WTI-UI/package-lock.json
```
SamanwaySadhu: I don't think this is required. The particular files in .gitignore can also be bypassed if need be.

kkarakas:
I am not sure how to test is so I couldn't. How do I define 4 problems? I could go to config file and change problemset.yaml would that be sufficient? Would I have to create files and do I need to create a blank contest to do all of this?

clevengr:
> I am not sure how to test is so I couldn't. How do I define 4 problems? I could go to config file and change problemset.yaml would that be sufficient? Would I have to create files and do I need to create a blank contest to do all of this?

@kkarakas You can define problems using the PC2 Admin "Problems>Add" function.  But yes, you could also just change the problemset.yaml to include more problems (for example, you could edit the local copy under samps/src/sumithello and add two problems to it).  See the second bullet under "Precise Steps for Testing the PR", above.

JoeTerlizzi (approved):
I followed the testing procedures as described in the PR.

While doing so I discovered a possible bug and have reported it
https://github.com/pc2ccs/pc2v9/issues/1216

This bug does not seem to be from the code changes in this PR.  When adding NEW problems manually everything worked as expected.

Approved.

kkarakas (approved):
Followed the testing procedure. I made a smal change though: I added '-' to the name of the subfolder and pdf. So for example instead of "A-Hello1" I named the pdf and the subfolder "A-Hello-1". After that small change it works. One nitpicking concern is the fact thatevery contestant would have access to every pdf. I imagine it is not a concern.

johnbrvc:
> Followed the testing procedure. I made a smal change though: I added '-' to the name of the subfolder and pdf. So for example instead of "A-Hello1" I named the pdf and the subfolder "A-Hello-1". After that small change it works. One nitpicking concern is the fact thatevery contestant would have access to every pdf. I imagine it is not a concern.

The contestants always had access to the PDF's; there's a link to them on the contest image.  Why would you think it might be a concern?

kkarakas:
> > Followed the testing procedure. I made a smal change though: I added '-' to the name of the subfolder and pdf. So for example instead of "A-Hello1" I named the pdf and the subfolder "A-Hello-1". After that small change it works. One nitpicking concern is the fact thatevery contestant would have access to every pdf. I imagine it is not a concern.
> 
> The contestants always had access to the PDF's; there's a link to them on the contest image. Why would you think it might be a concern?

Well then you have a document accessible when you write the exact url or from the image but not have a button for. Isn't that weird from a design perspective?

PR #1225, opened 2026-02-17
I1223 handle empty score.txt

johnbrvc:
### Description of what the PR does
Fixes the handling of an empty or bad score.txt file for point scoring contests.

### Issue which the PR addresses
Fixed #1223 

### Environment in which the PR was developed (OS,IDE, Java version, etc.)
Windows 11
Ubuntu 24.04.3

### Precise steps for _testing_ the PR (i.e., how to demonstrate that it works correctly)
See the Issue.  The setup is quite elaborate to test this.  I have tested it using the NAC2026 challenge2025_test contest and it works as described in the issue.  If necessary, we can discuss a test bed for this.

Some notes to help testing on Windows...

If you are testing with the NAC2026/contests/challenge2024_test contest, the problem that causes this issue is **treasurepacking** (the only one with a custom validator).
It turns out, the validator is a _python_ program.  PC2 does not know how to simply "execute" a python validator; on Linux it's easy because of the _shebang_ (`#!`) at the beginning of the python script.  On Windows, you have to tell PC2 to execute: `python validate.py ... ` , however, there's more to it than just that.

You'll have to edit the problem, turn off the "sandbox" option (**Sandbox** tab, select **Do not use a Sandbox**)
Then on the **Output Validator** tab, at the bottom, for the "**Validator Command Line**", **_prepend_** what is there already with: `..\..\..\..\..\..\..\..\<full path of where python is>\python.exe`  so on my system the full **Validator Command Line**, looks like this:
`..\..\..\..\..\..\..\..\Users\John Buck\AppData\Local\Programs\Python\Python312\python.exe {:validator} {:infile} {:ansfile} {:feedbackdir}`

You will also have to modify the "**Judge's Data Path**" (on the **Problems** tab) to be wherever you extracted the contest repo to, so if you extracted it to "`\MyRepos`", you would **Set Judge's Data Path** to:
`\MyRepos\NAC2026\contests\challenge2024_test\config`

Next, on the  **Auto Judge** tab, pick whatever autojudge you're going to use (presumably **judge1**), and edit it.  Uncheck all but "**Treasure Packing**".
image
If you don't do this... well... you'll see what happens...  It's not a catastrophe, but it's annoying.

You can then bring up a **pc2team** client, log in as **team1** with password "**frill-equity**" and submit ANY "cpp" file for problem **treasurepacking**.  The idea is to get a **WA**.  I used: **samps/src/hello.cpp**

PR #1226, opened 2026-02-17
I1224 point score decimal format

### LINKED ISSUE #1224: For point scoring contests, make the score value format configurable
This is more for uniformity in reading contest scores.   Some scores may have a large number of digits after the decimal point.  This is not desirable in scoreboard outputs, for example.  While we want to maintain the scoring accuracy internally, we don't necessarily need it displayed.  
...

**Feature Description**: 
Make the score format configurable in the `pc2v9.ini` file so that different applications may choose to use different formats.  For example, the **pc2board** utility may want a format of `0.0##`, while the **pc2ef** may want `0.0###`, while the `pc2admin `may want `0.0########`.  

...

There is no workaround for this.
...

Add a section called `[scoring] `to the pc2v9.ini file with a key of `format`.   This can be assigned any format supported by the `java.text.DecimalFormat` class.
...

**Additional context**:  
...

johnbrvc:
Allows configuring of the display format (`DecimalFormat`) used to format scores from a point scoring contest.

Fixes #1224 

Windows 11
Ubuntu 24.04.3

Load up any point scoring contest that is complete, eg. NAC2026 **challenge2024_test**, for example.  Details of obtaining the completed contest results can be directed to @johnbrvc .   Suffice to say, I have tested it with three different apps (**pc2admin**, **pc2board**, **pc2feeder**) all using different `pc2v9.ini` files with different formats for the point scores, and it appears to work as described in the issue.
#### The following additional details may also be helpful in testing the PR
- The "completed scoring contest" mentioned above, which can be used for testing, can be found at [this repo location](https://drive.google.com/drive/folders/11i55pRYbh4rJB4Vj2pCAY2TkK9eINysB?usp=drive_link). To use it:
  - Download the NAC2026-Challenge2024-test.zip contest from that link and unzip it.
  - Copy the `profiles` folder and the `profiles.properties` file into the PC2 folder where your PC2 distribution (what you are going to test) resides.  Be sure you have NOT started the PC2 Server before copying the profiles into PC2.
  - Start the PC2 Server.  This will automatically load the NAC2026-Challenge2024-test contest into PC2.
- Suggested testing steps:
1. **Verify that every Point Score display contains the correct default format, which is `0.000` (at least one whole digit and exactly 3 decimal digits), as follows:**
  - Start a PC2 Admin and look at the `Run Contest>Scoreboard` tab; ensure that the scores all have exactly three decimal places (this is the default).
  - Test that the PC2 scoreboard displays default-formatted scores properly:
    - On the PC2 Admin, check the `Configure Contest>Settings` tab and make sure the `Output HTML dir for Judges` and `Output Public HTML dir` paths point to a valid location on your machine (typically you'd want them pointing to locations under the PC2 distribution you are testing).
    - Start a PC2 Board client and verify that the point scores shown on the Board Client GUI all have exactly three decimal places.
    - In a browser, open the file `index_ps.html` in the "Judge's HTML" ("secret") folder.
    - Verify that the values in the "Score" column (next to each team name) all have exactly three decimal places.
  - Test that the PC2 Event Feed contains properly-formatted default scores:
    - Start feeder 1 client (**pc2ef --login ef1**), then press the Start button to start the feeder.
    - From a browser, open **https://localhost:50443/contests/nac25chall24practice/scoreboard**. When it prompts for user/password, use: **administrator1:administrator1**, or **admin:admin (from the realm.properties file, assuming you have one)**.
    - Execute the following command from a command line: 
        ```
           curl -k https://admin:admin@localhost:50443/contests/nac25chall24practice/scoreboard 
        ```
        (you can also use **administrator1:administrator1**. If you're using a pc2 account it must have the View Event Feeds permission. )
    - Verify that the "score" values in the resulting Event Feed output which are not integer numbers (e.g. 50.0) have exactly three decimal places.  (The CLICS API specifies that scores are _numbers_, which means that whole numbers will display in the event feed without decimal places.)
  - Test that the WTI displays default-formatted scores properly:
    - In the PC2 `projects` folder, unzip the WebTeamInterface.zip file.
    - In the resulting WebTeamInterface folder, start the WTI Server (`./bin/pc2wti`).
    - Open a browser and login to the WTI as a team (e.g. `team1` with password `decoy-sled-dwarf-idly`).
    - On the WTI `Runs` page, verify that every "Accepted" run has a score containing exactly 3 decimal digits.
    - On the WTI `Scoreboard` page, verify that:
      - Every green box (indicating an Accepted/Solved problem) has a score that contains exactly 3 decimal digits.
      - Every entry in the rightmost `Score` column has exactly 3 decimal digits.
2. **Verify that a bad `format` causes a client to fail to start, as follows:**
  - Shut down the PC2 Admin.
  - Edit the `pc2v9.ini` file in the root PC2 distribution folder by adding the following lines at the end:
    ```
    [scoring]
    format=0.00,00
    ```
    (The comma makes this an illegal format pattern.)
  - Restart the PC2 Admin, and verify that it fails to start and gives an error message that the format is illegal.
  
3. **Verify that changing to a non-default score format results in every Point Score display containing the correct (specified) format, as follows:**
  - Shut down the WTI browser, the WTI Server, and the PC2 Admin, Board, and Event Feed clients.
  - Edit the `pc2v9.ini` file in the root PC2 distribution folder by adding the following lines at the end:
    ```
    [scoring]
    format=0.00000
    ```
    (This forces the formatting of scores so that they have 5 decimal digits, regardless of the actual numeric value being displayed).
  - Restart the PC2 Admin, Board, and Event Feed clients (as described above), and verify that all the places where scores are displayed (as described above) now show FIVE decimal digits.
  - Edit the `pc2v9.ini` file under the `projects/WebTeamInterface` folder in a similar manner, except _use format `0.0000`_ (only **four** decimal digits).
  - Restart the WTI Server, then re-login to a team.
  - Verify that the `Runs` and `Scoreboard` pages now show scores containing exactly _four_ decimal digits.

4. **If you want to really go crazy:**
  - Start each of the PC2 Admin, Board, and EF clients in their own separate folder, each with its own `pc2v9.ini` file containing a different `format=x.xxx` value, and verify that each client displays its scores according to its own format specification.  Most people would consider this "overkill-testing"...

### DIFF
--- a/projects/WTI-API/src/main/controllers/ContestController.java
+++ b/projects/WTI-API/src/main/controllers/ContestController.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2024 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2026 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package controllers;
 
 import java.io.IOException;
@@ -12,6 +12,7 @@ import java.util.Properties;
 import java.util.Set;
 
 import javax.inject.Singleton;
+import javax.json.JsonObject;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.Path;
@@ -36,6 +37,7 @@ import edu.csus.ecs.pc2.api.exceptions.LoginFailureException;
 import edu.csus.ecs.pc2.api.exceptions.NotLoggedInException;
 import edu.csus.ecs.pc2.core.IniFile;
 import edu.csus.ecs.pc2.core.StringUtilities;
+import edu.csus.ecs.pc2.core.Utilities;
 import edu.csus.ecs.pc2.core.exception.IllegalContestState;
 import edu.csus.ecs.pc2.core.log.Log;
 import edu.csus.ecs.pc2.core.model.ClientId;
@@ -930,11 +932,116 @@ public class ContestController extends MainController {
 		
 		JSONObject jsonStandingsObject = XML.toJSONObject(xmlStandings);
 	
+		changeScoresToStrings(jsonStandingsObject);
+		
 		logger.fine("Standings JSON: " + jsonStandingsObject.toString(2));
 
 		return jsonStandingsObject.toString();
 	}
 
+    /**
+     * Scan the supplied JSONObject for problemSummaryInfo's that have numeric "score" properties and
+     * reformat them and put them back as Strings.  By default, the XML.toJSONObject() will attempt to determine
+     * what type a value is and create JSON that uses that type.  So, for example, an XML attribute such as:
+     * score="12.340000"
+     * would be converted to this JSON:
+     * "score": 12.34
+     *  
+     * The structure of the original XML we are interested in is:
+     * <contestStandings ...>
+     *  <standingsHeader ...>
+     *   <problem ...>
+     *  </standingsHeader>
+     *  <teamStanding ... >
+     *   <problemSummaryInfo ...>
+     *  </teamStanding>
+     *  <teamStanding ... >
+     *   <problemSummaryInfo ...>
+     *  </teamStanding>
+     *  ...
+     * </contestStandings>
+     *
+     * We will check the "probemSummaryInfo" json objects for "score" properties, fetch them, reformat them, and
+     * put them back in the json as Strings.
+     * 
+     * It should be noted that XML.toJSONObject() does have a mechansism to convert ALL values to Strings, but we
+     * do not want that.  We wish to keep other numeric properties as numbers and booleans and whatnot.
+     *
+     * @param xmlJSONObj The main JSONObject obtained from XML.toJSONObject()
+     */
+    private void changeScoresToStrings(JSONObject xmlJSONObj) {
+        JSONObject currentObj = xmlJSONObj;
+
+        String CONTEST_STANDINGS = "contestStandings";
+        String TEAM_STANDING = "teamStanding";
+
+        // Find top level
+        currentObj = currentObj.optJSONObject(CONTEST_STANDINGS);
+        if (currentObj == null) {
+            logger.warning(CONTEST_STANDINGS + " object not found in standings XML");
+            return;
+        }
+        Object teamStandings = currentObj.opt(TEAM_STANDING);
+        
+        // if there's more than one teamStanding, it's in a JSONArray
+        if(teamStandings instanceof JSONArray) {
+        	JSONArray jArray = (JSONArray)teamStandings;
+        	int nStandings = jArray.length();
+        	for(int idx = 0; idx < nStandings; idx++) {
+                fixProblemSummaryInfo(jArray.optJSONObject(idx));
+            }
+        } else if(teamStandings instanceof JSONObject){
+            // Only one teamStandings
+            fixProblemSummaryInfo((JSONObject) teamStandings);
+        }
+    }
+
+    /**
+     * Iterates through a (possible) array of a teamStanding's problemSummaryInfo's and change the "score" property
+     * to a string.
+     *
+     * @param teamStanding - use this teamStanding's problemSummaryInfo's
+     */
+    void fixProblemSummaryInfo(JSONObject teamStanding) {
+        String PROBLEM_SUMMARY_INFO = "problemSummaryInfo";
+
+        // The teamStanding record has the total score in it, so it has to be fixed as well
+        fixScore(teamStanding);
+        
+        // Find childElement that contains the itemElement
+        Object psis = teamStanding.opt(PROBLEM_SUMMARY_INFO);
+
+        if(psis instanceof JSONArray) {
+        	JSONArray jArray = (JSONArray)psis;
+        	int nStandings = jArray.length();
+        	for(int idx = 0; idx < nStandings; idx++) {
+                fixScore(jArray.optJSONObject(idx));
+            }
+        } else if(psis instanceof JSONObject) {
+            fixScore((JSONObject)psis);
+        }
+    }
+
+    /**
+     * Check if the supplied ProblemSummaryInfo contains a score key that is a double, if so,
+     * convert it to string.
+     * 
+     * @param psi - the ProblemSummaryInfo of interest
+     */
+    void fixScore(JSONObject psi) {
+        String SCORE_PROPERTY = "score";
+        
+        try {
+            double score = psi.getDouble(SCORE_PROPERTY);
+            String formattedScore = Utilities.formatScore(score);
+            // Overwrite existing property with string version for formatted score.
+            psi.put(SCORE_PROPERTY, formattedScore);
+        } catch(Exception e) {
+            // If any exception occurs, that is, the property isn't there, or it's not a double,
+            // just leave it as is - this would be the normal case for non-scoring contests.
+        } 
+    }
+    
 	/**
 	 * Returns the flag indicating whether the cached copy of the contest standings are current (true),
 	 * or instead that some event has occured which potentially makes the standings out of date (false).
--- a/src/edu/csus/ecs/pc2/core/Utilities.java
+++ b/src/edu/csus/ecs/pc2/core/Utilities.java
@@ -112,7 +112,7 @@ public final class Utilities {
      * and the key: format in the [scoring] section
      */
     private static final String POINT_SCORING_FORMAT_INI_KEY = "scoring.format";
-    private static final String DEFAULT_POINT_SCORING_FORMAT = "0.0##";
+    private static final String DEFAULT_POINT_SCORING_FORMAT = "0.000";
     private static DecimalFormat dfPointScore = null;
 
 
--- a/src/edu/csus/ecs/pc2/core/Utilities.java
+++ b/src/edu/csus/ecs/pc2/core/Utilities.java
@@ -1,4 +1,4 @@
-// Copyright (C) 1989-2025 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
+// Copyright (C) 1989-2026 PC2 Development Team: John Clevenger, Douglas Lane, Samir Ashoo, and Troy Boudreau.
 package edu.csus.ecs.pc2.core;
 
 import java.io.BufferedReader;
@@ -14,6 +14,7 @@ import java.io.RandomAccessFile;
 import java.security.InvalidParameterException;
 import java.text.CharacterIterator;
 import java.text.DateFormat;
+import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
@@ -105,6 +106,15 @@ public final class Utilities {
      * SimpleDateFormat for iso8601 including milliseconds
      */
     private static SimpleDateFormat iso8601formatterWithMS = new SimpleDateFormat(ISO_8601_TIMEDATE_FORMAT_WITH_MS);
+    
+    /**
+     * Allows changing of the point scoring score format from the default using the pc2v9.ini file
+     * and the key: format in the [scoring] section
+     */
+    private static final String POINT_SCORING_FORMAT_INI_KEY = "scoring.format";
+    private static final String DEFAULT_POINT_SCORING_FORMAT = "0.0##";
+    private static DecimalFormat dfPointScore = null;
+
 
     /**
      * List of extensions for files considered to be executable.
@@ -2126,5 +2136,26 @@ public final class Utilities {
         return newArray;
     }
 
-
+    /**
+     * Format a score according to the format specified in the INI file.
+     * If no format found in the ini file, use the default.
+     * Note: You may ask "Why this isn't a "Setting" in ContestInformation (or something similar)?"
+     *     The answer is that some applications may wish to format the point scoring score differently.  This
+     *     allows those apps to set a specific format in the pc2v9.ini file.  (pc2board, pc2feeder, pc2admin to name a few).
+     * 
+     * @param score the score to format
+     * @return formatted score string
+     */
+    public static String formatScore(double score) {
+        
+        // first time through, create the formatter
+        if(dfPointScore == null) {
+            String scoreFormatString = IniFile.getValue(POINT_SCORING_FORMAT_INI_KEY);
+            if(StringUtilities.isEmpty(scoreFormatString)) {
+                scoreFormatString = DEFAULT_POINT_SCORING_FORMAT;
+            }
+            dfPointScore = new DecimalFormat(scoreFormatString);
+        }
+        return(dfPointScore.format(score));
+    }
 }
--- a/src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
+++ b/src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
@@ -139,7 +139,6 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
 
     private boolean obeyFreeze = false;
 
-
     /**
      * @return the obeyFreeze
      */
@@ -877,8 +876,7 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
             standingsRecordMemento.putLong("lastSolved", standingsRecord.getLastSolved());
 
             if(contestInformation.isScoreboardTypeScore()) {
-                DecimalFormat df = new DecimalFormat("0.0###");
-                standingsRecordMemento.putString("score", df.format(standingsRecord.getScore()));
+                standingsRecordMemento.putString("score", Utilities.formatScore(standingsRecord.getScore()));
             } else {
                 standingsRecordMemento.putLong("points", standingsRecord.getPenaltyPoints());
             }
@@ -974,7 +972,7 @@ public class DefaultScoringAlgorithm implements IScoringAlgorithm {
                     psiMemento.putString("shortName", problems[problemsIndexHash.get(psi.getProblemId())-1].getShortName());
                     psiMemento.putInteger("attempts", psi.getNumberSubmitted());
                     psiMemento.putInteger("points", psi.getPenaltyPoints());
-                    psiMemento.putDouble("score", psi.getScore());
+                    psiMemento.putString("score", Utilities.formatScore(psi.getScore()));
                     psiMemento.putLong("solutionTime", psi.getSolutionTime());
                     psiMemento.putBoolean("isSolved", psi.isSolved());
                     psiMemento.putBoolean("isPending", psi.isUnJudgedRuns());
--- a/src/edu/csus/ecs/pc2/core/scoring/NewScoringAlgorithm.java
+++ b/src/edu/csus/ecs/pc2/core/scoring/NewScoringAlgorithm.java
@@ -67,7 +67,7 @@ public class NewScoringAlgorithm extends Plugin implements INewScoringAlgorithm 
     private PermissionList permissionList = new PermissionList();
 
     private boolean isPointScoring = false;
-
+    
     /**
      * Return a list of regional winners.
      *
@@ -576,8 +576,7 @@ public class NewScoringAlgorithm extends Plugin implements INewScoringAlgorithm 
         standingsRecordMemento.putLong("lastSolved", standingsRecord.getLastSolved());
 
         if(contestInformation.isScoreboardTypeScore()) {
-            DecimalFormat df = new DecimalFormat("0.0###");
-            standingsRecordMemento.putString("score", df.format(standingsRecord.getScore()));
+            standingsRecordMemento.putString("score", Utilities.formatScore(standingsRecord.getScore()));
         } else {
             standingsRecordMemento.putLong("points", standingsRecord.getPenaltyPoints());
         }
@@ -887,7 +886,7 @@ public class NewScoringAlgorithm extends Plugin implements INewScoringAlgorithm 
         summaryInfoMemento.putString("shortName", summaryInfo.getShortName());
         summaryInfoMemento.putInteger("attempts", summaryInfo.getNumberSubmitted());
         summaryInfoMemento.putInteger("points", summaryInfo.getPenaltyPoints());
-        summaryInfoMemento.putDouble("score", summaryInfo.getScore());
+        summaryInfoMemento.putString("score", Utilities.formatScore(summaryInfo.getScore()));
         summaryInfoMemento.putLong("solutionTime", summaryInfo.getSolutionTime());
         summaryInfoMemento.putBoolean("isSolved", summaryInfo.isSolved());
         summaryInfoMemento.putBoolean("isPending", summaryInfo.isUnJudgedRuns());
--- a/projects/WTI-API/src/main/controllers/TeamsController.java
+++ b/projects/WTI-API/src/main/controllers/TeamsController.java
@@ -18,6 +18,7 @@ import javax.ws.rs.GET;
 
 import edu.csus.ecs.pc2.api.exceptions.NotLoggedInException;
 import edu.csus.ecs.pc2.api.implementation.ProblemImplementation;
+import edu.csus.ecs.pc2.core.Utilities;
 import edu.csus.ecs.pc2.core.exception.SubmissionRejectedException;
 import edu.csus.ecs.pc2.core.model.IFile;
 import edu.csus.ecs.pc2.api.IClient;
@@ -455,7 +456,7 @@ public class TeamsController extends MainController {
 							run.isFinalJudged(),
 							run.isSolved(),
 							String.format("%s-%s", run.getSiteNumber(), run.getNumber()),
-							run.getScore()));
+							Utilities.formatScore(run.getScore())));
 				}
 				
 			}
--- a/projects/WTI-API/src/main/models/RunModel.java
+++ b/projects/WTI-API/src/main/models/RunModel.java
@@ -10,10 +10,10 @@ public class RunModel {
 	public long time;
 	public boolean isTestRun, isPreliminary, isFinal, isSolved;
 	public String id;
-	public double score;
+	public String score;
 
 	public RunModel(String teamID, String language, String problem, String judgement, List<File> result, long time, 
-			boolean isTestRun, boolean isPreliminary, boolean isFinal, boolean isSolved, String id, double score) {
+			boolean isTestRun, boolean isPreliminary, boolean isFinal, boolean isSolved, String id, String score) {
 		this.teamID = teamID;
 		this.language = language;
 		this.problem = problem;
--- a/projects/WTI-UI/src/app/modules/runs/components/new-run-alert/new-run-alert.component.html
+++ b/projects/WTI-UI/src/app/modules/runs/components/new-run-alert/new-run-alert.component.html
@@ -4,7 +4,7 @@
     <!-- <p>Type: {{ type }}</p> -->
     <p>Problem: {{ problem }}</p>
     <p>Judgement: {{ judgement }}</p>
-    <p *ngIf="showScore" class="score-animate">Score: {{ score | number:'1.3-3' }}</p>
+    <p *ngIf="showScore" class="score-animate">Score: {{ score }}</p>
     <p *ngIf='isPreliminary' style="color:red">This is a preliminary judgement!</p> 
 </div>
 <div mat-dialog-actions>
--- a/projects/WTI-UI/src/app/modules/runs/components/runs-page/runs-page.component.html
+++ b/projects/WTI-UI/src/app/modules/runs/components/runs-page/runs-page.component.html
@@ -88,7 +88,7 @@
 					<ng-container *ngIf="isPassFail"> {{ run.time }} </ng-container>					
 					<ng-container *ngIf="isPointScoring"> 
 						<ng-container *ngIf="hasAcceptedJudgement(run); else noScore"> 
-							{{ run.score | number:'1.3-3' }} <!-- format numbers with at least one whole digit and exactly 3 decimal digits -->
+							{{ run.score }}
 						</ng-container> 
 						<ng-template #noScore> 
 							<span title="Score shown only for accepted runs">--</span>
--- a/projects/WTI-UI/src/app/modules/scoreboard/components/scoreboard-page/scoreboard-page.component.html
+++ b/projects/WTI-UI/src/app/modules/scoreboard/components/scoreboard-page/scoreboard-page.component.html
@@ -61,7 +61,7 @@
   							</ng-container>
 
   							<ng-container *ngIf="isPointScoring">
-    							{{ problem.attempts }}/{{ problem.score | number:'1.3-3' }} <!-- At least one whole digit; exactly 3 fractional digits -->
+    							{{ problem.attempts }}/{{ problem.score }}
   							</ng-container>
 
 						</ng-container>
@@ -83,8 +83,8 @@
   				</ng-container>
   				
 				<ng-container *ngIf="isPointScoring"> 
-					<!-- display total score, number-formatted with at least one whole digit and exactly 3 decimal digits-->
-					<td style="text-align: center"> {{ team.score | number:'1.3-3' }} </td>
+					<!-- display total score -->
+					<td style="text-align: center"> {{ team.score }} </td>
 				</ng-container>
 				
             </tr>
--- a/src/edu/csus/ecs/pc2/clics/API202306/SubmissionService.java
+++ b/src/edu/csus/ecs/pc2/clics/API202306/SubmissionService.java
@@ -506,9 +506,9 @@ public class SubmissionService implements Feature {
                 return Response.status(Status.BAD_REQUEST).entity("invalid json supplied").build();
             }
 
-            // These next three are for admin users only
+            // These next two are for admin users only
             long overrideTimeMS = -1;
-            long overrideSubmissionID = -1;
+            long overrideSubmissionID = 0;
 
             Log log = controller.getLog();
             String user = sc.getUserPrincipal().getName();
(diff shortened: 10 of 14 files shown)


clevengr (commented):
I have reviewed src/edu/csus/ecs/pc2/clics/API202306/SubmissionService.java.  I don't understand what the changes in this file have to do with the PR to which they are attached. (The changes don't seem unreasonable -- they just appear to be related to a different issue...)

clevengr (commented):
I have reviewed src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java.  As with my previous comment, I don't see what the changes in this file have to do with this PR...

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -877,8 +876,7 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
             standingsRecordMemento.putLong("lastSolved", standingsRecord.getLastSolved());
 
             if(contestInformation.isScoreboardTypeScore()) {
-                DecimalFormat df = new DecimalFormat("0.0###");
-                standingsRecordMemento.putString("score", df.format(standingsRecord.getScore()));
+                standingsRecordMemento.putString("score", Utilities.formatScore(standingsRecord.getScore()));
```
clevengr, line 879: I see a potential problem here.  This change moves the specification of the exact desired format for the score which is to be put in `standingsRecordMemento` out of the local code and into a method in `Utilities`.  If future updates to the DefaultScoringAlgorithm wanted to have different formatting (for example, let's say there was a need for a "partial score" and a "total score"), the code wouldn't be able to do that using the `Utilities` method because that method only supports a _single_ specific format (whatever that format might be).  I'm not sure I agree that the specification of the formatting should be moved into a single Utility method... although I could be convinced otherwise if there was a strong benefit to doing so?

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
```diff
@@ -974,7 +972,7 @@ private void createStandingXML (TreeMap<StandingsRecord, StandingsRecord> treeMa
                     psiMemento.putString("shortName", problems[problemsIndexHash.get(psi.getProblemId())-1].getShortName());
                     psiMemento.putInteger("attempts", psi.getNumberSubmitted());
                     psiMemento.putInteger("points", psi.getPenaltyPoints());
-                    psiMemento.putDouble("score", psi.getScore());
+                    psiMemento.putString("score", Utilities.formatScore(psi.getScore()));
```
clevengr, line 975: This makes a substantial structural change to the Memento being created.  That Memento is used in a number of places; I'm struggling to see how to test all those places (or even to identify where those places are.

clevengr (commented):
I reviewed the changes to src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java.  I made a couple of notes: not so much "Request for changes" as "I'm not completely sure about doing this..."

src/edu/csus/ecs/pc2/core/scoring/NewScoringAlgorithm.java
```diff
@@ -576,8 +576,7 @@ private IMemento addTeamMemento(IMemento mementoRoot, IInternalContest contest,
         standingsRecordMemento.putLong("lastSolved", standingsRecord.getLastSolved());
 
         if(contestInformation.isScoreboardTypeScore()) {
-            DecimalFormat df = new DecimalFormat("0.0###");
-            standingsRecordMemento.putString("score", df.format(standingsRecord.getScore()));
+            standingsRecordMemento.putString("score", Utilities.formatScore(standingsRecord.getScore()));
```
clevengr, line 579: Same comments as I made regarding `DefaultScoringAlgorithm`...

clevengr (changes requested):
I've reviewed src/edu/csus/ecs/pc2/core/Utilities.java.  I think there is important error-handling missing (see my comments).

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -2126,5 +2136,26 @@ public static SerializedFile[] concatenateArrays(SerializedFile[] one, Serialize
...
+     * @return formatted score string
+     */
+    public static String formatScore(double score) {
+        
+        // first time through, create the formatter
+        if(dfPointScore == null) {
+            String scoreFormatString = IniFile.getValue(POINT_SCORING_FORMAT_INI_KEY);
+            if(StringUtilities.isEmpty(scoreFormatString)) {
+                scoreFormatString = DEFAULT_POINT_SCORING_FORMAT;
+            }
```
clevengr, line 2156: I believe this should be in a try/catch, to handle the possibility that IniFile.getValue() throws an exception.

Also, shouldn't there be some kind of validation that the returned scoreFormatString is a valid format?  Suppose the user puts an invalid format string in the INI file.  As I read the JavaDoc for `DecimalFormat(String)`, it says that `DecimalFormat` throws `IllegalArgumentException`.  I think this needs to be caught (and the Default Format used instead), with an appropriate log message.

clevengr:
I've reviewed all the file changes, made comments, and in one case requested change.  Moving on now to a runtime test. Noting that the "Steps to test" don't give much detail...

johnbrvc:
> I have reviewed src/edu/csus/ecs/pc2/clics/API202306/SubmissionService.java. I don't understand what the changes in this file have to do with the PR to which they are attached. (The changes don't seem unreasonable -- they just appear to be related to a different issue...)

Recall you had the same issue with the #1225 PR (empty score.txt file).    This was because I branched this PR off off an accepted PR branch that had not yet been merged into the  i1006_point_scoring branch, as such, it contains the files in that other (accepted) PR and they can be ignored.  Checking the [Commit List](https://github.com/pc2ccs/pc2v9/pull/1226/commits) will show this (this PR is only for the last commit.)

johnbrvc:
> I have reviewed src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java. As with my previous comment, I don't see what the changes in this file have to do with this PR...

Same reply as previous.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: Well, the `Utilities.formatScore()` is not a _single_ specific format... it is a configurable format.  If in the future, additional formats for different things are needed, the same mechanism can be used: add a new method to `Utilities `that does the necessary formatting, or, add an argument to `formatScore()` that takes an **enum** of the thing being formatted.  

The reason this was done is because several things format score values, not just DSA, and it made sense to make a single routine (single point of failure? ;-)) to do the formatting, rather than duplicate the same thing in several places.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: Does it really make a structural change to the Memento being created?  How so?  Can you think of a use case where it _would_ matter?  That is, a value formatted as a double precision number is being placed in an XML tag.  XML tags have no _type_ like JSON.  (eg.  `{ "score":"3.123"}` is different than `{ "score":3.123}` in JSON, but `<score>3.123</score>` is the same whether the memento is of type string or type double precision value.  Recall the DSA builds XML, not JSON - later, the XML is converted to JSON and any type conversion is done at that point.

The code using the memento should not care if the double precision value is formatted as: `3.123` or `3.123000000000`.  The point of the formatting the value is to make it so the code using the memento doesn't have to do it's own formatting.  Deep down, the **score** is still saved as a full double precision value, so, if a client needs the full precision, they can set the format in their own private `pc2v9.ini `file to be whatever they want to get the value they want or need.

src/edu/csus/ecs/pc2/core/scoring/NewScoringAlgorithm.java
johnbrvc: Same reply.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: `IniFIle.getValue()` can not throw any exceptions.  It's retrieving a `String`, which can be `null `if the key is not in the INI file.   `StringUtilities.isEmpty()` checks for `null`, that's why I used that to test the returned value.

Since I believe that part is a non-issue, onto the second part: possibly validating the format string.

I did think about this.  Personally, I'd rather know right away (via an _exception_) if my format was bad.  Blindly handling a bad format and replacing it with the default format can cause undesired surprises in output.  Even if a message is logged, no one will notice it.  An exception that travels back up the chain to the top will be noticed, I would expect, especially since no scoreboard would be generated.  I suppose one possible approach to dealing with this is: at client application startup time, when the INI file is first read, validation could be performed _there_, and then if the format is bad, abort the application startup altogether with an appropriate "**check the logs**" message.

Suggestions?

johnbrvc:
> I've reviewed all the file changes, made comments, and in one case requested change. Moving on now to a runtime test. Noting that the "Steps to test" don't give much detail...

I thought the testing procedure was terse, but adequate.  You don't have to load a new contest, or make submissions or anything.  Just load up a completed contest and look at the scoreboard, reports, event feed, WTI, whatever you want.  It's a visual thing.  BTW, we did use this at NAC for the WTI scoreboard (recall the "on-the-fly" `main.js` fix), the **pc2admin** scoreboards, the **pc2board** scoreboards and the event feed (resolver) and it appeared to work.  So, you could load up something like the `NAC2026-Challenge2024-test.zip` profile from here: [Point Scoring Profiles](https://drive.google.com/drive/folders/11i55pRYbh4rJB4Vj2pCAY2TkK9eINysB?usp=drive_link) and just look at the scoreboards and event feed as suggested in the PR, and make changes to the format in the `pc2v9.ini `files for each and restart the client as suggested in the PR.

If you would like me to supply step-by-step instructions on how to download the zip, unzip it, start the contest, test each app that uses the format stuff, make changes to the` pc2v9.ini `files, and repeat, I can surely do that.  I just thought a) we did all that "in production" at NAC and b) it's relatively straight forward,  that all that detail wasn't necessary.   Let me know and I'll be happy to put those steps above.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
clevengr: What I meant was that moving the selection of the desired formatting into `Utilities.formatScore()` removes the flexibility for different pieces of code to use different formats.  For example, suppose it was desirable to have the `DefaultScoringAlgorithm` class format the score differently from the `NewScoringAlgorithm` class  (I'm not saying that this particular case IS desireable; it's just an example -- and there might be anothe example, now or in the future, where that IS desireable).  With the existing (PR) code, that's not possible -- the format would have to be the same; i.e., it would have to be whatever is specified in the INI file which is read by the Utilities method.  It removes from the calling code the ability to define its own desired format.

However, I can see that the tradeoff is that putting it in a single (Utilities) method provides for increased flexibility in allowing the _user_ to define the format, so I guess it's a tradeoff.  But still think my earlier example holds:  what if a piece of client code (say, the DSA) wanted to format both a "partial score" (say, for a specific problem) AND a "total score" (say, for the aggregate score over all problems), and it wanted to use two different formats for them?  That wouldn't be possible with the PR structure; every formatted score would have to use the format read in from the INI file by the (first instantiation of) the Utilities class.

But, as I say, it's a tradeoff.  Not clear to me which is "better", so I'll go with the one proposed by the PR author.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
clevengr: One place where I think it COULD matter is where the XML gets converted to JSON and passed to clients like the WTI.  Currently the code which does that has to create JSON which matches the XML structure, and the Angular code on the WTI client has to know what that structure is supposed to look like.  It seems to me that changing the structure could break the linkage between the DSA and the Angular code (because the XML memento would have a field which is a String whereas the Angular code would be expecting a Number).  I'll have to test it to verify whether that's actually true or not...

src/edu/csus/ecs/pc2/core/scoring/NewScoringAlgorithm.java
clevengr: Yes.  Whatever resolution we apply to the DSA case will obviously need to hold for the NSA case as well.

src/edu/csus/ecs/pc2/core/Utilities.java
clevengr: Ok, I understand the logic about not replacing a bogus pattern with a default one (and expecting people to notice that in the logs).  I do like the idea of catching such a bogus pattern at initial startup and aborting startup. Unfortunately, `DecimalFormat` doesn't have anything like an `isValidFormat(String)` method.  Perhaps the following code (from my ChatGPT friend) could be used:
```
boolean isValidDecimalFormatPattern(String pattern) {
    try {
        new DecimalFormat(pattern);
        return true;
    } catch (IllegalArgumentException e) {
        return false;
    }
}
```
This could be called in the code encountering any `scoring.format` entry during initial INI processing, and aborting with a useful error message if it returns false.

clevengr:
> > I've reviewed all the file changes, made comments, and in one case requested change. Moving on now to a runtime test. Noting that the "Steps to test" don't give much detail...
> 
> I thought the testing procedure was terse, but adequate. You don't have to load a new contest, or make submissions or anything. Just load up a completed contest and look at the scoreboard, reports, event feed, WTI, whatever you want. It's a visual thing. BTW, we did use this at NAC for the WTI scoreboard (recall the "on-the-fly" `main.js` fix), the **pc2admin** scoreboards, the **pc2board** scoreboards and the event feed (resolver) and it appeared to work. So, you could load up something like the `NAC2026-Challenge2024-test.zip` profile from here: [Point Scoring Profiles](https://drive.google.com/drive/folders/11i55pRYbh4rJB4Vj2pCAY2TkK9eINysB?usp=drive_link) and just look at the scoreboards and event feed as suggested in the PR, and make changes to the format in the `pc2v9.ini `files for each and restart the client as suggested in the PR.
> 
> If you would like me to supply step-by-step instructions on how to download the zip, unzip it, start the contest, test each app that uses the format stuff, make changes to the`pc2v9.ini`files, and repeat, I can surely do that. I just thought a) we did all that "in production" at NAC and b) it's relatively straight forward, that all that detail wasn't necessary. Let me know and I'll be happy to put those steps above.

I personally don't need step-by-step details at the level of "how to down the zip, unzip it, start the contest, and so forth" (although it may be that some of our less-experienced testers might find that useful).  What I was commenting on was more about `Load up any point scoring contest that is complete, eg. NAC2026 challenge2024_test, for example.` -- but with no indication of how or where to get that from or how to use it.  You've answered the former question (where to get it from), and (implicitly) answered the latter question since what that download provides is a PC2 Profile for an already-run contest.    I was also sort of commenting on the fact that there were no instructions at all on what to do once that contest was loaded (in other words, it would have been helpful to point out that the intent was to repeatedly edit the `pc2v9.ini` file to contain different `scroring.format` strings and verify that they work as intended.   All in all it's no big deal; once you provided a link to the test contest it became obvious.

clevengr:
I have made a run-time test of the PR.  It works as expected for changing the `pc2v9.ini` file with regard to the Admin and Board clients.  However, I'm not able to get it to work with the WTI, even when I put
```
[scoring]
format=0.0
```
in the WTI's `pc2v9.ini` file.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: I still don't follow.   The JSON type will not change - it will be whatever it was before.  The XML will "look" the same to the JSON converter. ` DSA.getStandings()` returns an XML string.  The tag values have no type, well, technically I guess they are all **String**s.  The thing that takes the XML and converts it to JSON makes the determination what type to assign each JSON property.   Evidently, this is done when a `core.standings.TeamStanding` object is converted to JSON.  It should be noted, BTW, that ALL the fields in this class are **String** already.  It's my understanding that in `core.standings.ContestStandings` when the XML is converted to JSON ("`toString()`" method), the `ObjectMapper `looks at each XML string and determines what type it is based on its value.  So, a `String `of "1.123" will be a `double `as will a `String `of "1.123456789".

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: Exactly.  You have to instantiate to validate it.  I can add that.

johnbrvc:
> I have made a run-time test of the PR. It works as expected for changing the `pc2v9.ini` file with regard to the Admin and Board clients. However, I'm not able to get it to work with the WTI, even when I put
> 
> ```
> [scoring]
> format=0.0
> ```
> 
> in the WTI's `pc2v9.ini` file.

This took A WHILE to track down, and, just to be a blame-shifter, it's the fault of the Angular code writer... ;-)

For example, this from `/WTI-UI/src/app/modules/scoreboard/components/scoreboard-page/scoreboard-page.component.html`:
```
	<ng-container *ngIf="isPointScoring">
		{{ problem.attempts }}/{{ problem.score | number:'1.3-3' }} <!-- At least one whole digit; exactly 3 fractional digits -->
	</ng-container>
```
That is not very configurable then, is it?

Evidently, the scoreboard-page component overrides the format (with a lovely comment to that effect).  You may have noticed that while the WTI scoreboard had not obeyed the format in the `pc2v9.ini` file, it DID obey the rounding to one decimal place.   This is because the DSA returned the correctly formatted values, eg 63.0 instead of 63.001 or 491.7 instead of 491.675, but the angular code reformatted those as: 63.000 and 491.700.
image

This is why we didn't notice this at NAC since all the clients used a "3 after the decimal" format, which matched what the Angular HTML had. *sigh*.

As @lane55 was fond of saying, "this is why we test".  So, to blame-shift back to the PR author, ahem.  I should have checked formats other than "3 after the decimal" in my testing, and I would have seen the issue on the HTML and dealt with it.  

I made the change in the HTML angular files (and there were a bunch of those "number:'1.3-3'" formats scattered across  several files, and now it works - for the most part:
image

The issue is, since the value is a **double** NUMBER (not a **String**), when angular (javascript) goes to display it, it chops off any trailing zeroes as shown above.  I'm not sure how "bad" this is?  or good perhaps?  Hm.  I'm not sure how to address this since I don't know anything about how angular does stuff.  Maybe the app config file has to change and contain the format string:  number:'1.3-3' ?  Oh boy.  Can of worms.  Suggestions welcome from Angular experts.

(BTW I pushed a commit to remove the formatting from the HTML.  When you test it, be sure to flush your cache or use incognito mode since, "main.js" is not refreshed by default and it will not load a new one unless explicitly told to do so).

src/edu/csus/ecs/pc2/core/Utilities.java
```diff
@@ -105,6 +106,15 @@ public final class Utilities {
      * SimpleDateFormat for iso8601 including milliseconds
      */
     private static SimpleDateFormat iso8601formatterWithMS = new SimpleDateFormat(ISO_8601_TIMEDATE_FORMAT_WITH_MS);
+    
+    /**
+     * Allows changing of the point scoring score format from the default using the pc2v9.ini file
+     * and the key: format in the [scoring] section
+     */
+    private static final String POINT_SCORING_FORMAT_INI_KEY = "scoring.format";
+    private static final String DEFAULT_POINT_SCORING_FORMAT = "0.0##";
```
clevengr: The default pattern `0.0##` is problematic, in that it doesn't display all values in the same way.  For example, it will display a number whose value is 4.123456  as `4.123`, but it will display a number whose value is `4.000000` as `4.0` (only one trailing digit) -- because the `##` in the pattern indicates trailing zeroes should be omitted.  This is a problem because it means scores don't have uniform appearance in their scores, making the scoreboard a bit harder to read:

Image


For readability I think the default format pattern should display the same number of digits for all scores -- e.g. pattern `0.000`.

clevengr (changes requested):
Request change in the default score pattern, from `0.0##` to `0.000`, so all scores align.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: Agreed. Will make the change.

src/edu/csus/ecs/pc2/core/scoring/DefaultScoringAlgorithm.java
johnbrvc: This was discussed offline and determined to be a non-issue since the "type" of the value produced is still a String, whether its 1.23 or 1.23456789.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: Implemented in the latest push.

src/edu/csus/ecs/pc2/core/Utilities.java
johnbrvc: Change made in latest push.

johnbrvc:
This comment may be thought of as an addendum to the description of the PR.

It was determined that the WTI did not correctly format scores for scoring contests.  This was because during the conversion of XML to JSON in `XML.toJSONObject()`, the `score` property was converted to `double`, thereby losing any formatting done by the DSA.  Code was added (`changeScoresToStrings()`) to examine the generated JSON after the conversion from XML, and change all `score` properties (in **TeamStandings** and **ProblemSummaryInfo** objects) to strings instead of numbers.  To do this conversion, the score had to be _again_ formatted using the correct format (`Utilities.formatScore()`).  Yes, this is somewhat of a duplication of effort since the score WAS formatted properly by the DSA when generating the XML.  There was no reasonable way to preserve this information during the conversion to JSON without writing or using a different XML converter.  It was decided to post-process the JSON and undo the conversion of `score `to `double`.  This necessitated reformatting the value.

In addition, the "**Runs**" table on the WTI web page improperly showed the score for problems as well.  They used to be hardcode formatted as 1.3-3 (three after the decimal), and that was not configurable.  When the fixed formatting was removed, default Java formatting of the `double `value was performed leading to scores with many places after the decimal.  This was addressed by changing the `RunModel ` class to make the type of the `score `property be `String `instead of `double`.  Then, in the WTI-API, the value was formatted using `Utilities.formatScore() `(just like the DSA and `RunsTablePane `in PC^2 do).  This newly formatted value was then set to the `String score `property in the `RunModel `(constructor changed).  No other modules appear to use the `score `property besides the **Run** table so this was safe to change the type of score.

The changes described above appear to fix the issues related to formatting scores in the WTI.

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -930,11 +932,116 @@ private String getJSONStandings(String xmlStandings) throws IOException, JSONExc
...
+        if(teamStandings instanceof JSONArray) {
+        	JSONArray jArray = (JSONArray)teamStandings;
+        	int nStandings = jArray.length();
+        	for(int idx = 0; idx < nStandings; idx++) {
+                fixProblemSummaryInfo(jArray.optJSONObject(idx));
+            }
+        } else if(teamStandings instanceof JSONObject){
+            // Only one teamStandings
+            fixProblemSummaryInfo((JSONObject) teamStandings);
+        }
```
clevengr, line 996: It bothers me somewhat that the above block of code checks for an array of JSON objects and also checks for a single JSONObject, but it doesn't deal with the condition that the teamStandings Object might be neither of those.  In the previous code block (where it pulls the ContestStandings out of the JSON), it checks to see if the returned object is null:
```
        if (currentObj == null) {
            logger.warning(CONTEST_STANDINGS + " object not found in standings XML");
            return;
        }
```
It seems like a similar error-handling "else" condition should be applied in this block of code as well -- unless you think there's a reason that it's _not possible_ for the returned object to be anything but a JSONArray or a JSONObect.  But that depends on the correctness of the external code which generated the XML, along with the correctness of the external code which converts the XML to JSON.  For that reason I think a sanity check here is warranted.

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -930,11 +932,116 @@ private String getJSONStandings(String xmlStandings) throws IOException, JSONExc
...
+        if(psis instanceof JSONArray) {
+        	JSONArray jArray = (JSONArray)psis;
+        	int nStandings = jArray.length();
+        	for(int idx = 0; idx < nStandings; idx++) {
+                fixScore(jArray.optJSONObject(idx));
+            }
+        } else if(psis instanceof JSONObject) {
+            fixScore((JSONObject)psis);
+        }
+    }
```
clevengr, line 1023: Same comment as above: it seems like there should be an `else` on the above `if.. else if...` block.  It's more problematic here because you can't just "return"; there needs to be a way to log a message and notify the caller that things didn't go well...

projects/WTI-API/src/main/controllers/ContestController.java
```diff
@@ -930,11 +932,116 @@ private String getJSONStandings(String xmlStandings) throws IOException, JSONExc
...
+     * @param psi - the ProblemSummaryInfo of interest
+     */
+    void fixScore(JSONObject psi) {
+        String SCORE_PROPERTY = "score";
+        
+        try {
+            double score = psi.getDouble(SCORE_PROPERTY);
+            String formattedScore = Utilities.formatScore(score);
+            // Overwrite existing property with string version for formatted score.
+            psi.put(SCORE_PROPERTY, formattedScore);
```
clevengr, line 1038: It's not clear to me how this solves the problem we were trying to fix -- specifically, that the JSON entry for "score" needs to be a String (meaning there needs to be quotes around the numeric values of the score in the JSON).  However, the JavaDoc for the `org.json.JSONObject` class doesn't explicitly provide a `put` method with the signature `(String,String)`, so the above line of code must be invoking the method with signature `(String,Object)`.  It's not clear to me that this necessarily implies that the method will add quotes around the specified value object.  (I guess if it works then it must be doing that??)

clevengr (changes requested):
I reviewed the changes to `projects/WTI-API/src/main/controllers/ContestController.java`.  I noted two areas where I think extra error-catching code is warranted/needed.

projects/WTI-API/src/main/controllers/ContestController.java
johnbrvc: Why do we care?  If it's not a JSONObject nor an array, what's there to do?  Nothing.  I just don't think we care about that.  In fact, it may not be there at all, which I suppose is possible if there are no teams configured in the contest.  The ContestStandings thing is different.  There HAS to be one of those or something is seriously wrong.

projects/WTI-API/src/main/controllers/ContestController.java
johnbrvc: Again, I don't think we care if there are no ProblemSummaryInfo's.  This can happen if there are no problems configured I suppose, but again who cares.  We are just massaging "score" properties.  If there aren't any, we don't care - it would do what it has always done before.  It's not as if we are "breaking" something by not performing a substitution on something that isn't there in the first place.

projects/WTI-API/src/main/controllers/ContestController.java
johnbrvc: The sample code is in the description in the javadoc: [Link to JavaDOC](https://stleary.github.io/JSON-java/org/json/JSONObject.html), this is pretty much what I was doing; instead of a String constant, I used a variable.


```
 myString = new JSONObject()
         .put("JSON", "Hello, World!").toString();
 
produces the string {"JSON": "Hello, World"}.

```
It produces exactly what we want.

clevengr: Oops, I didn't see that example code in the JavaDoc.  I agree; marking this Resolved.

clevengr (commented):
I reviewed all the changed files and don't have any further issues with them.  I ran a pretty comprehensive runtime test.  I verified that formatting changes are properly handled in the PC2 Admin Scoreboards, the PC2 Board client and the HTML scoreboards it generates, and the WTI.   I was not able to figure out exactly how to verify that formatting changes are properly reflected in the Event Feed (there's no Scoreboard data in the Admin's `Event Feed JSON` report or the EF Client `View Event Feed` operation, and for some reason I couldn't connect to/login to the contest using a browser to hit the `/event-feed` endpoint myself.

My failure is undoubtedly my own fault, but I'd like to finish that EF verification - if I can get some help understanding how to do it.  Once I've done that I'm ready to approve the PR.

(Note: since I like to create "Test Procedure" documents to help myself keep track of how I'm testing, I took the liberty of going ahead and adding my own steps to the "How to Test the PR" section...  Hopefully it might help someone else approve the PR more quickly...)

johnbrvc:
> I reviewed all the changed files and don't have any further issues with them. I ran a pretty comprehensive runtime test. I verified that formatting changes are properly handled in the PC2 Admin Scoreboards, the PC2 Board client and the HTML scoreboards it generates, and the WTI. I was not able to figure out exactly how to verify that formatting changes are properly reflected in the Event Feed (there's no Scoreboard data in the Admin's `Event Feed JSON` report or the EF Client `View Event Feed` operation, and for some reason I couldn't connect to/login to the contest using a browser to hit the `/event-feed` endpoint myself.
> 
> My failure is undoubtedly my own fault, but I'd like to finish that EF verification - if I can get some help understanding how to do it. Once I've done that I'm ready to approve the PR.
> 
> (Note: since I like to create "Test Procedure" documents to help myself keep track of how I'm testing, I took the liberty of going ahead and adding my own steps to the "How to Test the PR" section... Hopefully it might help someone else approve the PR more quickly...)

Remember, when testing the event feed, keep in mind the CLICS spec says that the "score" values are _numbers_, meaning,  they are not _strings_.  The formatting is not preserved, of course.  However, the rounding is observed since it works kind of the same was as the WTI did.  XML standings are created and the scores are formatted according to the `[scoring].format` in the event feeder's `pc2v9.ini`.  When converted to JSON for the event feed endpoints, that formatted string is converted to a number.  So, for example, "100.0" will show up as 100 on the scoreboard endpoint.    That being said, I'm not sure exactly how you tested the Event Feed scoreboard endpoint But here's the procedure:

1. Start feeder 1 client (pc2ef --login ef1), the press the **Start** button to start the feeder.
2. From a browser:  [https://localhost:50443/contests/nac25chall24practice/scoreboard](https://localhost:50443/contests/nac25chall24practice/scoreboard) when it prompts for user/password, use: **adminsitrator1:administrator1**, or **admin:admin** (from the` realm.properties` file, assuming you have one).
3. From the command line curl: `curl -k https://admin:admin@localhost:50443/contests/nac25chall24practice/scoreboard`
        (you can also use **administrator1:administrator1**.  If you're using a pc2 account it must have the **_View Event Feeds_** permission. )

clevengr:
> 
> Remember, when testing the event feed, keep in mind the CLICS spec says that the "score" values are _numbers_, meaning, they are not _strings_. The formatting is not preserved, of course. However, the rounding is observed since it works kind of the same was as the WTI did. XML standings are created and the scores are formatted according to the `[scoring].format` in the event feeder's `pc2v9.ini`. When converted to JSON for the event feed endpoints, that formatted string is converted to a number. So, for example, "100.0" will show up as 100 on the scoreboard endpoint. That being said, I'm not sure exactly how you tested the Event Feed scoreboard endpoint But here's the procedure:
> 
> 1. Start feeder 1 client (pc2ef --login ef1), the press the **Start** button to start the feeder.
> 2. From a browser:  https://localhost:50443/contests/nac25chall24practice/scoreboard when it prompts for user/password, use: **adminsitrator1:administrator1**, or **admin:admin** (from the` realm.properties` file, assuming you have one).
> 3. From the command line curl: `curl -k https://admin:admin@localhost:50443/contests/nac25chall24practice/scoreboard`
>    (you can also use **administrator1:administrator1**.  If you're using a pc2 account it must have the **_View Event Feeds_** permission. )

Thanks.  I'm not quite sure why the URL contains `nac25chall24practice` when the Contest Title is `2026 ICPC NAC Challenge 2024 Test`, but I do see that it works as described.

clevengr (approved):
I've now verified that the Event Feed contains scores in the expected format, so as far as I can see everying in the PR works as intended. I approve the PR.

johnbrvc:
> > Remember, when testing the event feed, keep in mind the CLICS spec says that the "score" values are _numbers_, meaning, they are not _strings_. The formatting is not preserved, of course. However, the rounding is observed since it works kind of the same was as the WTI did. XML standings are created and the scores are formatted according to the `[scoring].format` in the event feeder's `pc2v9.ini`. When converted to JSON for the event feed endpoints, that formatted string is converted to a number. So, for example, "100.0" will show up as 100 on the scoreboard endpoint. That being said, I'm not sure exactly how you tested the Event Feed scoreboard endpoint But here's the procedure:
> > 
> > 1. Start feeder 1 client (pc2ef --login ef1), the press the **Start** button to start the feeder.
> > 2. From a browser:  https://localhost:50443/contests/nac25chall24practice/scoreboard when it prompts for user/password, use: **adminsitrator1:administrator1**, or **admin:admin** (from the` realm.properties` file, assuming you have one).
> > 3. From the command line curl: `curl -k https://admin:admin@localhost:50443/contests/nac25chall24practice/scoreboard`
> >    (you can also use **administrator1:administrator1**.  If you're using a pc2 account it must have the **_View Event Feeds_** permission. )
> 
> Thanks. I'm not quite sure why the URL contains `nac25chall24practice` when the Contest Title is `2026 ICPC NAC Challenge 2024 Test`, but I do see that it works as described.

Ugh... Yeah, when I first created the "nac25chall24practice" it was in January (21) 2026.  I guess my mind was still on "2025" so  when I created the contest.yaml with the contest id, I put in nac25chall24practice.  It's one of those "historical" mistakes that never got corrected.

JoeTerlizzi (approved):
I have followed the testing procedure and everything works as described.

Note that when using the curl command whole number scores are represented by only 1 decimal regardless of configuration while partial scores have the correct number of decimal places.  This is expected.

Approved

PR #1263, opened 2026-04-20
i_1262 Improve sandbox scripts for better reporting CPU time accuracy

### LINKED ISSUE #1262: Check CPU time for submissions judged in a sandbox
**Describe the issue**:
When using the `pc2sandbox*.sh` scripts to execute contestant submissions, it is sometimes observed that the CPU times reported by the **cgroup**  `cpu.stat` file is slightly (10-40ms) longer than they should be.     

**To Reproduce**:

1. This is only for Linux based systems since sandboxes are currently supported on Linux only.
2. Compile the `samps/src/hello.cpp` program.
3. Execute it from a Linux `bash `prompt using the Linux "`time`" program: `time ./a.out`
4. Note that it takes on the order of a few milliseconds to run.
5. Start up a clean PC2 server using the **clics_sumithello** sample contest.
6. Start up an administrator client.
7. Edit the "**hello**" problem and change it to use a **Sandbox**.
8. Start the contest.
9. Start up an autojudge.
10. Start up a **pc2team** client and log in using **team1** with password **team1**.
11. Submit the `samps/src/hello.cpp` program for the "**hello**" problem.
12. Wait for the submission to be judged.
13. From a `bash `command prompt, change to the "`execute`" folder where the judge executed the submission.
14. View/edit the `sandbox.log` file.
15. Search for the string "**CPU ms**" - this is the header of 2 line table showing the resources used by the submission.
16. Note the CPU time below the **CPU ms** heading is on the order of 10's of ms, depending on your hardware.

**Expected behavior**: 
The CPU time should be on the order of a few milliseconds, not 10's of milliseconds.

**Actual behavior**:
The CPU time is on the order of 10's of ms.

**Environment**:
Linux Ubuntu 24.04.3 (AWS EC2 m5.2xlarge)

**Log Info**: 

**Screenshots**: 

**Additional context**: 
This happens because the setup CPU resources for the submission executable is included in the **cgroup**.  The scripts should be changed so that only the executable itself and minimal required setup is in the **cgroup**.

johnbrvc:
Both the `pc2sandbox.sh` and` pc2sandbox_interactive.sh` scripts have been modified to only put the bare minimum amount of work in the **cgroup**.  Previously, some debug logging and other bash related stuff was charged to the **cgroup**.
CI: Use _CPU #_ (0-(N-1)) instead of ordinal (1-N) to determine the CPU to pin to.  Change `taskset `to use the `-c CPU#` option instead of the default "mask" option for clarity.
CI: better handling of error conditions in the `pc2sandbox_interactive.sh` script to create the resource summary report even on failures and PC2 wall time limit exceeded.  Previously, resource were not created in these cases.  (This makes the _Web Judge Interface_ better).
CI: Add `TDEBUG `flag to control timing debugging.  It is ON by default and logs stuff to the `sandbox.log` file.  It's mostly keeping track of the **cgroup** `cpu.stat` file.

Fixes #1262 
Linux 24.04.3

1. This is only for Linux based systems since sandboxes are currently supported on Linux only.
2. Compile the `samps/src/hello.cpp` program.
3. Execute it from a Linux `bash `prompt using the Linux "`time`" program: `time ./a.out`
4. Note that it takes on the order of a few milliseconds to run.
5. Start up a clean PC2 server using the **clics_sumithello** sample contest.
6. Start up an administrator client.
7. Edit the "**hello**" problem and change it to use a **Sandbox**.
8. Start the contest.
9. Start up an autojudge.
10. Start up a **pc2team** client and log in using **team1** with password **team1**.
11. Submit the `samps/src/hello.cpp` program for the "**hello**" problem.
12. Wait for the submission to be judged.
13. From a `bash `command prompt, change to the "`execute`" folder where the judge executed the submission.
14. View/edit the `sandbox.log` file.
15. Search for the string "**CPU ms**" - this is the header of 2 line table showing the resources used by the submission.
16. Note the CPU time below the **CPU ms** heading is about the same as that observed in step 4.

### DIFF
--- a/scripts/pc2sandbox.sh
+++ b/scripts/pc2sandbox.sh
@@ -375,7 +375,7 @@ else
   echo "max" > $PC2_SANDBOX_CGROUP_PATH/memory.swap.max  
 fi
 
-# Create variables for various ulimit values.  These will be set in once inside the cgroup.
+# Create variables for various ulimit values.  These will be set once inside the cgroup.
 # We use ulimit to limit CPU time, not cgroups.  Time is supplied in seconds.  This may have to
 # be reworked if ms accuracy is needed.  The problem is, cgroups do not kill off a process that
 # exceeds the time limit, ulimit does.
--- a/scripts/pc2sandbox.sh
+++ b/scripts/pc2sandbox.sh
@@ -13,6 +13,7 @@
 #  $7... : command arguments
 # 
 # Author: John Buck, based on earlier versions by John Clevenger and Doug Lane
+
 DEFAULT_CPU_NUM=3
 CPU_OVERRIDE_FILE=$HOME/pc2_cpu_override
 
@@ -43,7 +44,7 @@ MAXPROCS=32
 
 # Compute taskset cpu mask for running submission on single processor
 
-# Get system's maximum CPU number
+# Get system's maximum CPU number - numbered starting at 0
 MAX_CPU_NUM=`lscpu -p=cpu | tail -1`
 
 # See if the admin wants to override the CPU by reading the override file
@@ -71,13 +72,11 @@ then
 	if [[ "$cpunum" =~ ^[1-9][0-9]*$ ]]
 	then
 		# Restrict to number of CPU's.
-		cpunum=$(((cpunum-1)%(MAX_CPU_NUM+1)))
+		cpunum=$((cpunum%(MAX_CPU_NUM+1)))
 	else
-		cpunum=$(((DEFAULT_CPU_NUM+1)))
+		cpunum=$DEFAULT_CPU_NUM
 	fi
 fi
-cpunum=$((cpunum-1))
-CPUMASK=$((1<<cpunum))
 
 # Process ID of submission
 submissionpid=""
@@ -108,6 +107,13 @@ DEBUG_FILE=sandbox.log
 function DEBUG()
 {
   [ "$_DEBUG" == "on" ] && "$@" >> $DEBUG_FILE
+}
+
+# control whether the script outputs debug info for timing details
+_TDEBUG="on"   # change this to anything other than "on" to disable debug/trace output
+function TDEBUG()
+{
+  [ "$_TDEBUG" == "on" ] && "$@" >> $DEBUG_FILE
 }
 
 # For per-testcase reporting/logging
@@ -369,19 +375,17 @@ else
   echo "max" > $PC2_SANDBOX_CGROUP_PATH/memory.swap.max  
 fi
 
+# Create variables for various ulimit values.  These will be set in once inside the cgroup.
 # We use ulimit to limit CPU time, not cgroups.  Time is supplied in seconds.  This may have to
 # be reworked if ms accuracy is needed.  The problem is, cgroups do not kill off a process that
 # exceeds the time limit, ulimit does.
 TIMELIMIT_US=$((TIMELIMIT * 1000000))
 REPORT_DEBUG Setting cpu limit to $TIMELIMIT_US microseconds "("ulimit -t $TIMELIMIT ")"
-ulimit -t $TIMELIMIT
 
 MAXPROCS=$((MAXPROCS+`ps -T -u $USER | wc -l`))
 REPORT_DEBUG Setting maximum user processes to $MAXPROCS
-ulimit -u $MAXPROCS
 
 REPORT_DEBUG Setting stack size to unlimited
-ulimit -s unlimited
 
 # Keep track of details for reports
 REPORT_BRIEF ${JUDGEIN}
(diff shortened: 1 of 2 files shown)


scripts/pc2sandbox.sh
```diff
@@ -71,13 +72,11 @@ then
 	if [[ "$cpunum" =~ ^[1-9][0-9]*$ ]]
```
clevengr, line 72: It appears to me that the regex in this line requires cpunum to start with a digit between 1 & 9.  That excludes CPU 0.  I don't understand why.

scripts/pc2sandbox.sh
```diff
@@ -369,19 +375,17 @@ else
   echo "max" > $PC2_SANDBOX_CGROUP_PATH/memory.swap.max  
 fi
 
+# Create variables for various ulimit values.  These will be set in once inside the cgroup.
```
clevengr: I don't understand the second sentence in this comment.  Is there an extra word, "in", that doesn't belong there?

scripts/pc2sandbox.sh
```diff
@@ -110,6 +109,13 @@ function DEBUG()
   [ "$_DEBUG" == "on" ] && "$@" >> $DEBUG_FILE
 }
 
+# control whether the script outputs debug info for timing details
+_TDEBUG="on"   # change this to anything other than "on" to disable debug/trace output
+function TDEBUG()
+{
+  [ "$_TDEBUG" == "on" ] && "$@" >> $DEBUG_FILE
```
clevengr, line 116: Should this perhaps also capture stderr as well as stdout?  That might be relevant...

scripts/pc2sandbox.sh
```diff
@@ -390,49 +394,63 @@ REPORT_BRIEF $cpunum
...
 # Wait for child
 wait $submissionpid
-
 COMMAND_EXIT_CODE=$?
 
+# Get cpu time used by submission
+cputime=`grep usage_usec $PC2_SANDBOX_CGROUP_PATH/cpu.stat | cut -d ' ' -f 2`
+
+# Get wall time - we want it close to when we fetch the sub-process (submission) completes
+endtime=`GetTimeInMicros`
```
clevengr, line 439: It appears to me that computing `endtime` here (after the calculation of `cputime`) means that the time taken by the `grep` and the `cut` in the preceeding statement will be included in the process' execution time.  Am I wrong? Is that intentional?  (It seems like computing `endtime` should be done before the computation of `cputime` -- and possibly even before the saving of `COMMAND_EXIT_CODE`; i.e. immediately after the `wait` is finished...)

clevengr (commented):
I reviewed scripts/pc2sandbox.sh and left several questions. I also reviewed the interactive sandbox version; I have the same basic questions about that, although I admit I don't understand that code quite as well.

scripts/pc2sandbox.sh
johnbrvc: In that block of code, the script is trying to "guess" which CPU to use based on the logged in user's account ($USER).  It will isolate the judge # from the logged in user name, eg. judge1 -> 1.  Since we do not allow "judge0",  we'd want that to use the default.  This, code is used, for example, in non-WF type events, if you have multiple judges running on a multi-core system.  Each judge can be pinned to it's own CPU based on its judge number, eg. judge1 -> cpu1, judge2 -> cpu2, etc.

scripts/pc2sandbox.sh
johnbrvc: Probably not.  TDEBUG is basically a debug print.  If any of the echo's or grep's it does fail, the errors will be sent to the pc2 validator standard error file by default.

scripts/pc2sandbox.sh
johnbrvc: No the times are NOT included in the process' execution time.  That was the point of the moving the code around.  Line 416-425 put a SUBSHELL (`$BASHPID`) only in the cgroup:

```
# This will create a new process group and put it into the created cgroup
( if ! echo $BASHPID > $PC2_SANDBOX_CGROUP_PATH/cgroup.procs
  then
    echo $0: Could not add current process to $PC2_SANDBOX_CGROUP_PATH/cgroup.procs - not executing submission.
    SysFailure Could not add current process to $PC2_SANDBOX_CGROUP_PATH/cgroup.procs
    exit $FAIL_SANDBOX_ERROR
  fi
  ulimit -t ${TIMELIMIT} -u ${MAXPROCS} -s unlimited
  exec setsid taskset -c ${cpunum} $COMMAND $*
) <&0 >&1 2>&2 &
```
Hence the `( ...) &`.    Parens in bash create a subshell.  So, process `$! `is the PID of the created subshell in the parens, which we assign to `submissionpid`.  (`$submissionpid` is the process id that is in the **cgroup**, NOT the current shell.)  The only thing charged to the process' cpu time is `ulimit -t ...` and the "`exec setsid taskset`", all of which are necessary and have to be done before firing off the submission's executable in the context of the new submissions PID to set the time limit and session ID and pin to a specific CPU.

We DO charge the wall-time used by `grep usage_usec` from `$PC2_SANDBOX_GROUP_PATH/cpu.data` to the submission, but since we don't use wall time for anything other than display, it's not really important to worry about that.  I guess technically, we could move the:
~~~
endtime=`GetTimeInMicros`
~~~
to right before the:
~~~
cputime=`grep ...`
~~~
but again, it's not important.

scripts/pc2sandbox.sh
johnbrvc: That "in" is left over.  I think I originally was going to say, "These will be set in the cgroup.", but changed my mind to "once inside the cgroup"; forgot to remove extra word.  Fixed in latest push.
https://github.com/pc2ccs/pc2v9/pull/1263/changes/f03b7c0020928a204ce7e97aa334bf72b5fb905b

scripts/pc2sandbox.sh
clevengr: This was exactly my point: it seems like `endtime=GetTimeInMicros` should come _before_ `cputime=grep ...`  Otherwise, as you seem to agree, the code is charging the time to execute the `grep` (and also the `cut`, I think?) to the execution time of the submission.  It's not clear to me why you think it's "not important"; doesn't that add inaccuracy to the computed execution time?

clevengr (commented):
I've added some comments in reply to your answer to my original comments.  I still haven't had time to set up a Linux environment to do a runtime test.

scripts/pc2sandbox.sh
johnbrvc: `endtime `is the WALL time; which is NOT that important.  `cputime `is the important one.  If wall time is off by, say, 5ms, it wont make any difference to anyone.  Wall is affected by many things; system load, disk I/O, etc., all of which are not charged to the CPU time of the run since we only look at the real CPU time from the cgroup.  We only use wall time as a failsafe, eg. if some contestant does "sleep(10);"  That's what those thresholds are for that we added to the Settings (so-called "grace periods").  Since sleep(10) uses no cpu time, the pc2 walltime timers will kick in and kill it.

Now, this IS different from the way that PC^2 USED to work when we only used wall time to measure CPU time (and, the way it does it if sandboxes aren't used).  Now, wall time is just for information purposes and really isn't used (unless the failsafe trips as described above).

PR #1275, opened 2026-05-12
I1251 wti scoreboard groups

### LINKED ISSUE #1251: WTI should allow selection of Scoreboard based on Groups
Yes -- the WTI currently only allows a team to see one "scoreboard" -- that of the entire contest.  In a contest with multiple groups (e.g. the North America East Division), teams are much more likely to want to see the scoreboard _just for their group (division)_, since that is what determines advancment to the NAC.

**Feature Description**: 
Add to the WTI "Scoreboard" component the ability to select which "Group" scoreboard should be displayed.  Include the following:
-  ability for the Admin to configure, both via YAML and via the Admin Settings GUI, which groups are visible (selectable) on the WTI screen.  
- By default arrange that groups containing no teams (e.g. "Admin-Site", "ICPC", etc.) are NOT displayed/selectable. 
- Retain the ability for a team to see the "Overall Combined Scoreboard" (that is, the scoreboard which the WTI already displays, showing ALL teams).
- Ensure that when a group scoreboard is selected, the rankings showing in the scoreboard are group-specific (in other words, the team at the top of the scoreboard for the group should be shown as Rank 1, regardless of their current standing in the overall contest; the second team in the group is shown as Rank 2; etc.).

Nothing comes to mind.

Add a drop-down list on the WTI "Scoreboard-page" component that allows selection of which scoreboard to display.  Then, several approaches come to mind:
  - add filtering to the scoreboard-page component `loadStandings()` method to filter standings based on the selected group (this has the potential problem of the rankings still being "overall" instead of "per group", and would require the method to adjust the rankings on the fly).
  - add filtering to the `getTeamStandingsArray()` method (this also still has the problem of incorrect group-based rankings).
  - update the `ContestService.getStandings()` method to respect the selected group. This is complicated by the fact that this method invokes an HTTP request to get the standings; somehow either the HTTP request would have to know how to return only the desired-group standings, or else the method would have to filter the returned result (and deal with incorrect group-specific rankings).  It is further complicated by the fact that the ContestService maintains a cached copy of the standing (which is returned if there has been no standings-changing event).
  - consider whether the DSA `useDivisionFilter` facility could be used to ensure standings always contain the desired group.

clevengr:

Adds support for selecting different groups on the WTI scoreboard.  Specifically,
- Updates WTI-UI `scoreboard-page.component.{ts,html,sccs}` so that if there is more than one `group` containing teams in the contest, a dropdown list for group selection appears on the WTI Scoreboard page; selecting a group from the dropdown causes the scoreboard to display (only) teams in the selected group, ranked from 1..N within the selected group.
- Updates WTI-API `ContestController` so that it augments the JSON returned to a `scoreboard` endpoint query to include all groups of which the requesting team is a member (the existing implementation only provides a team's "primary group", which is not sufficient to determine what teams to display if teams can be members of more than one group).

### What the PR does NOT do:
- Does not provide support for contests set up with "divisions" (for example, PacNW contests).  That is, the PR only supports contests where there is no `wtiBoardUseDivisions` entry in the WTI `pc2v9.ini` file.  Support for divisions will be handled via work under a separate Issue:  #689.

Fixes #1251

Windows 11, java version "1.8.0_381", Chrome Version 148.0.7778.96 (Official Build) (64-bit)

#### Setup
- Download and unzip the PR distribution (or build it on your own machine from the branch).

#### Verify that the WTI Scoreboard works correctly when there are multiple displayable groups but a given team only appears in single group.
1. Download and unzip the `WF49BakuFinalized.zip` file from https://drive.google.com/drive/folders/1GPl7c8MzmhSUZ905vvPkpE59FR_Y4EwC?usp=drive_link. This is a PC2 profile for the WF Baku contest, in which there are multiple displayable groups but each team is a member of only one group.
1. Copy the `/profiles` folder and the `profiles.properties` file into the root of the unzipped PC2 distribution.
1. Start a PC2 server and PC2 Admin.
1. Verify that the Admin shows accounts, problems, etc. from the proper contest.
1. On the PC2 Admin `ConfigureContest>Accounts` tab, edit the `scoreboard2` account:  change the password to "scoreboard2" (an alternative would be to update the WTI pc2v9.ini file to specify the existing scoreboard2 password prior to starting the WTI Server).
1. In the PC2 distribution "projects" folder, unzip the `WTI-Interface.zip` file.
1. In the resulting WTI-Interface folder, start the WTI Server (`./bin/pc2wti`).
1. Verify that the last line on the console indicates the server was started.
1. Open a browser. Either clear the browser's cache, or open an "Icognito mode" browser window.
1. Connect to the WTI server (typically by opening http://localhost:8080).
1. Login as a team.  You can use any team in the contest, but you'll have to look on the PC2 Admin Accounts tab, select a team, and Edit the team to get the correct password.
1. Select the `Scoreboard` menu on the WTI.
1. Verify that a `Group` dropdown list appears at the top of the scoreboard, and that the default (selected) entry is "All teams".
1. Verify that the rankings shown on the WTI scoreboard match those shown on the PC2 Admin `RunContest>Standings` screen.
1. On the WTI Scoreboard `Groups` dropdown, select a specific group.
1. On the PC2 Admin `RunContest>Standings` screen, click the checkbox in the right-side panel that matches the selected group.
1. Verify that standings shown in the WTI Scoreboard match those shown on the PC2 Admin for the same group.
1. If desired (just to convince yourself), select different groups on the WTI and PC2 Admin and verify they match.

#### Verify that the WTI Scoreboard works correctly when there are multiple displayable groups and a given team appears in multiple groups.
1. Shut down the WTI browser and the WTI Server.
1. On the PC2 Admin `Configure Contest` screen, `Accounts` tab, select any team.  Notice what Group is listed for that team (in the `Groups` column).
1.  Click the `Edit` button, and scroll down in the `Groups` pane and verify that the correct Group (from the previous step) is checked.
1.  Add a checkmark for a DIFFERENT GROUP -- but make sure it is one of the actual groups for this contest (Africa&Arab, Asia East, Asia West, Asia Pacific, Europe, Latin America, Northern Eurasia, or North America).    Do NOT disable the checkmark which already existed.  (The effect of this is that you have put this particular team in multiple groups).
1. Repeat the above step for an additional one or two teams -- so that you have several teams in multiple different groups.
1. Restart the WTI server, open a browser, login as a team, and select the WTI "Scoreboard" menu.  
1. Select each of the groups in the dropdown "Group" list; verify that each team you assigned to multiple groups does appear in that group's scoreboard (and also that each team still appears in its original group scoreboard).
1. Verify that each group's WTI scoreboard matches the PC2 Admin Standings for that group.

#### Verify that the WTI Scoreboard works correctly when there is exactly one displayable group and all teams are in that group.
1. Shut down the WTI browser, the WTI Server, the PC2 Admin, and the PC2 server.
1. In the PC2 distribution root, DELETE the `profiles` and `logs` folders, and the `profiles.properties` file.  
1. In the PC2 distribution `projects` folder, DELETE the `WebTeamInterface` folder (but not the `.zip` file from which that folder was extracted).  This puts you back to a clean PC2 installation.
1. Download and unzip the `GNY2025Finalized.zip` file from https://drive.google.com/drive/folders/1GPl7c8MzmhSUZ905vvPkpE59FR_Y4EwC?usp=drive_link.  (This is a contest with just one displayable group.)
1. Copy the `/profiles` folder and the `profiles.properties` file into the root of the unzipped (clean) PC2 distribution.
1. Follow steps 3-12 (only) above to get to the WTI Scoreboard page.  (Note: the Admin password is "Gny2o25".)
1. Verify that **there is no "Group" dropdown** on the WTI scoreboard.  (This is because the PR only provides a "Group" dropdown when there's more than one selectable group.)
1. If desired, verify that the WTI scoreboard results as shown match PC2.  Since the contest is "Finalized" but the Scoreboard is still "Frozen", you will see "Pending Runs" on the WTI scoreboard.  To unfreeze the scoreboard, you'll need to do the following:
    -   On the PC2 Admin Settings tab, click `Unfreeze Scoreboard`.
    -  Click `Update` at the bottom of the Admin Settings screen.
1. Verify that the WTI scoreboard standings match those shown in the PC2 Admin `ConfigureContest>Standings` tab.

#### Verify that contests without groups work correctly.
1. Shut down the WTI browser, the WTI Server, the PC2 Admin, and the PC2 server.
1. In the PC2 distribution root, DELETE the `profiles` and `logs` folders, and the `profiles.properties` file.  
1. In the PC2 distribution `projects` folder, DELETE the `WebTeamInterface` folder (but not the `.zip` file from which that folder was extracted).  This puts you back to a clean PC2 installation.
1. Download and unzip the `PC2MultiSumitHelloSample.zip` file from https://drive.google.com/drive/folders/1GPl7c8MzmhSUZ905vvPkpE59FR_Y4EwC?usp=drive_link.  This is a PC2 profile for a contest with no groups.
1. Copy the `/profiles` folder and the `profiles.properties` file into the root of the unzipped (clean) PC2 distribution.
1. Follow steps 3-12 (only) above to get to the WTI Scoreboard page. The PC2 Admin password is the default; there should be 8 problems in the contest named Hello1, Hello2, Hello3, Hello4, Sumit1, Sumit2, Sumit3, and Sumit4.  There should be 28 runs in the contest.
1. Verify that **there is no "Group" dropdown** on the WTI scoreboard.  (This is because the PR only provides a "Group" dropdown when there's more than one selectable group.)
1. Verify that the WTI scoreboard standings match those shown in the PC2 Admin `ConfigureContest>Standings` tab.

### DIFF
--- a/projects/WTI-API/src/main/controllers/ContestController.java
+++ b/projects/WTI-API/src/main/controllers/ContestController.java
@@ -18,6 +18,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.XML;
@@ -38,9 +39,11 @@ import edu.csus.ecs.pc2.core.IniFile;
 import edu.csus.ecs.pc2.core.StringUtilities;
 import edu.csus.ecs.pc2.core.exception.IllegalContestState;
 import edu.csus.ecs.pc2.core.log.Log;
+import edu.csus.ecs.pc2.core.model.Account;
 import edu.csus.ecs.pc2.core.model.ClientId;
 import edu.csus.ecs.pc2.core.model.ClientType.Type;
 import edu.csus.ecs.pc2.core.model.ElementId;
+import edu.csus.ecs.pc2.core.model.Group;
 import edu.csus.ecs.pc2.core.model.IInternalContest;
 import edu.csus.ecs.pc2.core.model.Run;
 import edu.csus.ecs.pc2.core.scoring.DefaultScoringAlgorithm;
@@ -802,7 +805,8 @@ public class ContestController extends MainController {
 						//					logger.fine("Got the following XML from DSA:");
 						//					logger.fine(xmlStandings);
 						logger.info("Converting DSA XML to JSON");
-						currentJSONStandings = this.getJSONStandings(xmlStandings);
+						currentJSONStandings = this.enrichScoreboardTeamGroups(this.getJSONStandings(xmlStandings),
+								internalContest);
 						//					logger.fine("Got the following JSON standings:");
 						//					logger.fine(currentJSONStandings);
 					} catch (IllegalContestState e) {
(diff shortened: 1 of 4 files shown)


johnbrvc:
I found the following during testing using the procedure in the PR description.
1) General: Pressing **F5** when a group is selected on the scoreboard resets it to show all teams.  This is probably the desired operation, after all, **F5** is "reset".  I'm just making the comment.
2) For the PacNWSpring24 contest with divisions.  I logged in as team151 (a D2 team).  The WTI scoreboard shows both sets of problems on the same page which is somewhat confusing:

image

This is because I did not set `wtiBoardUseDivisions=true` in the `pc2v9.ini` file.  That being said, the procedure above should say to set that flag.  However, proceeding along, if I select group "British Columbia - D1" (for lack of anything better), it still shows the problems for D2 on the right of the scoreboard, albeit, all zeros:

image

I'm not sure what is supposed to be shown here.  Similarly, if I select, say, "California", I see both divisions and both sets of problems:

image

So, I decided to rerun the WTI server with the `wtiBoardUseDivisions=true` flag set in the `pc2v9.ini` file.  What I then see are the correct teams listed for the selected group, but their scores are always 0. 

image

I don't think that's right.  The Division filtering appears to work; that is, I only see D2 teams, even if a D1 group is selected:

image

Perhaps if `wtiBoardUseDivisions `is true, then omit the groups for the other division(s) from the drop down?

3) On the WTI server command prompt screen, I am seeing debug messages:

image

Are the accidental left over messages?
4) The only contest I noticed issues with was the one with divisions, PacNWSpring24. All other tests worked as described.

johnbrvc (changes requested):
I'm thinking something has to be done in the `.ts` code to deal properly with divisions.  I'm not sure where though since there is A LOT of new code to support groups.  Nevertheless, I think the division issue described in the comments of the PR should be addressed somehow.

clevengr:
> I'm thinking something has to be done in the `.ts` code to deal properly with divisions. I'm not sure where though since there is A LOT of new code to support groups. Nevertheless, I think the division issue described in the comments of the PR should be addressed somehow.

Dealing properly with divisions is such a significant process that it should be done under a separate issue; probably #689.  I'll work on that separately.

clevengr:
I updated the PR to reflect that it only applies to contests without divisions (i.e., no `wtiBoardUseDivisions` flag in the WTI Server's `pc2v9.ini`.  I'll work on handling divisions under Issue #689.

I also updated the "How to test the PR" steps to correctly test for a contest with teams in multiple groups without having to use the PacNW2024 contest (which uses Divisions).  The second test in the steps is the new one.

johnbrvc (approved):
Verified it works using the modified steps in the PR.

kkarakas (commented):
Looked at the code, looks like it would work but I am a bit uneasy with unknown and any types. Although I can imagine it would be annoying to know its shape and not that useful so it's okay.

kkarakas:
Verified that it works as stated. Tried a contest without any groups. The dropdown menu is not there just like expected. Tried a contest with groups. Group dropdown was there as expected. 

However one thing I noticed is after you select a group from dropdown menu Up, down buttons in keyboard, instead of controlling the scrolling of the page, now start to control the dropdown menu and change the selected group for the scoreboard. You have to click somewhere else in the page to return to original state.To me this behavior is unexpected and therefore a bug. Willing to discuss. 

Screenshot 2026-07-08 235414
Screenshot 2026-07-08 234449

clevengr:
@kkarakas I finally found some time to look into the issue you reported with the WTI Group dropdown stealing keyboard focus (sorry it took so long).  

As I suspected, the issue was that the \<select\> (the dropdown in the HTML) was retaining keyboard focus after selection.  The fix was to check in the TS code if the selected component was the Group dropdown, and if so to remove focus from it (an operation called `blur()`, as I learned).  Four lines of code changed (plus a few additional comments).

If you could retest and verify it works (and approve the PR if so) that would be great.  All you should have to do is run the very first test (the one with `WF49BakuFinalized.zip`) to verify it works (you'll need to make sure your browser window is small enough that it requires a scrollbar to fully test it).

kkarakas (approved):
Just tested it, works very well. Approved!

PR #1290, opened 2026-08-19
i1235: enforce uncompressed source size for pc2submit zip submissions

### LINKED ISSUE #1235: Teams can submit super big files as a problem submission.

Teams, using pc2submit (command line tool) can submit huge files causing problems downstream. WTI is fine.

As a team use pc2submit command line tool to submit an submission. Notice you can submit huge files. 

Reject huge file submissions. 

It can pass through. 

 - OS name and version under which the issue was detected (e.g. Windows 10 or Ubuntu 18.04):
 - PC2 Version (e.g. 9.7.0):   
 - Browser (if the bug is related to web-based team clients) (e.g. _Chrome Version 75.0.3770.142_ or _Safari Version 11.1.2_):

**Log Info**:  PC^2 generates logs (in the **logs** folder for client modules; under the **profiles** folder for the PC2 Server).  Please review the relevant logs and provide copies of them as attachments -- particularly if the logs show Exceptions being thrown.  

clevengr:

Previously, `pc2submit` only applied submission size limits based on the HTTP `Content-Length` of the **compressed** (zipped) POST. The result was that highly compressible source could exceed the contest limit on source code submission size (default 256 KiB) and still be accepted. (This happened in the NAC 2026 dress rehearsal, where submission 747 was 31 MB uncompressed, but only ~92 KB zipped.) This PR closes that loophole.

Specifically, CLICS `POST /submissions` (class `SubmissionService`) now checks **claimed uncompressed** zip entry sizes (`ZipEntry.getSize()` via the zip central directory) against `ContestInformation.getMaxSourceSizeInBytes()` **before** extracting (uncompressing) the zip file.

In addition, `SubmissionService` now utilizes new methods in `EventFeedUtilities` to monitor the actual in-memory unzipping process by keeping a **running inflated-byte cap** during extraction (uncompression) so a zip that understates sizes cannot expand without bound in memory.  (Such a condition could happen if a zip file lied about the total payload size -- which would be easy for an attacker to spoof by creating a zip with bad headers).

Oversized submits made via the `pc2submit` command now return HTTP **413** with the same “Source file(s) are too large" error message as WTI.

Additional notes:
- The submission paths for WTI and  `pc2team` are unchanged (they use `submitJudgeRun`, which is unaffected by this PR). 
- Running PC2 in Shadow mode still uses unlimited `getIFiles(byte[])` (see separate issue #1291).

Fixes #1235

### Environment in which the PR was developed 
Windows 11, Eclipse 2020-12 (4.18.0), Java version "1.8.0_271", Cursor version 3.13.25

The important case is a source code file submitted using `pc2submit` whose **uncompressed size > contest max**, but where the **zip + JSON is still small enough** to pass the Jetty `SubmitPostSizeLimitFilter` used in a POST to `SubmissionService`. Note that random large files often get rejected by the POST-size filter first and do **not** exercise this PR.

### Setup

1. Ensure you have Python3 on your machine.
1. Start a PC2 Server loading the `clics_sumithello` contest.
1. Start a PC2 Admin.
1. On the Admin **Configure Contest->Times** tab, start the contest clock running.
1. Start a PC2 Event Feeder client and login as `feeder1`.
1. Click the `Start` button on the EF GUI to start the **Web Server** running (note that the default port is **50443**). Examine the console and confirm the webserver is running.
1. Verify that your **.netrc** file contains an entry like
`machine localhost login team1 password team1`.  (If not, create such an entry.)  
In Windows, the **.netrc** file is located at `C:\USERS\<username>`.
1. Open a terminal window in the PC2 distribution and enter the command
```
python ./bin/pc2submit --url https://localhost:50443 --problem A ./samps/src/hello.cpp
```
This should return a message like
```
Warning: `./samps/src/hello.cpp' has not been modified for 482857 minutes!
Submission information:
  filename:    ./samps/src/hello.cpp
  contest:     SumH
  problem:     A
  language:    GNU C++
  url:         https://localhost:50443/
There are warnings for this submission!
Do you want to continue? (y/n)
```
If you do not get a message like the above, **stop** and check your network setup; you need to be able to connect to the contest in order to run the following tests.
(Note: the "warning" is simply because you are submitting a PC2 sample source file that hasn't been changed in quite a while).

9. Enter `y` to submit the program, then verify that you get a message on the console like `Submission received: id = s1, time = 20:19:47`.
1. On the PC2 Admin **Run Contest->Runs** tab, verify that you see a newly-submitted run.  This proves that you can submit a run using `pc2submit`.  

### Primary Test: submisson of a compressible source file which is over the uncompressed limit
1. Download the file https://github.com/icpcsysops/NAC2026/blob/main/contests/dress/results/primary/replay/747.zip from the NAC2026 Dress Rehearsal.  This is a zip file of about 91K (so, much less than the default source size limit of 256K) but which contains a (compressed) source file that is over 30MB when uncompressed.
1. Unzip the downloaded zip file.  This should produce a 30MB file named `jeirr.cpp`.
1. Submit the `jeirr.cpp` file with `pc2submit` as above.
1. Verify that when you answer `y` to the question `Do you want to continue?`, a message like the following is returned:
```
Submission failed (code 413 - Unable to submit run: Source file(s) are too large (31610880 bytes) - maximum is 262144 bytes.)
```
5. Check the PC2 Admin RUNS grid; verify that no new submission was accepted.

### Optional additional tests:  WTI / pc2team unchanged

Submit an oversized file via **pc2team** and/or **WTI**.  Verify that such a file is still rejected (it should not be accepted because **pc2team** and **WTI** already contained code to prohibit such a submission, and no changes were made to either of these submission components).

### Optional additional test:  POST filter still works

Submit an **incompressible** file larger than the max (e.g. 300 KiB of random bytes).  Such a file should still be rejected by the existing POST-size filter which is applied before the file is ever unzipped. That is how it should work; this PR does not replace that filter.

### DIFF
--- a/src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
+++ b/src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
@@ -4,15 +4,19 @@ package edu.csus.ecs.pc2.convert;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 import java.util.zip.ZipInputStream;
 
+import edu.csus.ecs.pc2.core.exception.SubmissionRejectedException;
 import edu.csus.ecs.pc2.core.model.IFile;
 import edu.csus.ecs.pc2.core.model.IFileImpl;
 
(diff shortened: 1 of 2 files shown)


src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
```diff
@@ -131,6 +135,94 @@ public static List<String> fetchRunFileNames(String dirname, String runId) {
...
+     * Reject a zip if the claimed uncompressed size of its file entries exceeds {@code maxSourceSizeBytes}.
+     * Uses zip central-directory metadata ({@link ZipEntry#getSize()}); does not inflate entry contents.
+     * A {@code maxSourceSizeBytes} of 0 or less means there is no source-size limit.
+     *
+     * @param base64Data base64-encoded zip bytes
+     * @param maxSourceSizeBytes contest maximum combined source size in bytes
+     * @throws SubmissionRejectedException if the claimed uncompressed size exceeds the limit
+     * @throws IllegalArgumentException if the data is not valid base64, the zip is unreadable, or an entry size is unknown
+     */
+    public static void enforceUncompressedZipSourceSizeLimit(String base64Data, long maxSourceSizeBytes)
```
johnbrvc, line 148: What is the purpose of this method?  Was the intent to call this from SubmissionService instead of doing the base64 stuff there?  I might suggest either removing this method, or, calling it from SubmissionService, getting rid of the Base64 decoder there.  Or, removing all these "enforcexxxx" methods as described in the next comment?

src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
```diff
@@ -131,6 +135,94 @@ public static List<String> fetchRunFileNames(String dirname, String runId) {
...
+
+    /**
+     * Sum the claimed uncompressed sizes of file entries in a zip, without extracting those entries.
+     * Uses {@link ZipFile} so sizes come from the central directory rather than by inflating payloads.
+     *
+     * @param zipBytes raw zip file bytes
+     * @return total uncompressed size in bytes of non-directory entries
+     * @throws IllegalArgumentException if {@code zipBytes} is null, the zip cannot be read, or any file entry has unknown size ({@link ZipEntry#getSize()} {@code < 0})
+     */
+    public static long getUncompressedZipSize(byte[] zipBytes) {
```
johnbrvc, line 183: Basically, you're reading the entire zip file twice (yes, it has to read the whole file just to look at the entries - there is not a directory at the beginning) AND creating a seemingly unnecessary temporary file.  Once just to check the length of the unzipped files and once to actually extract files.  Why not do this all in getIFiles() so it only reads the file once using the streamer without re-reading the entire thing.  This seems overly complex, unnecessary and inefficient to me.  I know there's something to be said for modularity, but I think the cost here is too high (doing more than twice the work and I/O).  In addition, I'm not sure why you're creating a temporary disk file?  That also seems inefficient when you can uses a memory zip streamer as in getIFiles().  For 99.9% of the cases, there is only one file in the zip anyway (the source), so it can easily be handled in the getIFiles loop I think - no need to create temp files, etc.  Maybe I'm missing something?

johnbrvc (commented):
Comments made inline.

src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
clevengr: Agreed; method removed in next push.

src/edu/csus/ecs/pc2/convert/EventFeedUtilities.java
clevengr: Agreed.  The whole "temp file" thing, and the corresponding "enforcexxx" methods, were AI-created.  When asked about it, the AI essentially responded "oh, yeah; that's not really neccessary".    Removed in next push.

clevengr:
> Comments made inline.

I agreed with all your comments; the latest push removes those unnecessary "enforcexxx" methods and also removes use of the temp file.   Re-review requested.

johnbrvc (approved):
Code looks good.
Executes as expected.  (Did not do WTI "optional" test.)

